PR #1, opened 2017-08-31
Fixed Inf comparison

cscherrer:
In the original code
```julia
while L(θ′) == -Inf || ∇L(θ′) == -Inf
```
the intent seems to be to determine whether the log-likelihood or any component of its gradient is infinite. This is supported by the referenced Python, in the line
```python
while np.isinf(logpprime) or np.isinf(gradprime).any():
```

The Julia code is subtly different:
1. It tests only for *negatively* infinite log-likelihood
2. The test `∇L(θ′) == -Inf` compares a scalar and a gradient, almost certainly a bug.

This is easily corrected by changing the line to 
```julia
while isinf(L(θ′)) || any(isinf(∇L(θ′)))
```


xukai92:
Thanks for the fix!

PR #5, opened 2019-02-22
Hamiltonian type RFC

xukai92:
Current design of type system for HMC:

- A type `Metric` to indicate which metric space we are working on
- A type `Hamiltonian` which contains the metric space, the log density function of the target distribution and its gradient function
- A type `Leapfrog` (a sub-type of `AbstractIntegrator`) to indicate we are using leapfrog integrator
  - We may want other integrators later
- A type `Trajectory` to indicate the way we build trajectory
  - `StaticTrajectory` is just HMC and `NoUTurnTrajectory` will be NUTS
- A type `Proposal` to wrap `Trajectory` with what kind of trajectory sampler is used
  - ~~Question: do we want this type, or may be an alternative type called `Proposal`, which build abstraction on proposal distribution directly.~~
- HMC samplers are just different combination of these. This package is not aiming to provide integrated samplers but those building blocks. However, some simple examples like normal HMC and NUTS will be there.
  - [x] HMC
  - [x] NUTS

Side note:
- All functions are supposed to be immutable
- During adaptation, we are meant to keep constructing new instances for `Metric` and `Leapfrog` types as their parameters are changing. However, after adaptation, all instances should be fixed.
- I hope this implementation is type stable on its own.


trappmartin:
Sounds like a good plan. Some minor comments:

* Regarding the `Trajectory` type, I think it would be better to have a `AbstractProposal` type as this would be more generic and opens more possibilities.
* I really like the idea of a `Metric` or `AbstractMetric` type!
* Maybe we should add an `AbstractIntegrator` type and make `Leapfrog` a subtype?

xukai92:
> * Regarding the `Trajectory` type, I think it would be better to have a `AbstractProposal` type as this would be more generic and opens more possibilities.

I was also thinking about this. One question related to this is that do we want this package to handle all MCMC interface, or mainly just core HMC stuff. If we only suppose to handle HMC related stuff, building a proposal using HMC related functions should happen in Turing, where a `Proposal` type can be designed to unify MH, HMC, PG and other MCMC samplers.

> * Maybe we should add an `AbstractIntegrator` type and make `Leapfrog` a subtype?

Nice point - will do it.

xukai92:
I realised that there is actually no conflict to also have a proposal here so I added it. It also make the sampling from trajectory abstraction clearer.

src/hamiltonian.jl
```diff
@@ -0,0 +1,61 @@
...
+    return p' * h.metric.M⁻¹ * p / 2
+end
+
+# Momentum sampler
+function rand_momentum(h::Hamiltonian{UnitMetric}, θ::AbstractVector{T}) where {T<:Real}
+    return randn(length(θ))
+end
+
+function rand_momentum(h::Hamiltonian{DiagMetric{T}}, θ::AbstractVector{T}) where {T<:Real}
+    return randn(length(θ)) ./ sqrt.(h.metric.M⁻¹)
```
mohamed82008: If one `temp` is defined, we can do:
```julia
temp .= randn.()
temp ./= sqrt.(h.metric.M⁻¹)
```

src/hamiltonian.jl
```diff
@@ -0,0 +1,61 @@
...
+end
+
+function rand_momentum(h::Hamiltonian{DenseMetric{T}}, θ::AbstractVector{T}) where {T<:Real}
+    C = cholesky(Symmetric(h.metric.M⁻¹))
```
mohamed82008: Would be nice to cache the Cholesky factor, and a `temp` so we can do:
```julia
temp .= randn.()
@! temp = C.U \ temp
return temp
```

src/diagnosis.jl
```diff
@@ -0,0 +1,3 @@
+function EBFMI(Es)
+    return sum((Es[2:end] - Es[1:end-1]).^2) / var(Es)
```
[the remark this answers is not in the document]
xukai92: Thanks - will take your version! Which operation allocates unnecessary memory?

src/hamiltonian.jl
```diff
@@ -0,0 +1,61 @@
+# TODO: add a type for kinetic energy
+
+struct Hamiltonian{M<:AbstractMetric}
+    metric      ::  M
+    _logπ       ::  Function
```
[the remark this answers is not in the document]
xukai92: Got it thanks!

src/metric.jl
```diff
@@ -0,0 +1,13 @@
+abstract type AbstractMetric end
+
+struct UnitMetric <: AbstractMetric end
+
+struct DiagMetric{T<:Real} <: AbstractMetric
+    # Diagnal of the inverse of the mass matrix
+    M⁻¹ ::  AbstractVector{T}
```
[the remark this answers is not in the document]
xukai92: Do you mean instead of `AbstractVector{T}` I should do `T` or `A{T}`?

src/hamiltonian.jl
```diff
@@ -0,0 +1,61 @@
...
+function _K(h::Hamiltonian{UnitMetric}, r::AbstractVector{T}, ::AbstractVector{T}) where {T<:Real}
+    return sum(abs2, r) / 2
+end
+
+function _K(h::Hamiltonian{DiagMetric{T}}, r::AbstractVector{T}, ::AbstractVector{T}) where {T<:Real}
+    return sum(abs2, r .* sqrt.(h.metric.M⁻¹)) / 2
+end
+
+function _K(h::Hamiltonian{DenseMetric{T}}, r::AbstractVector{T}, ::AbstractVector{T}) where {T<:Real}
+    return p' * h.metric.M⁻¹ * p / 2
```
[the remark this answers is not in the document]
xukai92: In this case `Metric` would be mutable. Does it affect any efficiency?

src/hamiltonian.jl
```diff
@@ -0,0 +1,61 @@
...
+end
+
+# Kinetic energy
+# NOTE: the general form of K depends on both θ and r
+function _K(h::Hamiltonian{UnitMetric}, r::AbstractVector{T}, ::AbstractVector{T}) where {T<:Real}
+    return sum(abs2, r) / 2
+end
+
+function _K(h::Hamiltonian{DiagMetric{T}}, r::AbstractVector{T}, ::AbstractVector{T}) where {T<:Real}
+    return sum(abs2, r .* sqrt.(h.metric.M⁻¹)) / 2
```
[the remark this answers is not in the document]
xukai92: Thanks! Can you explain a bit more?

src/hamiltonian.jl
xukai92: Are you talking about reducing memory allocation or something? And by `temp1 .= randn.()` do you mean `temp1 .= randn.(length(θ))`?

src/hamiltonian.jl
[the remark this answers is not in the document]
xukai92: BTW I'm happy to change it to mutable if this doesn't have any efficiency drawback.

src/diagnosis.jl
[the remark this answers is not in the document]
mohamed82008: Indexing into an array with a range allocates memory for the result array. Subtracting 2 arrays allocates memory for the result. Squaring the result allocates another vector for the squared difference. If you do `(a .- b).^2`, the subtraction and squaring will be fused so only one result vector will be allocated for the squared difference. But even this vector is not necessary to allocate, as all we need is the sum of the values in there. The range indexing can also be avoided. So my version calculates the squared difference on the fly as the sum is taking place, which avoids any unnecessary allocations for `getindex`, subtraction or squaring, so it's shorthand for a loop.

src/metric.jl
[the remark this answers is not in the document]
mohamed82008: More like:
```julia
struct DiagMetric{T<:Real, A <: AbstractVector{T}}
    M⁻¹ ::  A
end
```

src/hamiltonian.jl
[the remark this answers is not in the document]
mohamed82008: No, `Metric` doesn't need to be mutable. `@! h.temp = h.metric.M⁻¹ * p` is syntactic sugar for `mul!(h.temp, h.metric.M⁻¹, p)` which writes the result of the multiplication into `h.temp`.

src/hamiltonian.jl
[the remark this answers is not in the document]
mohamed82008: `r .* sqrt.(h.metric.M⁻¹)` does column-wise multiplication and allocates a result matrix that has the column-wise product. This matrix is unnecessary as we are only interested in the sum. `LazyArrays` does lazy broadcasting. `t1` above is a lazy form of `r.^2`, and `t2` is a lazy `r.^2 .* h.metric.M⁻¹`, so the column-wise multiplication will be done on the fly as the `sum` function is asking for elements to add.

src/hamiltonian.jl
mohamed82008: Yes this is for reducing the memory allocations. `temp1 .= randn.()` puts a different random number in each element of `temp1`. `temp1 .= randn.(length(θ))` tries to put a random vector in each of element of `temp1` so it will error. `temp1 .= randn()` will put the *same* random number in all the elements of `temp1`. So we want the first.

src/diagnosis.jl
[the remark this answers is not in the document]
xukai92: Thanks for the explanation!

src/hamiltonian.jl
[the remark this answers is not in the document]
xukai92: OK I meant to ask about the mutability about `Hamiltonian`. Is that a concern?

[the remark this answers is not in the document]
xukai92: OK I realised that there should be no issue.

src/hamiltonian.jl
xukai92: Great thanks!

src/hamiltonian.jl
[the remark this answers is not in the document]
xukai92: @mohamed82008 I've resolved all memory comments from you except this one (they do give me some improvement! thanks again!). For this one, only `DenseEuclideanMetric` need this `temp` variable, which seems to be unnecessary if I put it as a field of `h` as other metric types do not use it. Any idea how I can work around it? Is there a way to do this computation similar to what you suggest via `BroadcastArray`?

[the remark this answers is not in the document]
mohamed82008: In general immutable types are more performant than mutable ones, because they are stack allocated and incur no allocations or loading time from the heap memory. I also find immutable types easier to reason about in many cases. 

As for the `temp` problem, linear algebra on normal arrays in Julia is done with the heavily optimized BLAS library which is typically faster than a hand-written loop of the naive approach. For instance, `sum(A[i,j]*b[i]*b[j] for i in 1:size(A, 1), j in 1:size(A, 2))` is mathematically equivalent to `b' * A * b` but the latter is generally faster.  Unfortunately, I can't think of any workaround as BLAS doesn't have a routine for this extremely common operation, aka Galerkin projection. I plan to give this branch a second thorough read next weekend and make a PR to your branch if I see some bugs or obvious improvements. You can think of a way to reuse the temporaries which are not used anymore in the rest of the current iteration. If there is no way, then try to benchmark the pre-allocated and normal versions and see which is faster.

trappmartin:
Does your current RF consider transformations for constrained variables? I might be wrong, but wouldn't you need an additional mechanism that maps `theta` back to the constrained space to compute the Metropolis acceptance probability? Or is this unnecessary?

src/hamiltonian.jl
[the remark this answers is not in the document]
mohamed82008: FWIW, I was under the impression that `M⁻¹` above is a matrix which is not the case for a diagonal metric, so the above solution can be simplified to `sum(abs2(r[i]) * h.metric.M⁻¹[i] for i in 1:length(r)) / 2`. `BroadcastArray` is only necessary if some matrix-vector broadcasting is done.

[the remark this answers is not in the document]
xukai92: OK thanks!

src/hamiltonian.jl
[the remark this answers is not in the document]
xukai92: I see. Thanks!

xukai92:
@trappmartin This package is not going to handle transformations but only sampling from an unconstrained space. Building a log-joint which supports constrained variables is done automatically with the compiler and `VarReply` in Turing.

src/hamiltonian.jl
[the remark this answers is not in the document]
xukai92: OK. I realised that `h.temp` in your example is just `h._dHdr`. Maybe I can just reuse that allocation.

mohamed82008:
@xukai92 Please change the `Integer` fields to `Int` for performance, there are a few of them.

src/hamiltonian.jl
```diff
@@ -0,0 +1,77 @@
...
+
+function rand_momentum(h::Hamiltonian{T,DiagEuclideanMetric{T,M},F1,F2,A}) where {T<:Real,M<:AbstractVector{T},F1,F2,A<:AbstractVector{T}}
+    h._r .= randn.()
+    h._r .= h._r ./ h.metric.sqrtM⁻¹
+    return h._r
+end
+
+function rand_momentum(h::Hamiltonian{T,DenseEuclideanMetric{T,M},F1,F2,A}) where {T<:Real,M<:AbstractMatrix{T},F1,F2,A<:AbstractVector{T}}
+    h._r .= randn.()
+    h._r .= h.metric.cholM⁻¹ \ h._r
```
mohamed82008: `ldiv!(h.metric.cholM⁻¹, h._r)` is the inplace version.

src/hamiltonian.jl
```diff
@@ -0,0 +1,77 @@
+# TODO: add a type for kinetic energy
+
+struct Hamiltonian{T<:Real,M<:AbstractMetric{T},F1,F2,A<:AbstractVector{T}}
+    metric      ::  M
+    logπ        ::  F1
+    dlogπdθ     ::  F2
+    # Below are for efficient memory allocation
+    _dHdθ       ::  A
+    _dHdr       ::  A
+    _r          ::  A
```
yebai: Some comment on variable meaning would be helpful.

src/hamiltonian.jl
```diff
@@ -0,0 +1,77 @@
+# TODO: add a type for kinetic energy
+
+struct Hamiltonian{T<:Real,M<:AbstractMetric{T},F1,F2,A<:AbstractVector{T}}
+    metric      ::  M
+    logπ        ::  F1
+    dlogπdθ     ::  F2
```
yebai: perhaps `dlogπdθ` ==> `∂logπ∂θ`?

xukai92:
> Looks good. I left some comments below. Are you planning to introduce some adaption types (e.g. a `Tuner` type)?

Not planned at the moment. The way I'm thinking to use this package in Turing.jl is independent on the adaptation code. What we could do is creating a new `Metric` each time we adapt pre-conditioning matrix and creating a new `Leapfrog` each time we adapt step-size.

src/hamiltonian.jl
xukai92: Yes.

src/hamiltonian.jl
xukai92: I put a note in `HMC.jl` on the notations and the variable name is actually informative enough on what's the term is. But sure we can improve the documentation a bit if we want.

src/hamiltonian.jl
xukai92: Thanks!

xukai92:
@yebai One design issue I haven't resolved yet is in which level do we want to unify the interface. In the current code, all interface are unified. But this cause problem on NUTS in which the `lastpoint` function for `NoUTurnTrajectory` need to some how pass the `da_stats` (the statistics needed for dual averaging) to the caller. There are three options here
1. Not unify the interface on the trajectory level but the proposal level. In this case we just return an additional number from `lastpoint` for `NoUTurnTrajectory` and handle that in the corresponding `propose` function.
2. Allow both `lastpoint` and `propose` to take a call-back function as keyword argument, for which if provided in `lastpoint` for `NoUTurnTrajectory` it can be used to update the `da_stats`
3. Introduce a new type called `AbstractTrajStats`. For static HMC, this only contains `n_valid` and for NUTS, it contains an additional term for `da_stats`.
What do you think?

yebai:
> Not planned at the moment. The way I'm thinking to use this package in Turing.jl is independent on the adaptation code. What we could do is creating a new `Metric` each time we adapt pre-conditioning matrix and creating a new `Leapfrog` each time we adapt step-size.

Sounds good!

> * Not unify the interface on the trajectory level but only the proposal level. In this case we just return an additional number from `lastpoint` for `NoUTurnTrajectory` and handle that in the corresponding `propose` function.

Seems option 1 is sufficient. Actually, I don't really have a preference for this.

BTW, is there any chance that we can merge the `Proposal` and `Trajectory` types? It seems the functionality of `Proposal` is merely to sample a point from `Trajectory`, which can be implemented as `propose(t::Trajectory)`.

xukai92:
We could do that but what's the way to specify which way we want to use?

yebai:
> We could do that but what's the way to specify which way we want to use?

I was thinking about each HMC algorithm implements a `step` function, which calls one of these functions (e.g. `last`, `uniform` and `multinomial`) explicitly. This is less general than the design you proposed, but there aren't many HMC variants anyway.

xukai92:
I think there still some degrees of freedom for the same HMC algorithm on the sampling method for trajectory. This is related to https://github.com/TuringLang/Turing.jl/issues/453. Basically even for the static HMC we can use different ways to take from the trajectory.

yebai:
> I think there still some degrees of freedom for the same HMC algorithm on the sampling method for trajectory. This is related to [TuringLang/Turing.jl#453](https://github.com/TuringLang/Turing.jl/issues/453). Basically even for the static HMC we can use different ways to take from the trajectory.

I agree; I'm only trying to see whether we can remove one type without losing a lot flexibility.

src/hamiltonian.jl
```diff
@@ -0,0 +1,77 @@
...
+    h._∂H∂r .= h.metric.M⁻¹ .* r
+    return h._∂H∂r
+end
+
+function ∂H∂r(h::Hamiltonian{T,DenseEuclideanMetric{T,M},F1,F2,A}, r::AbstractVector{T}) where {T<:Real,M<:AbstractMatrix{T},F1,F2,A<:AbstractVector{T}}
+    mul!(h._∂H∂r, h.metric.M⁻¹, r)
+    return h._∂H∂r
+end
+
+function H(h::Hamiltonian, θ::AbstractVector{T}, r::AbstractVector{T}) where {T<:Real}
```
yebai: H ==> hamiltonian_energy / hamiltonian_e

src/hamiltonian.jl
```diff
@@ -0,0 +1,77 @@
...
+    return K(h, r, θ) + V(h, θ)
+end
+
+function V(h::Hamiltonian, θ::AbstractVector{T}) where {T<:Real}
```
yebai: V ==> potential_energy / potential_e

src/hamiltonian.jl
```diff
@@ -0,0 +1,77 @@
...
+    return -h.logπ(θ)
+end
+
+# Kinetic energy
+# NOTE: the general form of K depends on both θ and r
+function K(h::Hamiltonian{T,UnitEuclideanMetric{T},F1,F2,A}, r::AbstractVector{T}, ::AbstractVector{T}) where {T<:Real,F1,F2,A<:AbstractVector{T}}
```
yebai: K ==> kinetic_energy / kinetic_e

src/integrator.jl
```diff
@@ -0,0 +1,51 @@
+abstract type AbstractIntegrator end
+
+struct Leapfrog{T<:Real} <: AbstractIntegrator
```
yebai: Nice to have a type for `Integrator`s!

src/proposal.jl
```diff
@@ -0,0 +1,18 @@
+abstract type AbstractProposal{T<:AbstractTrajectory} end
```
yebai: Perhaps consider merging the `AbstractProposal` and `AbstractTrajctory` types by: 

```julia
abstract type AbstractProposal end
abstract type AbstractHamiltonianTrajectory <: AbstractProposal end
abstract type StaticTrajectory <: AbstractHamiltonianTrajectory end
abstract type NoUTurnTrajectory <: AbstractHamiltonianTrajectory end
abstract type UniformProposal <: NoUTurnTrajectory
abstract type MultinomialProposal <: NoUTurnTrajectory
abstract type TakeLastProposal <: StaticTrajectory # this type might be unnecessary since we always take the last point for static trajectory

# then define `propose` for each `t`
propose(t::AbstractHamiltonianTrajectory, ...) 
```

This design of types might be easier to reason about since intuitively Hamiltonian simulation is a way to make MCMC proposals. It also allow us to extend `AbstractProposal` with non-Hamiltonian proposals.

yebai:
Hi @xukai92. Excellent work. I made some additional comments above. Do you want to do more refactoring work in this PR, or merging this PR into master first and leave additional refactoring in seprate PRs?

src/proposal.jl
xukai92: > this type might be unnecessary since we always take the last point for static trajectory
Is there benefit of using a multinomial sampling instead?

xukai92:
I'd like to resolve your comments first. I will do it by tomorrow.

xukai92:
TODO
- [ ] resolve Hong's comments
- [ ] add interface to provide RNG

xukai92:
@mohamed82008 I reverted some functions about pre-allocating memory so that only intermediate variables are pre-allocated and all functions remains immutable. This is to avoid un-expected behaviour due to mutability. We probably want to implement mutable version of functions to pre-allocate memory for return variables explicitly later. Does it sound good to you?

src/proposal.jl
xukai92: @yebai Does using a multinomial sampling from static trajectory make sense?

xukai92: I'd like to merge them if arbitrary combination of sampling and ways of building trajectory doesn't make (practical) sense.

xukai92:
@yebai All solved except https://github.com/TuringLang/AdvancedHMC.jl/pull/5#discussion_r261375771.

src/proposal.jl
yebai: > Is there benefit of using a multinomial sampling instead?

It avoids an auxilairy slice variable; by the virtue of Rao-Blackwell theorem, this can reduce variance in the samples and ensure better performance when computing expectations using HMC samples. 

> Does using a multinomial sampling from static trajectory make sense?

No, it's not a valid transition with the current static trajectory implementation https://github.com/TuringLang/AdvancedHMC.jl/pull/5/files#diff-f48defe9803ddea4f759bce857b70de0R8.

xukai92:
TODO:
- [ ] remove `Proposal`
- [ ] more tests for leapfrog
  - [ ] long simulation and check average Hamiltonian energy
  - [ ] Eq 2.11 from https://arxiv.org/abs/1206.1901

xukai92:
@yebai I'v merged the trajectory type and the proposal type as agreed. For the `Point` thing I need to think a bit more (together with the stats interface). We'd better to do that in another PR (I created an issue https://github.com/TuringLang/AdvancedHMC.jl/issues/17 for this). Can you take a look at this PR and merge it if everything looks fine.

mohamed82008:
> @mohamed82008 I reverted some functions about pre-allocating memory so that only intermediate variables are pre-allocated and all functions remains immutable. This is to avoid un-expected behaviour due to mutability. We probably want to implement mutable version of functions to pre-allocate memory for return variables explicitly later. Does it sound good to you?

Sorry, I missed this question. I think it would be nice to support mutable and immutable array types, e.g. `StaticArrays.SArray` and `StaticArrays.MArray` or `Vector`. I saw Chris mention somewhere that DiffEq does this by defining 2 versions of the function, one for mutable arrays and one for immutable. We might want to explore something like this in a future PR. The mutating version takes a first argument the pre-allocated output vector whereas the non-mutating version doesn't. Supporting `StaticArrays` would be an interesting direction to pursue to gain some speedup for parameter vectors of size <1000 or so. Small arrays is the sweet spot of StaticArrays putting ordinary arrays to shame in many cases.

PR #44, opened 2019-04-01
Remove dim and resolve the resulting problems

willtebbutt:
Removes `dim` as a field of `DiagEuclideanMetric` and `DenseEuclideanMetric` in favour of just using `getproperty` methods. This introduces some ambiguities, so we have to go with the following interface for construction:
```julia
UnitEuclideanMetric{Float64}(D)
DiagEuclideanMetric(::AbstractVector)
DenseEuclideanMetric(::AbstractMatrix)
```
and we've dropped the construct-with-element-from-space interface, which was really only a convenience thing anyway (as I understand it).

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ n_adapts = 2_000
 θ_init = randn(D)
 
 # Define metric space, Hamiltonian and sampling method
-metric = UnitEuclideanMetric(D)
+metric = DenseEuclideanMetric(D)
 h = Hamiltonian(metric, logπ, ∂logπ∂θ)
 prop = NUTS(Leapfrog(find_good_eps(h, θ_init)))
 adaptor = StanNUTSAdaptor(n_adapts, PreConditioner(metric), NesterovDualAveraging(0.8, prop.integrator.ϵ))
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ n_adapts = 2_000
 θ_init = randn(D)
 
 # Define metric space, Hamiltonian and sampling method
-metric = UnitEuclideanMetric{eltype(θ_init)}(D)
+metric = UnitEuclideanMetric(D)
 h = Hamiltonian(metric, logπ, ∂logπ∂θ)
 prop = NUTS(Leapfrog(find_good_eps(h, θ_init)))
 adaptor = StanNUTSAdaptor(n_adapts, PreConditioner(metric), NesterovDualAveraging(0.8, prop.integrator.ϵ))
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ n_adapts = 2_000
 θ_init = randn(D)
 
 # Define metric space, Hamiltonian and sampling method
-metric = DenseEuclideanMetric(θ_init)
+metric = UnitEuclideanMetric{eltype(θ_init)}(D)
 h = Hamiltonian(metric, logπ, ∂logπ∂θ)
 prop = NUTS(Leapfrog(find_good_eps(h, θ_init)))
 adaptor = StanNUTSAdaptor(n_adapts, PreConditioner(metric), NesterovDualAveraging(0.8, prop.integrator.ϵ))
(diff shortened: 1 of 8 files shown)


willtebbutt:
Latest commits:
- remove element types from the `AbstractMetric`s, as they don't appear to have been necessary, and caused interface problems further down the line.
- fix an issue in `rand_momentum(::AbstractRNG, ::DiagEuclideanMetric)` whereby the `_temp` variable was being returned, and causes a bug. The fact that it works at the minute was purely a consequence of the `_temp` field being initialised to contain zeros.

As a consequence of the above, we have the following interface for metric construction:
```julia
UnitEuclideanMetric(D)
DiagEuclideanMetric(::AbstractVector)
DenseEuclideanMetric(::AbstractMatrix)
```
The only difference relate to the above is that we no longer need the type parameter to construct `UnitEuclideanMetric`.

@xukai92 this is really for review now.

willtebbutt:
Current break seems to only happen on 1.0 and may well be related to [this](https://github.com/JuliaLang/julia/pull/31512). Beyond that, I have _literally_ no idea what might be happening.

xukai92:
I see. Then I will ignore the check and review first.

README.md
```diff
@@ -25,7 +25,7 @@ n_adapts = 2_000
 θ_init = randn(D)
 
 # Define metric space, Hamiltonian and sampling method
-metric = DenseEuclideanMetric(θ_init)
+metric = UnitEuclideanMetric(D)
```
xukai92: Why do we change the example of using dense?

willtebbutt:
@xukai92 just resolved some small issues.

src/metric.jl
```diff
@@ -32,54 +28,39 @@ struct DiagEuclideanMetric{T<:Real,A<:AbstractVector{T}} <: AbstractMetric{T}
...
-function (dem::DenseEuclideanMetric)(M⁻¹::A) where {T<:Real,A<:AbstractMatrix{T}}
-    return DenseEuclideanMetric(dem.dim, M⁻¹)
+function DenseEuclideanMetric(M⁻¹::AbstractMatrix{T}) where {T<:Real}
+    _temp = Vector{T}(undef, size(M⁻¹, 1))
+    return DenseEuclideanMetric(M⁻¹, cholesky(Symmetric(M⁻¹)).U, _temp)
 end
 
-function DenseEuclideanMetric(θ::A) where {T<:Real,A<:AbstractVector{T}}
-    dim = length(θ)
-    return DenseEuclideanMetric(dim, Matrix{T}(I, dim, dim))
```
xukai92: I suggest we provide the interface `AbstractMetric(d::Int)` to allow users initialise with identity `M⁻¹` for both `DenseEuclideanMetric` and `DiagEuclideanMetric`

README.md
xukai92: If this is because of we don't have easy interface, see my comments below.

xukai92:
Have you check type stability after the changes?

xukai92:
By which I mean run `@code_warntype` for `sample()` function.

src/metric.jl
willtebbutt: Good shout, will implement.

willtebbutt:
> Have you check type stability after the changes?

Yes, appears to be fine.

README.md
willtebbutt: It was, will revert with proposed interface below.

xukai92:
> Yes, appears to be fine.

Awesome!

src/metric.jl
xukai92: Happy to merge it when this is done.

willtebbutt:
Right, I've created the default methods. We get `Float64`s by default, but the user can of course provide a matrix explicitly if this isn't the desired behaviour.

xukai92:
Great work! Thanks!

xukai92:
I'll merge it when CI passes.

Issue #45, opened 2019-04-02
Instantiation of new `AbstractMetric`s


willtebbutt:
We currently have a method for each `AbstractMetric` which enables us to make another instance of the metric of the following form:
```julia
metric = ...
new_metric_of_same_type = metric(inverse_mass_matrix)
```
Why do we need such a method?

xukai92:
This is for adaptation, during which we create a new `metric` based on the new M inverse. Hong and I believe make metric immutable instead of change in place would make code easier to maintain.

willtebbutt:
I see -- I'm fully on board with the motivation for this, but I don't know that I think the current interface is ideal.

In particular, it's not clear to me why calling an object of type `AbstractMetric` should create another object of type `AbstractMetric` -- I would naively have expected the evaluation of a metric to take two elements from the corresponding space and compute the distance between them. Would you consider renaming this to something more informative?

I do appreciate that I'm being a bit pedantic...

xukai92:
Yes I'm open to rename it.

PR #47, opened 2019-04-02
Improve NUTS interface

xukai92:
- Move hyper-parameters from function keywords to NUTS's proposal type
- Allow passing RNG to `sample()`
- Logging during adaptation period
- Fixed a numerical issue we missed when importing the code https://github.com/TuringLang/AdvancedHMC.jl/pull/47/commits/1abf11a8fece390505ca90d99b320313971817ad#diff-adfee0db137e1cc302dd54b62f980fcfR32
  - Corresponding Turing line: https://github.com/TuringLang/Turing.jl/blob/master/src/inference/support/hmc_core.jl#L121

### DIFF
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -1,35 +1,21 @@
 # TODO: add a type for kinetic energy
 
-struct Hamiltonian{T<:Real,M<:AbstractMetric{T},F1,F2}
-    metric      ::  M
-    logπ        ::  F1
-    ∂logπ∂θ     ::  F2
+struct Hamiltonian{M<:AbstractMetric, Tlogπ, T∂logπ∂θ}
+    metric::M
+    logπ::Tlogπ
+    ∂logπ∂θ::T∂logπ∂θ
 end
 
 # Create a `Hamiltonian` with a new `M⁻¹`
-function (h::Hamiltonian)(M⁻¹)
-    return Hamiltonian(h.metric(M⁻¹), h.logπ, h.∂logπ∂θ)
-end
+(h::Hamiltonian)(M⁻¹) = Hamiltonian(h.metric(M⁻¹), h.logπ, h.∂logπ∂θ)
 
-function ∂H∂θ(h::Hamiltonian, θ::AV)::AV where {T<:Real,AV<:AbstractVector{T}}
-    return -h.∂logπ∂θ(θ)
-end
+∂H∂θ(h::Hamiltonian, θ::AbstractVector) = -h.∂logπ∂θ(θ)
 
-function ∂H∂r(h::Hamiltonian{T,UnitEuclideanMetric{T},F1,F2}, r::A) where {T<:Real,F1,F2,A<:AbstractVector{T}}
-    return copy(r)
-end
+∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric}, r::AbstractVector) = copy(r)
+∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric}, r::AbstractVector) = h.metric.M⁻¹ .* r
+∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric}, r::AbstractVector) = h.metric.M⁻¹ * r
 
-function ∂H∂r(h::Hamiltonian{T,DiagEuclideanMetric{T,A},F1,F2}, r::A) where {T<:Real,A<:AbstractVector{T},F1,F2}
-    return h.metric.M⁻¹ .* r
-end
-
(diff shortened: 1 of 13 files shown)


src/sampler.jl
```diff
@@ -4,31 +4,44 @@ function mh_accept(rng::AbstractRNG, H::AbstractFloat, H_new::AbstractFloat)
...
-        θs[i], Hs[i], αs[i] = step(h, prop, i == 1 ? θ : θs[i-1])
+        θs[i], Hs[i], αs[i] = step(rng, h, prop, i == 1 ? θ : θs[i-1])
         if i <= n_adapts
             adapt!(adaptor, θs[i], αs[i])
             h, prop = update(h, prop, adaptor)
-            verbose && i == n_adapts && @info "Finished $n_adapts adapation steps" typeof(adaptor) prop.integrator.ϵ h.metric
+            if verbose
+                if i == n_adapts
+                    @info "Finished $n_adapts adapation steps" typeof(adaptor) prop.integrator.ϵ h.metric
+                elseif i % Int(n_adapts / 10) == 0
```
yebai: nice!

src/proposal.jl
```diff
@@ -162,8 +162,7 @@ function propose(rng::AbstractRNG, nt::NoUTurnTrajectory{I}, h::Hamiltonian, θ:
     return θ_new, r_new, α / nα
 end
 
-propose(nt::NoUTurnTrajectory{I}, h::Hamiltonian, θ::AbstractVector{T}, r::AbstractVector{T};
-        j_max::Int=10) where {I<:AbstractIntegrator,T<:Real} = propose(GLOBAL_RNG, nt, h, θ, r; j_max=j_max)
+propose(nt::NoUTurnTrajectory{I}, h::Hamiltonian, θ::AbstractVector{T}, r::AbstractVector{T}) where {I<:AbstractIntegrator,T<:Real} = propose(GLOBAL_RNG, nt, h, θ, r)
```
yebai: I think `transition` is more accurate terminology since this function also performs the `MH` step inside. Often, we would expect `proposal` to be a candidate before the `MH` step.

src/proposal.jl
```diff
@@ -21,21 +21,22 @@ abstract type DynamicTrajectory{I<:AbstractIntegrator} <: AbstractHamiltonianTra
...
     integrator  ::  I
+    max_depth   ::  Int
+    Δ_max       ::  AbstractFloat
 end
 
+# Helper function to use default values
+NUTS(integrator::AbstractIntegrator) = NUTS(integrator, 10, 1000.0)
+
 # Create a `NUTS` with a new `ϵ`
 function (snuts::NUTS)(ϵ::AbstractFloat)
```
yebai: This helper's argument is potentially confusing since both `Δ_max` and `ϵ` can be float numbers. What if you want to create a new `NUTS` algorithm, with a different `Δ_max`? Perhaps consider the following

```julia
function (snuts::NUTS)(i::AbstractIntegrator)
   return NUTS(i, snuts.max_depth, snuts.Δ_max)
end
```

as an alternative

src/hamiltonian.jl
```diff
@@ -16,7 +16,8 @@ end
 ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric}, r::AbstractVector) = h.metric.M⁻¹ * r
 
 function hamiltonian_energy(h::Hamiltonian, θ::AbstractVector, r::AbstractVector)
-    return kinetic_energy(h, r, θ) + potential_energy(h, θ)
+    H = kinetic_energy(h, r, θ) + potential_energy(h, θ)
+    return isnan(H) ? Inf : H
```
yebai: nice catch! can we throw a warning message here?

yebai:
Good work - the code looks correct to me. I only left a few minor terminology comments above.

src/proposal.jl
xukai92: Yes I agree. Those functions were actual proposal before (e.g. MH was outside the one for HMC) but I moved the MH step inside to keep interface more consistent as there is no MH step for the one for NUTS. Let me rename them :)

xukai92:
Please don't merge for now. I'm playing with this branch on a diff. eq. model and to see if it works stably.

src/proposal.jl
xukai92: I will change as suggested. But we will probably change this helper interface (https://github.com/TuringLang/AdvancedHMC.jl/issues/45).

xukai92:
All good now. Will merge it tomorrow conditioned on CI passes and no objection.

Issue #17, opened 2019-03-06
Consider introduce the `PhasePoint` type


xukai92:

yebai:
Do you mean `(θ, r)` (parameter, momentum) by `Point`?

xukai92:
Yes, and possibly have some additional information just like https://github.com/tpapp/DynamicHMC.jl/blob/master/src/buildingblocks.jl#L124

xukai92:
Closed by #51.

Issue #16, opened 2019-03-05
More robust and modular way of detecting divergence 


yebai:
Approximation error of leapfrog integration (i.e. accumulated Hamiltonian energy error) can sometimes explode, for example when the curvature of the current region is very high. This type of approximation error is sometimes called `divergence` [1] since it shifts a leapfrog simulation away from the correct solution. 

In Turing, this type of errors is currently caught a relatively ad-hoc function called `is_valid `, 

https://github.com/TuringLang/AdvancedHMC.jl/blob/734c0fa6d802a852d6b0bff7fc9f70a049a3f367/src/integrator.jl#L7

`is_valid` can catch cases where one or more elements of the parameter vector is either `nan` or `inf`. This has several drawbacks

- it's not general enough for catching all leapfrog approximation errors, e.g. when the parameter vector is valid but Hamiltonian energy is invalid.
- it may be delayed because numerical errors can appear in Hamiltonian energy earlier than in parameter vectors
- it's hard to propagate the exception around, i.e. at the moment we use a heuristic to find the previous valid point before approximation/numerical error happens
   - https://github.com/TuringLang/AdvancedHMC.jl/blob/734c0fa6d802a852d6b0bff7fc9f70a049a3f367/src/integrator.jl#L26
   - https://github.com/TuringLang/AdvancedHMC.jl/blob/734c0fa6d802a852d6b0bff7fc9f70a049a3f367/src/integrator.jl#L45

Therefore, we might want to refactor the current code a bit for a more robust mechanism for handling leapfrog approximation errors. Perhaps we can learn from the `DynamicHMC` implementation: 

https://github.com/tpapp/DynamicHMC.jl/blob/master/src/buildingblocks.jl#L168

[1]: Betancourt, M. (2017). A Conceptual Introduction to Hamiltonian Monte Carlo. arXiv preprint arXiv:1701.02434.

yebai:
Update:

The current NUTS implementation rejects the whole Hamiltonian trajectory when numerical errors happen. In practice, a better way is to reject the subtree that contains the error (e.g. numerical error or divergent Hamiltonian). This will allow us to use part of the trajectory when numerical errors are detected. Keeping valid parts of the trajectory also improves the stability of the dual averaging algorithm, since rejecting a whole trajectory means throwing accept ratio `a` off a cliff suddenly (i.e. setting `a=0` from its current value). I suspect this is the reason for sharp fluctuation of dual averaging for NUTS. 

[...]

PR #51, opened 2019-04-16
Type design for `PhasePoint` and `DualValue`.

yebai:
This PR aims at introducing some additional types that are discussed in https://github.com/TuringLang/AdvancedHMC.jl/issues/16. More specifically, the following types are introduced:

- `PhasePoint`: stores `θ`, `r` and cached Potential energy (and its gradient). https://github.com/TuringLang/AdvancedHMC.jl/issues/17
- `DualValue`: stores log density `logπ(θ)`, and cache its gradient.
- ~~LogDensityFunction: stores `logπ` and its gradient function `∂logπ∂θ`.~~

TODOs
- [x] Refactor numerical error handling code `leapfrog`  using `PhasePoint`
- [x] Remove `DualFunction`?
- [x] Refactor `build_tree` using `PhasePoint`
-  ~~Return more information for each step #59~~
-  ~~Add a `Termination` type.~~


src/hamiltonian.jl
```diff
@@ -1,8 +1,29 @@
 # TODO: add a type for kinetic energy
 
+# The constructor of `LogDensityValue` will check numerical errors in
+#   `value` and `gradient`.  That is `is_valid` will be performed automatically.
+struct LogDensityValue{Tv<:AbstractFloat, Tg<:AbstractVector{Tv}}
+    value::Tv    # Cached value, e.g. logπ(θ).
+    gradient::Tg # Cached gradient, e.g. ∇logπ(θ).
+end
+
+struct LogDensityFunction{Tf<:Function}
```
xukai92: Will this work? I think `f` and `∇f` will have different types.

src/hamiltonian.jl
```diff
@@ -1,8 +1,29 @@
 # TODO: add a type for kinetic energy
 
+# The constructor of `LogDensityValue` will check numerical errors in
+#   `value` and `gradient`.  That is `is_valid` will be performed automatically.
```
xukai92: Nice!

src/hamiltonian.jl
```diff
@@ -1,8 +1,29 @@
...
+struct LogDensityValue{Tv<:AbstractFloat, Tg<:AbstractVector{Tv}}
```
xukai92: Not sure if `LogDensityValue` is a good name:
- We are working on log-joint probability instead of simply `LogDensity`. But I'm OK with it since it's also a log density.
- We are storing both value and gradient here. Why do we only name it with `Value`?

src/hamiltonian.jl
```diff
@@ -1,8 +1,29 @@
...
+    f::Tf
+    ∇f::Tf
+end
+
+# The constructor of `PhasePoint` will check numerical errors in
+#   `θ`, `r` and `h`. That is `is_valid` will be performed automatically.
+struct PhasePoint{T<:AbstractVector, Th<:LogDensityValue}
+    θ::T # position variables / parameters
+    r::T # momentum variables
+    h::Th # cached Hamiltonian energy for the current (θ, r)
```
xukai92: This is confusing. Hamiltonian energy doesn't seem to have a type of `LogDensityValue ` right? Should it just be `AbstractFloat`?

src/hamiltonian.jl
yebai: I'm not sure I understand the issue here. Can you clarify a bit here?

src/hamiltonian.jl
yebai: yeah, perhaps we could use the following 

- `DualValue`: store value and its corresponding gradients
- `DualFunction`: store function and its corresponding gradient function

This echoes the naming scheme of AD packages, so should be self-explanatory.

src/hamiltonian.jl
yebai: There's a typo here, `Hamiltonian` energy should be `Potential` energy.

src/hamiltonian.jl
xukai92: See minimal example below:

```julia
julia> struct A{T<:Function}
       v1::T
       v2::T
       end

julia> f() = 1

julia> A(f, f)
A{typeof(f)}(f, f)

julia> g() = :b
g (generic function with 1 method)

julia> A(f, g)
ERROR: MethodError: no method matching A(::typeof(f), ::typeof(g))
Closest candidates are:
  A(::T<:Function, ::T<:Function) where T<:Function at REPL[1]:2
Stacktrace:
 [1] top-level scope at none:0
```

xukai92:
I updated the version in `Project.toml` from `v0.1.2` to `v0.1.3`. You may want to merge with the master again.

xukai92:
Do you plan to address the problem of returning `H_new` for `NUTS` in this PR?

yebai:
> Do you plan to address the problem of returning H_new for NUTS in this PR?

yes, this will be addressed in this PR since `PhasePoint` caches `H` and its gradients.

xukai92:
So for the purpose of https://github.com/TuringLang/AdvancedHMC.jl/issues/59, maybe we should consider here somehow make it flexible to return more information within a single step?

xukai92:
Hi Hong, when can I do a review?

yebai:
It's ready for a look!

src/hamiltonian.jl
```diff
@@ -15,24 +16,53 @@ end
...
+
+Base.isfinite(v::DualValue) = all(isfinite, v.value) && all(isfinite, v.gradient)
+Base.isfinite(v::AbstractVector) = all(isfinite, v)
+Base.isfinite(z::PhasePoint) = isfinite(z.logπ) && isfinite(z.logκ)
+
+###
+### Negative energy (or log probability) functions.
+###
+
+neg_energy(z::PhasePoint) = - z.logπ.value - z.logκ.value
```
xukai92: Why do we change the name here? I think (potential) energy usually refers to `-z.logπ.value` only (as what is is said in the comments above but what this computes is the Hamiltonian. Maybe we could call it "total_energy" if you don't like the old "hamiltonian_energy".

test/hamiltonian.jl
```diff
@@ -1,4 +1,5 @@
 using Test, AdvancedHMC
+import AdvancedHMC:DualValue, PhasePoint
```
xukai92: Please add a space after `:`.

test/hamiltonian.jl
```diff
@@ -7,3 +8,9 @@ h = Hamiltonian(UnitEuclideanMetric(D), logπ, ∂logπ∂θ)
 r_init = ones(D)
 
 @test AdvancedHMC.kinetic_energy(h, r_init, θ_init) == D / 2
+
+z1 = PhasePoint([NaN], [NaN], DualValue(0.,[0.]), DualValue(0.,[0.]))
+z2 = PhasePoint([Inf], [Inf], DualValue(0.,[0.]), DualValue(0.,[0.]))
+
+@test z1.logπ.value == z1.logπ.value
+@test z1.logκ.value == z1.logκ.value
```
xukai92: We could add some tests for constructions that would throw errors?

xukai92:
I commented on some places for discussion. Please take a look!

src/hamiltonian.jl
yebai: I plan to unify the interface for all energy functions (e.g. potential_energy, kinetic_energy) to `negative_energy`. This requires some further work, e.g. creating a type for `kinect` energy.

src/sampler.jl
```diff
@@ -33,36 +55,24 @@ function sample(rng::AbstractRNG,
...
+                    @info "Adapting $i of $n_adapts steps" typeof(adaptor) τ.integrator.ϵ h.metric
                 end
             end
         end
+        z = rand_momentum(rng, z, h)
     end
-    verbose && @info "Finished $n_samples sampling steps in $time (s)" typeof(h.metric) typeof(prop) EBFMI(Hs) mean(αs)
+    verbose && @info "Finished $n_samples sampling steps in $time (s)" typeof(h.metric) typeof(τ) EBFMI(Hs) mean(αs)
     return θs
 end
```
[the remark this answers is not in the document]
yebai: cool.

src/trajectory.jl
```diff
@@ -20,29 +22,31 @@ struct StaticTrajectory{I<:AbstractIntegrator} <: AbstractTrajectory{I}
     n_steps     ::  Int
 end
 
+"""
+Termination (i.e. no-U-turn).
+"""
+struct Termination end
```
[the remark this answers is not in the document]
yebai: There is no need to create a new type for numerical errors (since `PhasePoint` can handle it), but we still need a type for other types of terminations, e.g. no-u-turn, maximal tree-depth, and maximal allowed hamiltonian error during leapfrog.

test/hamiltonian.jl
yebai: These test cases already test `NaN` and `Inf` values for Kinect and potential energy. `PhasePoint` constructors will throw a warning by default; then we have to detect and handle errors explicitly by calling `isfinite(z::PhasePoint)`.

src/hamiltonian.jl
```diff
@@ -15,24 +16,53 @@ end
...
+
+# TODO: replace logπ and logκ with ℓπ, ℓκ??
+# The constructor of `PhasePoint` will check numerical errors in
+#   `θ`, `r` and `h`. That is `is_valid` will be performed automatically.
+struct PhasePoint{T<:AbstractVector, V<:AbstractFloat}
+    θ::T # position variables / parameters
+    r::T # momentum variables
+    logπ::V # cached potential energy for the current θ
+    logκ::V # cached kinect energy for the current r
+    function PhasePoint(θ::T, r::T, ℓπ::V, ℓκ::V) where {T,V}
```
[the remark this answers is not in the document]
yebai: Will fix this.

src/hamiltonian.jl
```diff
@@ -15,24 +16,53 @@ end
...
-        @warn "Kinetic energy is `NaN` and is set to `Inf`."
-    end
-    V = potential_energy(h, θ)
-    if isnan(V)
-        V = Inf
-        @warn "Potential energy is `NaN` and is set to `Inf`."
+
+# The constructor of `DualValue` will check numerical errors in
+#   `value` and `gradient`.  That is `is_valid` will be performed automatically.
+struct DualValue{Tv<:AbstractFloat, Tg<:AbstractVector{Tv}} <: AbstractFloat
```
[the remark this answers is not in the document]
yebai: I found the terminolgoy `ValueGradient` a bit confusing, it sounds like some special `gradient`...

yebai:
Ready for another look! @xukai92

src/hamiltonian.jl
yebai: I changed all `*_energy` functions to `neg_energy`. Please let me know if this works for you.

xukai92: So the design is intended to do
`neg_energy(z::PhasePoint)` => potential + kinetic
`neg_energy(z::Potential)` => potential energy
`neg_energy(z::Kinetic)` => kinetic energy
?

Sound like a good design to me!

test/hamiltonian.jl
xukai92: I simply mean testing these warnings explicitly as they are quite important. I added it in https://github.com/TuringLang/AdvancedHMC.jl/pull/51/commits/0d4aed334c70627ce4f5ec8dc5da9c26691c341a

src/hamiltonian.jl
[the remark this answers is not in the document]
xukai92: I agree. Ideally we also want something more specific and informative than `DualValue`. How about `ValueGradientTuple`?

xukai92:
Looks great to me! Just wonder what's "bugfix" in the commit message of 435ac08 mean? I didn't see bugfix?

yebai:
> Looks great to me! Just wonder what's "bugfix" in the commit message of 435ac08 mean? I didn't see bugfix?

It refers to the leftover `logπ` (should be `ℓπ`).

src/hamiltonian.jl
yebai: yes

src/hamiltonian.jl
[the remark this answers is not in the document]
yebai: Not sure - we can always change this type name later if we got a better idea.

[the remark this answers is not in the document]
xukai92: Yes. Let's keep it as `DualValue` for now.

yebai:
Ready to merge from my side @xukai92

xukai92:
Merged! Thanks for your great work!

Issue #34, opened 2019-03-28
Simplifying `Metric` types by using `AbstractPDMat`


yebai:
The following 3 types in `src/metric.jl`:

- https://github.com/TuringLang/AdvancedHMC.jl/blob/a0af7ccc668a2a1e537dc5acdbe8c7bd22f11a86/src/metric.jl#L16-L24

- https://github.com/TuringLang/AdvancedHMC.jl/blob/a0af7ccc668a2a1e537dc5acdbe8c7bd22f11a86/src/metric.jl#L41-L49

- https://github.com/TuringLang/AdvancedHMC.jl/blob/a0af7ccc668a2a1e537dc5acdbe8c7bd22f11a86/src/metric.jl#L3-L5

can be merged into one type using `AbstractPDMat`

```julia

struct EuclideanMetric{T<:Real,A<:AbstractPDMat{T}} <: AbstractMetric{T}
    M⁻¹     ::  A # contains dim, M⁻¹ and its Cholesky decomposition. 
    # Pre-allocation for intermediate variables
    _temp   ::  A
end
```

Reference: https://github.com/JuliaStats/PDMats.jl

xukai92:
Got it!

yebai:
Also, the `Metric._temp` can be removed if we introduce a `PhasePoint` type (see also https://github.com/TuringLang/AdvancedHMC.jl/issues/17):

```julia
z2::PhasePoint = rand_momentum(rng::AbstractRNG, h::Hamiltonian, z::PhasePoint)
```

Since `rand_momentum` can perform in-place update for `z::PhasePoint`.  However, I'm not sure whether this is necessary since moment variables are not updated as frequent as position variables.

xukai92:
@yebai There is some werid design in PDMats.jl that is not consistent with AHMC. Basically `PDiagMat` will take a vector and compute the `inv` it - not `sqrt` it, which is what we need in AHMC.

yebai:
> @yebai There is some werid design in PDMats.jl that is not consistent with AHMC. Basically PDiagMat will take a vector and compute the inv it - not sqrt it, which is what we need in AHMC.

Can we create a wrapper type for `AbstractPDMat`? For example

```julia

type Metric{T<:Real,M<:AbstractPDMat{T},V<:AbstractVector{T}} <: AbstractPDMat{T}
 m::M
 sqrt_diag::V                     # the vector of diagonal elements
end
```

xukai92:
[...]

PR #61, opened 2019-05-02
Constructing pre-conditioning matrix adaptor with only metric type

yebai:
This PR adds support for constructing pre-conditioning matrix adaptor with only metric type (i.e. diag, unit and dense). 

```julia
pc1 = Preconditioner(:UnitEuclideanMetric) # default dim = 2
```

The adaptor's internal Welford estimator dimension is increased when the first sample is added. 

```julia
# Var adaptor dimention should be increased to length(θ) from 2
θ = [0.0, 0.0, 0.0, 0.0]
AdvancedHMC.adapt!(pc1, θ, 1.)
```
This is required for resolving some adaptor simplification in https://github.com/TuringLang/Turing.jl/pull/739


src/adaptation/precond.jl
```diff
@@ -111,72 +117,185 @@ function get_cov(wc::WelfordCov{T})::Matrix{T} where {T<:AbstractFloat}
...
 
-reset!(dpc::DensePreConditioner) = reset!(dpc.ce)
+reset!(dpc::DensePreconditioner) = reset!(dpc.ce)
 
-function getM⁻¹(dpc::DensePreConditioner)
+function getM⁻¹(dpc::DensePreconditioner)
     return dpc.covar
 end
+
+####
```
xukai92: Why do we move the code here? Does it make AHMC dependent on Adaptation then?

xukai92: I think ideally we would like AHMC without the Adaptation model will work as well.

src/adaptation/precond.jl
```diff
@@ -111,72 +117,185 @@ function get_cov(wc::WelfordCov{T})::Matrix{T} where {T<:AbstractFloat}
...
+
+function Preconditioner(m::DiagEuclideanMetric)
+    return DiagPreconditioner(m.dim)
+end
+
+function Preconditioner(m::DenseEuclideanMetric)
+    return DensePreconditioner(m.dim)
+end
+
+function Preconditioner(m::Type, dim::Integer=2)
```
xukai92: Why the default is 2? If we want to update is anyway can we just set is as 0 or something

yebai: yeah, it's arbitrary; `0` might work, or might cause an issue in a co-variance matrix.

src/adaptation/precond.jl
yebai: I think we can replace `Metric` types with `PDMat` (see https://github.com/TuringLang/AdvancedHMC.jl/issues/34), so this doesn't matter in the future.

yebai: Also, AHMC depends on Adaption anyway: some core features like step-size adaption is impossible without Adaption.

xukai92: But shouldn't a call of `sample()` without adaptation is expected to work?

yebai: I see your point - but I don't think this matters a lot.  Besides, so this dependency will go away if we fix #34, right?

xukai92: Yes that's right! I'm fine with the current code then.

yebai:
I'm still working on the `Turing` side of this PR. Please do not merge yet.

yebai:
@xukai92 This is ready for merge from my side.

src/adaptation/stan_adaption.jl
```diff
@@ -16,14 +16,14 @@ end
 # Acknowledgement: this adaption settings is mimicing Stan's 3-phase adaptation.
 struct StanNUTSAdaptor <: AbstractCompositeAdaptor
     n_adapts    :: Int
-    pc          :: AbstractPreConditioner
+    pc          :: AbstractPreconditioner
```
mohamed82008: Make a type parameter?

PR #65, opened 2019-05-21
Support progress meter

xukai92:
- Add progress meter (https://github.com/TuringLang/AdvancedHMC.jl/issues/57)
  - Also add flag to turn on or off the progress meter. This allows AHMC to run in silent mode, which is very necessary for large scale parallel use of the sampler. FYI, a friend of mine has been using AHMC on a quite large dataset. He runs the sampler in parallel using 40 cores for > 2 hours, and so far it seems to be all good. He also checked the inference results to that he obtained from his VAE implemented and found they are very consistent.
- Remove old intermediate `@info` for adaptation
- Merged `sample()` for adapted and non-adaptive `sample()`

### DIFF
--- a/src/sampler.jl
+++ b/src/sampler.jl
@@ -6,38 +6,8 @@ sample(
     h::Hamiltonian,
     τ::AbstractProposal,
     θ::AbstractVector{T},
-    n_samples::Int;
-    verbose::Bool=true
-) where {T<:Real} = sample(GLOBAL_RNG, h, τ, θ, n_samples; verbose=verbose)
-
-function sample(
-    rng::AbstractRNG,
-    h::Hamiltonian,
-    τ::AbstractProposal,
-    θ::AbstractVector{T},
-    n_samples::Int;
-    verbose::Bool=true
-) where {T<:Real}
-    θs = Vector{Vector{T}}(undef, n_samples)
-    Hs = Vector{T}(undef, n_samples)
-    αs = Vector{T}(undef, n_samples)
-    r = rand(rng, h.metric)
-    z = phasepoint(h, θ, r)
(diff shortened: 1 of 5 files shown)


src/sampler.jl
```diff
@@ -37,40 +37,43 @@ sample(
     τ::AbstractProposal,
     θ::AbstractVector{T},
     n_samples::Int,
-    adaptor::Adaptation.AbstractAdaptor,
+    adaptor::Union{Nothing,Adaptation.AbstractAdaptor}=nothing,
```
yebai: Can we define a special `NoAdaptation <: AbstractAdaptor ` instead of using `nothing`?

src/sampler.jl
```diff
@@ -37,40 +37,43 @@ sample(
...
     n_adapts::Int=min(div(n_samples, 10), 1_000);
-    verbose::Bool=true
-) where {T<:Real} = sample(GLOBAL_RNG, h, τ, θ, n_samples, adaptor, n_adapts; verbose=verbose)
+    verbose::Bool=true,
+    progress::Bool=false
+) where {T<:Real} = sample(GLOBAL_RNG, h, τ, θ, n_samples, adaptor, n_adapts; verbose=verbose, progress=progress)
 
 function sample(
```
yebai: Nice to unify `sample` interfaces.

src/sampler.jl
xukai92: Added in https://github.com/TuringLang/AdvancedHMC.jl/pull/65/commits/a0e1a9f770b38b7bee57a10aab17be950c065ea6

yebai:
Looks nice, many thanks! @xukai92

PR #67, opened 2019-05-22
Fix adaptation for HMCDA

xukai92:
- Add `finalize!` to all adaptors
  - HMCDA shall use `NaiveCompositeAdaptor` instead of Stan's windowed.
- Fix a bug in progress meter for which it only shows adaptation stats during adaptation but not afterwards.
- Small clean-ups in step size adaptation (no functional change).

### DIFF
--- a/src/adaptation/Adaptation.jl
+++ b/src/adaptation/Adaptation.jl
@@ -7,6 +7,7 @@ import LinearAlgebra, Statistics
 using ..AdvancedHMC: DEBUG
 
 abstract type AbstractAdaptor end
+finalize!(::AbstractAdaptor) = nothing
 struct NoAdaptation <: AbstractAdaptor end
 
 include("stepsize.jl")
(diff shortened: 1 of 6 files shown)


src/adaptation/Adaptation.jl
```diff
@@ -7,6 +7,7 @@ import LinearAlgebra, Statistics
 using ..AdvancedHMC: DEBUG
 
 abstract type AbstractAdaptor end
+finalize!(::AbstractAdaptor) = nothing
```
yebai: Consider replacing this line with a generic function, e.g.

```julia
function finalize!(::AbstractAdaptor)  end
```

src/sampler.jl
```diff
@@ -36,11 +36,18 @@ function sample(
...
         progress && (showvalues = Tuple[(:iteration, i), (:hamiltonian_energy, Hs[i]), (:acceptance_rate, αs[i])])
-        if !(adaptor isa Adaptation.NoAdaptation) && i <= n_adapts
-            adapt!(adaptor, θs[i], αs[i])
-            h, τ = update(h, τ, adaptor)
+        if !(adaptor isa Adaptation.NoAdaptation)
+            if i <= n_adapts
+                adapt!(adaptor, θs[i], αs[i])
+                h, τ = update(h, τ, adaptor)
+                # Finalize adapation
+                if i == n_adapts
```
yebai: Nice!

src/adaptation/Adaptation.jl
xukai92: I'm following this: https://github.com/jrevels/YASGuide
> For single-line method definitions, use the form f(args...) = ....

> If a function does not have a clearly appropriate return value, then explicitly return nothing.

The first one is consistent to ours:

> Only use short-form function definitions when they fit on a single line:

yebai: Ah, I mean can we try to define a generic function (interface) for abstract types, instead of just return `nothing`? That is,

```julia
finalize!(::AbstractAdaptor) = nothing
```

===>

```julia
function finalize!(::AbstractAdaptor)  end 
```

xukai92: I don't think there is a generic one for `AbstractAdaptor`. E.g. for preconditioner there is not `finalize!` needed but for step size adaptor there is. And thus for composite adaptor we should do `finalize!` to the step size component.

yebai: Since `function finalize!(::AbstractAdaptor)  end` is a valid Julia function, so we can use it as a "generic" function. For more details, you can take a look at https://pkg.julialang.org/docs/julia/THl1k/1.1.0/manual/methods.html#Empty-generic-functions-1, although it's not entirely the same.

xukai92: I'm confused. So you mean it's more about a styling suggestion?

xukai92: I understand that we can do `function finalize!(::AbstractAdaptor)  end` as well but I kind of prefer the idea that 

> If a function does not have a clearly appropriate return value, then explicitly return nothing.

yebai: I just tried this myself, tt turns out that 

```julia
function finalize!(::AbstractAdaptor) end
```

produces the same output (ie. `nothing`) as your code

finalize!(::AbstractAdaptor) = nothing

Sorry for the confusion.

PR #68, opened 2019-05-22
Move rejection to MH

xukai92:


xukai92:
@yebai Seems sampling doing so will allow MH to reject samples with numerical issue? The drawback is that we don't use the "partial numerical OK" trajectory any more.

xukai92:
I would prefer leave this optimisation aside for the moment.

yebai:
> @yebai Seems sampling doing so will allow MH to reject samples with numerical issue? The drawback is that we don't use the "partial numerical OK" trajectory any more.

Correct - this PR will reject trajectories with numerical issues as a whole.

PR #69, opened 2019-05-28
[WIP] New design for `Trajectory` and `Termination` types.

yebai:
This PR aims at a modular design for `Trajectory` types by separating the stopping criteria for HMC (e.g.  No-U-Turn, a fixed number of leapfrog steps, Hamiltonian divergence threshold) into a separate type called `Termination`. Then the `transition` function can be overloaded for different combinations of `Trajectory` and `Termination`:

https://github.com/TuringLang/AdvancedHMC.jl/blob/9f00187e8ab453c7bbed4a455b05cc36a22e17c4/src/trajectory.jl#L76-L81.

This design allows us to use the same `Trajectory` type for most HMC samplers.  In addition, this PR also try to make the `Trajectory` more general by including more trajectory-specific information, e.g. 

https://github.com/TuringLang/AdvancedHMC.jl/blob/9f00187e8ab453c7bbed4a455b05cc36a22e17c4/src/trajectory.jl#L57-L63

This is still a draft PR, so any comments are appreciated.

### DIFF
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -2,6 +2,7 @@
 ## Hamiltonian dynamics numerical simulation trajectories
 ##
 
+
 abstract type AbstractProposal end
 abstract type AbstractTrajectory{I<:AbstractIntegrator} <: AbstractProposal end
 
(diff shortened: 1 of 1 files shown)


cpfiffer:
These are all still going to return the `PhasePoint` type, right?

yebai:
> These are all still going to return the PhasePoint type, right?

I'm planning to let `transition` to return a `Trajectory` type; this allows us to separate code for combing sub-trajectories (e.g. those created by `build_tree`). Of course, `Trajectory` will contain an instance of the `PhasePoint` type inside its definition.

yebai:
@cpfiffer The short answer is yes; it will still return `PhasePoint`.

xukai92:
Update from this branch/PR will include all changes of merging master as well. I will start a new branch for this reason.

PR #74, opened 2019-06-10
Improving the performance and generality of adaptation

mohamed82008:
This PR:
1. Enables full-fledged arbitrary precision handling, e.g. Float64 or Float32, 
2. Resolves some type instability issues from using abstract type fields, and
3. Reduces allocations in a few places.

### DIFF
--- a/src/adaptation/precond.jl
+++ b/src/adaptation/precond.jl
@@ -16,8 +16,9 @@ mutable struct NaiveVar{T} <: VarEstimator{T}
     S :: Vector{Vector{T}}
 end
 
-NaiveVar() = NaiveVar(0, Vector{Vector{Float64}}())
-NaiveVar(::Int) = NaiveVar()
+NaiveVar(::Type{T}=Float64) where {T} = NaiveVar(0, Vector{Vector{T}}())
+NaiveVar(::Type{T}, ::Int) where {T} = NaiveVar(T)
+NaiveVar(::Int) = NaiveVar(Float64)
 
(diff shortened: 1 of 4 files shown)


mohamed82008:
@xukai92 there is no functional change in this PR AFAIK, but the tests are failing. Any idea why?

mohamed82008:
It seems that tests pass now!

src/adaptation/precond.jl
```diff
@@ -29,38 +30,42 @@ function reset!(nc::NaiveVar{T}) where {T<:AbstractFloat}
...
 function add_sample!(wv::WelfordVar, s::AbstractVector)
     wv.n += 1
-    δ = s .- wv.μ
-    wv.μ .+= δ ./ wv.n
-    wv.M .+= δ .* (s .- wv.μ)
+    @unpack μ, M, n = wv
+    for i in eachindex(s)
+        δ = s[i] - μ[i]
+        μ[i] += δ / n
+        M[i] += δ * (s[i] - μ[i])
```
yebai: Why switching from broadcasting to loop? Is this for readability or for efficiency?

mohamed82008: Efficiency to avoid allocating `δ`.

mohamed82008:
`Float64` is the default even on 32-bit Julia, that is `typeof(1.0) === Float64` on both 32-bit and 64-bit Julia.

yebai:
Thanks!

PR #77, opened 2019-06-30
Fixed bug preventing StanHMCAdaptor from adapting

chriselrod:
StanHMCAdaptor was using positional arguments, but all the relevant signatures had keyword arguments. There was a generic fallback signature with positional arguments that was just a `nothing` stub, and it was always called instead. Turning the keyword args into positional args lets them be called instead, so that `StanHMCAdaptor` will now in fact adapt.

I'm not sure if it's wise to actually have that fallback. I'd think an error is preferable.


xukai92:
Great! Thanks for catching this!

Issue #75, opened 2019-06-11
Get more comments and docstrings


xukai92:

xukai92:
Closed by PR #79.

Issue #71, opened 2019-05-29
Add more informative error messages.


trappmartin:
I just came across the error message at: 
https://github.com/TuringLang/AdvancedHMC.jl/blob/0814ae16a9f550bda9e4b3be5360b519bac5c144/src/hamiltonian.jl#L33

This error message should be more informative so that the user gets an idea why he gets it, i.e. in my case I suppose the model I'm looking at is misspecified. It would be good to know what is  `Inf`.

xukai92:
I don't think the error message could be informative to that level, but I can at least make it indicate in which part (position, mometum, potential energy or its gradient, kinect energy or itsgradient) the numerical issue occurs.

xukai92:
Closed by PR #79.

Issue #62, opened 2018-07-08
HMC: Multinomial Sampling Instead of Slice Sampling


wesselb:
As mentioned on page 59 in [1], multinomial sampling instead of slice sampling may provide an improvement in overall performance.

[1] https://arxiv.org/abs/1701.02434

xukai92:
@yebai Do you think it make more sense for us to implement this first so that our algorithm is the same as Stan and DHMC before we try to find a possible bug?

yebai:
It sounds good to me, although I suspect that the bug is in the `build_tree` part of the code. Maybe we can improve the code a bit by implementing the `multinomial` sampling scheme.

xukai92:
Closed by PR #79.

Issue #59, opened 2019-04-21
Return more information for each step


xukai92:
- `\alpha`
- `H`
- Leapfrog step size
- Number of leapfrog steps
- Number of value evals
- Number of grad evals

PR #79, opened 2019-07-06
[WIP] Support multinomial sampling for NUTS, and more

xukai92:
This PR resolves / will resolve:
- [x] HMC: Multinomial Sampling Instead of Slice Sampling #62
- [x] Get more comments and docstrings #75
- ~~More robust and modular way of detecting divergence #16~~
- ~~Return more information for each step #59~~
- [x] Add more informative error messages. #71
- ~~Identify divergent transition #50~~
- [x] RFC `find_good_eps` #27
- [x] Update gradient function to return both value and gradient #56


src/trajectory.jl
```diff
@@ -13,6 +13,70 @@ transition(
...
+
+struct Trajectory{I<:AbstractIntegrator} <: AbstractTrajectory{I}
+    integrator :: I
+    n_steps :: Int # Counter for total leapfrog steps already applied.
+    Δ :: AbstractFloat # Current hamiltonian energy minus starting hamiltonian energy
+    # TODO: replace all ``*Trajectory` types with `Trajectory`.
+    # TODO: add turn statistic, divergent statistic, proposal statistic
+end
+
+isterminated(
```
xukai92: How this function is supposed to use? I mean we can't use it in our HMC implementation as we don't build state one by one but run the integrator with the designated numbers right?

xukai92:
@yebai Please take a look over the changes. So far I didn't change the `transition` interface (which means I don't touch the previous `AbstractTrajectory` stuff) yet. The only main change I did is introducing a tree struct called `FullBinaryTree` that contains a `sampler::AbstractTreeSampler`. With the `FullBinaryTree` I refactored the `build_tree` function with two tree sampling methods (slic and multinomial dispatched by the sampler.

src/trajectory.jl
```diff
@@ -82,121 +141,200 @@ end
...
+
+mh_accept(rng::AbstractRNG, s::SliceNUTSSampler, s′::SliceNUTSSampler) = mh_accept(rng, s.n, s′.n)
+mh_accept(rng::AbstractRNG, s::MultinomialNUTSSampler, s′::MultinomialNUTSSampler) = mh_accept(rng, s.w, s′.w)
+mh_accept(s::AbstractNUTSSampler, s′::AbstractNUTSSampler) = mh_accept(GLOBAL_RNG, s, s′)
+
+InitSampler(rng::AbstractRNG, ::SliceNUTSSampling, H::AbstractFloat) = SliceNUTSSampler(log(rand(rng)) - H, 1)
+InitSampler(rng::AbstractRNG, ::MultinomialNUTSSampling, H::AbstractFloat) = MultinomialNUTSSampler(H)
+InitSampler(s::AbstractNUTSSampling, H::AbstractFloat) = InitSampler(GLOBAL_RNG, s, H)
 
 function transition(
```
xukai92: @yebai So here I didn't use `FullBinaryTree` to initialize the starting point. 
- The main reason being the different sampling steps: in `build_tree` sampling is done by the sampler while here is done by a MH step. 
- Additionally the conditon to do the sampling are different: here MH step is only done if the proposed tree is not terminated while that in `build_tree` is performed regardless (maybe this is something we can optimize?). 
- For this two reasons I didn't find a good way to unify the `merge` so that it can also be used here.
- Also I don't want to return the tree anyway so there might be less necessity to do so.

xukai92:
Just check the code is still type stable.

yebai:
> Please take a look over the changes. So far I didn't change the transition interface (which means I don't touch the previous AbstractTrajectory stuff) yet. The only main change I did is introducing a tree struct called FullBinaryTree that contains a sampler::AbstractTreeSampler. With the FullBinaryTree I refactored the build_tree function with two tree sampling methods (slic and multinomial dispatched by the sampler.

Thanks, Kai. Below are some issues I found so far. 

The current PR contains 2 types of trajectory sampling method (slice/multinomial). 

- `TreeSampling` 
- `TreeSampler`

These 2 types are very similar. I think we can unify them to slightly reduce clutter and improve readability.  Also, the following field names can be unified since they both indicate a trajectory sampling method (i.e., slice/multinomial) 

- `NUTS.sampling`  # maybe change this to sampler
- `FullBinaryTree.sampler`

xukai92:
The reason I have `NUTS.sampling` is I don't want to initalize any sampler (which has some parameters based on the actual methods) when constructing `NUTS`.

One solution is to change `NUTS.sampling` to `NUTS.samplerType::Type{T} where {T<:TreeSampler}`. How do you think?

yebai:
> One solution is to change NUTS.sampling to NUTS.samplerType::Type{T} where {T<:TreeSampler}. How do you think?

Yeah,  that could work.

xukai92:
@yebai Also did you see my comments starting with `REVIEW` in the code? Those are thing that needs double check for the algorithm correctness.

xukai92:
@yebai 

Breaking changes

- `∂ℓπ∂θ` is now assumed to retunr a `(value, graident)` tuple
- `∂H∂θ` now returns `(logprob, -∂ℓπ∂θ)` tuple

Almost 2x speed-up by using the cache.

src/hamiltonian.jl
```diff
@@ -42,8 +46,8 @@ phasepoint(
     h::Hamiltonian,
     θ::T,
     r::T;
-    ℓπ = DualValue(neg_energy(h, r, θ), ∂H∂θ(h, θ)),
-    ℓκ = DualValue(neg_energy(h, θ), ∂H∂r(h, r))
+    ℓπ=DualValue(∂H∂θ(h, θ)...),
+    ℓκ=DualValue(neg_energy(h, r, θ), ∂H∂r(h, r))
```
xukai92: @yebai Caught a bug here. Luckily the only usecase is to take a sum of these two values.

src/integrator.jl
```diff
@@ -25,14 +25,14 @@ function step(
...
+    _, ∇θ = ∂H∂θ(h, θ)
     for i = 1:abs(n_steps)
         r = r - ϵ/2 * ∇θ # Take a half leapfrog step for momentum variable
         ∇r = ∂H∂r(h, r)
         θ = θ + ϵ * ∇r   # Take a full leapfrog step for position variable
-        ∇θ = ∂H∂θ(h, θ)
+        logprob, ∇θ = ∂H∂θ(h, θ)
         r = r - ϵ/2 * ∇θ # Take a half leapfrog step for momentum variable
-        z = phasepoint(h, θ, r)
+        z = phasepoint(h, θ, r; ℓπ=DualValue(logprob, ∇θ))
```
xukai92: Acutal use of the cache!

src/hamiltonian.jl
yebai: nice catch!

yebai:
> @yebai Also did you see my comments starting with REVIEW in the code? Those are thing that needs double check for the algorithm correctness.

I checked - it looks correct!

src/trajectory.jl
yebai: Let's leave this as is - I'll fix it in a separate PR.

yebai:
We need to update the example in `README` (and probably Turing at some point) since `∂ℓπ∂θ` needs to return both value and gradient now.

xukai92:
Sure. I will update it.

xukai92:
Just merged it. Thanks for the help! @yebai 

I will finish the rest issues for v0.2 in another PR in the next few days. Those are mostly interface changes and RFC so I hope it won't take as long as this one.

PR #82, opened 2019-07-24
Bugfix for the use of metric

xukai92:
Our code for how to use the metric (aka M_inverse) was based on Stan's. For a long time, we found our diag and dense preconditioning behaves weridly. This is still true for the current master and I meet this issue when running integrated test on Turing.jl again today, even in gdemo.

After checking Section 4.2.1 of Michael's paper again, I don't see any clue how would Stan's way to use M_inverse would work (at least in our current setting). So I changed it based on my current understanding and at least it works normally on gdemo now.

Below is what I did:
- When sample momentum, I return the momentum variable in the transformed space.
  - The transformation is the one in Michael's paper instead of the previous one. The difference is whether we multiply or divide r by M_inverse.
  - I think @yebai was trying to do the same thing before but we ended up following Stan's code.
- When compute grad_r that is needed to update q, I transform grad_r back to its original space.
  - By doing so, the update of q is just as normal.
- Nowhere else need to use M_inverse.
  - The changes in energy are cancelled and we only use Hamiltonian as a sum of two energy.
    - End of page 30 of https://arxiv.org/pdf/1701.02434.pdf is the one for knetic energy, and for potential energy it's just the origional - log(abs(det(jacob)), the latter of which is cancelled.
  - The logabsdetjacob in potential space doesn't matter as it's a constant w.r.t potential variable thus will be 0 when we take the gradient.

### DIFF
--- a/src/AdvancedHMC.jl
+++ b/src/AdvancedHMC.jl
@@ -3,7 +3,7 @@ module AdvancedHMC
 const DEBUG = Bool(parse(Int, get(ENV, "DEBUG_AHMC", "0")))
 
 using Statistics: mean, var, middle
-using LinearAlgebra: Symmetric, UpperTriangular, mul!, ldiv!, dot, I, diag, cholesky
+using LinearAlgebra: Symmetric, UpperTriangular, mul!, dot, I, diag, cholesky
 using LazyArrays: BroadcastArray
 using Random: GLOBAL_RNG, AbstractRNG
 using ProgressMeter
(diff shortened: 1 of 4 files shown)


xukai92:
@yebai Any comments?

xukai92:
Emmm. I asked Michale about why they are different between the code and the paper and he explained to me what happens. The r.v. we sample from `N(0, 1)` is in fact `q'` but not `q`, thus all the multiplication and division are "reversed".

I'm not more confused about this :(

xukai92:
This might be related to the U-Turn criteria. Ref: https://github.com/stan-dev/stan/blob/1bb054027b01326e66ec610e95ef9b2a60aa6bec/src/stan/mcmc/hmc/nuts_classic/diag_e_nuts_classic.hpp#L24 is different.

xukai92:
I don't think this fix is quite right now. I will reopen it or create a new one when I feel more confident on it.

PR #83, opened 2019-07-25
Bugfix for metric renew

xukai92:


src/trajectory.jl
```diff
@@ -508,7 +513,7 @@ function update(
     τ::AbstractProposal,
     pc::Adaptation.AbstractPreconditioner
 ) 
-    metric = reconstruct(h.metric, M⁻¹=getM⁻¹(pc))
```
xukai92: Baiscally before only the new `M⁻¹` was passed to the new `metric` and `sqrtM⁻¹` or `cholM⁻¹` was not even re-computed @yebai.

src/sampler.jl
```diff
@@ -62,12 +62,9 @@ function sample(
...
             if i <= n_adapts
                 adapt!(adaptor, θs[i], αs[i])
-                # Finalize adapation
-                if i == n_adapts
-                    finalize!(adaptor)
-                    (verbose && !progress) && @info "Finished $n_adapts adapation steps" adaptor τ.integrator h.metric
-                end
+                i == n_adapts && finalize!(adaptor)
                 h, τ = update(h, τ, adaptor)
+                (i == n_adapts && verbose && !progress) && @info "Finished $n_adapts adapation steps" adaptor τ.integrator h.metric
```
xukai92: I change the order of `update` and finish-of-adapation info so that we indeed report the number after the last update.

src/trajectory.jl
```diff
@@ -239,6 +239,11 @@ function isturn(h::Hamiltonian, zleft::PhasePoint, zright::PhasePoint)
     s = (dot(θdiff, ∂H∂r(h, zleft.r)) >= 0 ? 1 : 0) * (dot(θdiff, ∂H∂r(h, zright.r)) >= 0 ? 1 : 0)
     return Termination(s == 0, false)
 end
+# function isturn(h::Hamiltonian, zleft::PhasePoint, zright::PhasePoint)
+#     θdiff = zright.θ - zleft.θ
+#     s = (dot(-θdiff, ∂H∂r(h, zleft.r)) < 0) && (dot(θdiff, ∂H∂r(h, zright.r)) < 0)
+#     return Termination(s, false)
+# end
```
xukai92: When debugging I was wondering if our U turn criteria is correct or not. Maybe we don't need to touch it now and we will probably implement the generalized U turn criteria later anyway.

xukai92:
Let me add the gdemo posterior mean test to this repo and then merge it.

yebai:
ok - better to make a new release after this PR is merged.

xukai92:
Sure. Shall I increment to v0.2.1?

xukai92:
I'm happy to have it merged when CI passes.

test/models.jl
```diff
@@ -0,0 +1,24 @@
+using Test, Random, AdvancedHMC
```
xukai92: @yebai I added a new test file call `models.jl` so we could add more typical models in the future.

yebai: nice!

PR #85, opened 2019-07-26
Fix resize

xukai92:


src/adaptation/precond.jl
```diff
@@ -227,7 +227,7 @@ function Base.resize!(
     if length(θ) != length(dpc.var)
         @assert dpc.ve.n == 0 "Cannot resize a var estimator when it contains samples."
         dpc.ve = WelfordVar(T, length(θ))
-        dpc.var = zeros(T, length(θ))
+        dpc.var = ones(T, length(θ))
```
yebai: nice catch!

yebai:
Seems that some tests are broken?

xukai92:
I guess we were testing against a vector of zeros for the function I fixed? I am not with my computer ATM. Can you check and maybe fix the test?

yebai:
Sure - Fixed.

xukai92:
Thanks!

xukai92:
Shall we register it?

PR #93, opened 2019-08-27
Build interface for Turing

xukai92:
Simplified the AHMC call in Turing easy; in favour of https://github.com/TuringLang/Turing.jl/pull/793

### DIFF
--- a/src/sampler.jl
+++ b/src/sampler.jl
@@ -2,17 +2,12 @@
 ## Interface functions
 ##
 
-struct Sample{P<:PhasePoint, NT<:NamedTuple}
-    z       ::  P
-    stat    ::  NT
-end
-
 function sample_init(rng::AbstractRNG, h::Hamiltonian, θ::AbstractVector{T}) where {T<:Real}
     # Ensure h.metric has the same dim as θ.
     h = update(h, θ)
     # Initial transition
-    s = Sample(phasepoint(rng, θ, h), NamedTuple())
-    return h, s
+    t = Transition(phasepoint(rng, θ, h), NamedTuple())
+    return h, t
 end
 
 # A step is a momentum refreshment plus a transition
@@ -20,8 +15,7 @@ function step(rng::AbstractRNG, h::Hamiltonian, τ::AbstractProposal, z::PhasePo
     # Refresh momentum
     z = refresh(rng, z, h)
     # Make transition
-    z, stat = transition(rng, τ, h, z)
-    return Sample(z, stat)
+    return transition(rng, τ, h, z)
 end
 
 adapt!(
@@ -119,23 +113,23 @@ function sample(
     # Prepare containers to store sampling results
     θs, stats = Vector{Vector{T}}(undef, n_samples), Vector{NamedTuple}(undef, n_samples)
     # Initial sampling
-    h, s = sample_init(rng, h, θ)
+    h, t = sample_init(rng, h, θ)
     # Progress meter
     pm = progress ? Progress(n_samples, desc="Sampling", barlen=31) : nothing
     time = @elapsed for i = 1:n_samples
         # Make a step
-        s = step(rng, h, τ, s.z)
+        t = step(rng, h, τ, t.z)
         # Adapt h and τ; what mutable is the adaptor
-        h, τ, isadapted = adapt!(h, τ, adaptor, i, n_adapts, s.z.θ, s.stat.acceptance_rate)
+        h, τ, isadapted = adapt!(h, τ, adaptor, i, n_adapts, t.z.θ, t.stat.acceptance_rate)
         # Update progress meter
         if progress
-            pm_next!(pm, s.stat, i, h.metric)
+            pm_next!(pm, t.stat, i, h.metric)
         # Report finish of adapation
         elseif verbose && isadapted && i == n_adapts
             @info "Finished $n_adapts adapation steps" adaptor τ.integrator h.metric
         end
         # Store sample
-        θs[i], stats[i] = s.z.θ, s.stat
+        θs[i], stats[i] = t.z.θ, t.stat
     end
     # Report end of sampling
     EBFMI_est = EBFMI(map(s -> s.hamiltonian_energy, stats)) 
--- a/src/sampler.jl
+++ b/src/sampler.jl
@@ -2,6 +2,15 @@
 ## Interface functions
 ##
 
+function init(rng::AbstractRNG, h::Hamiltonian, θ::AbstractVector{T}) where {T<:Real}
+    # Ensure h.metric has the same dim as θ.
+    h = update(h, θ)
+    # Prepare phase point for sampling
+    r = rand(rng, h.metric)
+    z = phasepoint(h, θ, r)
+    return h, z
+end
+
 function step(rng::AbstractRNG, h::Hamiltonian, τ::AbstractProposal, z::PhasePoint)
     # Make transition
     z, stat = transition(rng, τ, h, z)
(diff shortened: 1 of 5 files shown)


xukai92:
Basically wrapped functions a bit so that it's easier on the Turing side for the new interface.

And also renamed `precondition` -> `mass_matrix`, which is always reported now with/without adaption.

xukai92:
Please do not merge for now. I realised there is a bug I introduced.

src/sampler.jl
```diff
@@ -1,3 +1,55 @@
+##
+## Interface functions
+##
+
+function init(rng::AbstractRNG, h::Hamiltonian, θ::AbstractVector{T}) where {T<:Real}
```
yebai: Perhaps change `init` into `sample_init`to to match `Interface.jl` API in https://github.com/TuringLang/Turing.jl/pull/793?

cc @cpfiffer Do we have immutable versions of `sample_init!` and `step!` etc? If not, are there any reasons?

src/sampler.jl
```diff
@@ -1,3 +1,55 @@
...
+    # Ensure h.metric has the same dim as θ.
+    h = update(h, θ)
+    # Prepare phase point for sampling
+    r = rand(rng, h.metric)
+    z = phasepoint(h, θ, r)
+    return h, z
+end
+
+function step(rng::AbstractRNG, h::Hamiltonian, τ::AbstractProposal, z::PhasePoint)
```
yebai: Perhaps try to match the function signature with the following if possible?

https://github.com/TuringLang/Turing.jl/blob/csp/interface-pt2/src/interface/Interface.jl#L238

We can probably treat `h::Hamiltonian` as an equivalent to `ModelType`.

src/sampler.jl
```diff
@@ -50,28 +102,25 @@ function sample(
...
-            # Progress info for adapation
-            progress && (showvalues[:step_size] = τ.integrator.ϵ; showvalues[:precondition] = h.metric)
+        z, θs[i], αs[i], Hs[i], stats[i] = step(rng, h, τ, z)
+        # Adapt h and τ; what mutable is the adaptor
+        h, τ, isadapted = adapt!(h, τ, adaptor, i, n_adapts, θs[i], αs[i])
+        # Adapation finish
+        if isadapted && i == n_adapts && verbose && !progress
+            @info "Finished $n_adapts adapation steps" adaptor τ.integrator h.metric
+        end
+        if progress
```
yebai: Perhaps wrap this block of code into a re-usable function, e.g.

```julia
ProgressMeter.next!(pm; v::NamedTuple) == ProgressMeter.next!(pm; showvalues=collect(zip(keys(v), values(v))))
```

then write

```julia
v=(mass_matrix=h.metric, iteration=i, stats...)
progress && ProgressMeter.next!(pm; showvalues=v)
```

src/sampler.jl
xukai92: I guess this is more similar to `transition_init`?

src/sampler.jl
xukai92: What do you mean by the signature here?

src/sampler.jl
xukai92: Good idea!

src/sampler.jl
yebai: I mean function arguments.

src/sampler.jl
yebai: It feels like a sampler level initialization since it is only performed once for a chain. Maybe @cpfiffer can give a better answer here.

src/sampler.jl
xukai92: You mean the `N` and `debug` in the function? 

```julia
function step!(
    rng::AbstractRNG,
    ℓ::ModelType,
    s::SamplerType,
    N::Integer;
    debug::Bool=false,
    kwargs...
) where {ModelType<:Sampleable, SamplerType<:AbstractSampler}
    # Do nothing.
    debug && @warn "No step! function has been implemented for objects of types \n- $(typeof(ℓ)) \n- $(typeof(s))"
end
```

I don't see other parts could be mimiced.

src/sampler.jl
xukai92: We can actually remove this `init` by performing momentum sampling in the beginning of each transition.

src/sampler.jl
xukai92: Done in 964757d.

src/sampler.jl
xukai92: OK we can't, there is still `h = update(h, θ)`

xukai92: `init` is currently used in Turing in the `HMCState` constructor.

I think from the AHMC side or other sampler side, it would be nice to support an immutable way to intergate, right? Also I don't think there is a solution that a `step` implementation in AHMC can be directly used in Turing. We need to wrap it in some way so that things like `Sampler` and `State` works. Is it a correct understanding?

src/sampler.jl
xukai92: I see. It makes sense.

src/sampler.jl
xukai92: > z (I think) contains the state information

Yes. But just to add that it doesn't have all the information needed for adapation. Adapation requires additionally the acceptane rate.

xukai92: So what do you suggest me to do now?

src/sampler.jl
xukai92: So here sampler is actually closer to `τ::AbstractProposal`.

src/sampler.jl
xukai92: By `transition` do you mean `step` or something?

xukai92: Now I get it.

src/sampler.jl
```diff
@@ -53,12 +53,20 @@ function adapt!(
     return h, τ, isadapted
 end
 
-function ProgressMeter.next!(pm, stat::NamedTuple, i::Int, metric)
+"""
+Progress meter update with all trajectory stats, iteration number and metric shown.
+"""
+function pm_next!(pm, stat::NamedTuple, i::Int, metric::AbstractMetric)
```
yebai: Nice trick to use dispatch-based reporting. Perhaps consider 

- renaming `pm_next!` ==> `show_progress!` for clarity? 
- use keyward arguments, instead of positional arguments for `iteration` and `metric`, e.g.

```julia
simple_pm_next!(pm::NoReport, stat::NamedTuple; kargs...) = ProgressMeter.next!(pm)
simple_pm_next!(pm::Progress, stat::NamedTuple; kargs...) = ProgressMeter.next!(pm, (stat...,kargs...))
```

yebai: It will be nice if we can replace all `progress` flags with `r::AbstractReport`, and define a void `NoReport` type for the case `progress=false`. 

@cpfiffer  Does this make sense to be part of `Interface.jl`?

src/sampler.jl
```diff
@@ -1,3 +1,72 @@
+##
+## Interface functions
+##
+
+struct Sample{P<:PhasePoint, NT<:NamedTuple}
```
yebai: Should this function be named `HMCTransition` according to https://github.com/TuringLang/Turing.jl/blob/csp/interface-pt2/src/interface/Interface.jl#L37

src/sampler.jl
yebai: Looks cool! I guess we have to wait until `Interface.jl` is transferred into `MCMCChains.jl` before we can use this `callback` functionality?

src/sampler.jl
xukai92: I think it a bit and feel `Transition` seems to be a better name. Ideally I don't think we would store the momentum variable in a `Sample` - all we need is position. On the other side `Transition` sounds more wider than `Sample`.

xukai92: Maybe in transition we should return a namedtuple `(z=z, stat...)`? How do you think?

yebai: I dont’t have A preference here, although I think a Transition type is more consistent with Interface.jl

xukai92: I see. That's not a bit problem I think. I can create a specific type for that named tuple to store transition. I guess this probably the ideal way (as one of reason to use NamedTuple is just for convenience).

xukai92: OK it's not easy either. If we make a new struct rather than using `NamedTuple` we also need to implment some helper functions like `iterate`. Maybe not worth it for now. 

I will just rename it to `Transition` for now.

xukai92: @cpfiffer I updated this PR and push a corresonding update to TuringLang/Turing.jl#793

xukai92:
@JuliaRegistrator register()

Issue #52, opened 2019-04-16
Tempering with multiplication of momentum leap frog step


xukai92:
As requested by @jonasmac16, it would be nice to have this implemented; see Section 5.7 of Neal, 2012.

jonasmac16:
As suggested by Kai on Slack tempering would need to be added to the 'step' function

>  Secondly, you need to add the tempering steps to (https://github.com/TuringLang/AdvancedHMC.jl/blob/637c2be9b424a09d228657b10ee8fa511a0767b8/src/integrator.jl#L37-L65)

Making sure to that we "heat up" and "cool down" symmetrically as suggested by [Neal, 2012](https://arxiv.org/pdf/1206.1901.pdf). As pointed out by Kai due to nature of the leapfrog implementation taking full step for the momentum within the trajectory we need to take extra care.

An `if elseif` scheme for tempering with the trajectory like this should probably work:
```
n_steps =  5

println("Step:0")
println("half*sqrt(α)")
for i in 1:n_steps
  if i*2 == n_steps && i != n_steps
      #θ, r = θ_new, r_new/α
      println("half*sqrt(α) + half/sqrt(α) = full")
      println("Step:",i,".5")
  elseif 2*i < n_steps
      #θ, r = θ_new, r_new*α
      println("full*α")
      println("Step:",i,".5")
  elseif i*2 > n_steps  && i != n_steps
      #θ, r = θ_new, r_new/α
      println("full/α")
      println("Step:",i,".5")
  else
      #θ, r = θ_new, r_new/α_sqrt
      println("full/sqrt(α)")
      println("Step:",i)
  end
end
```

Lastly, @xukai92 raised the issue of how to exactly implement the tempering code within the existing code structure as well as how much effect it will have on `AHMC` sampling:

> So it comes to a design issue, should I implement this conditions inside `step` and check if `lf.α == 1`, or we’d better introduce another `Integrator` type which does this job and implement `lf_momentum` and `lf_position` separate for each type and rely on multiple dispatch to deal with this conditions. Not sure what’s the better way. But for your purpose add this conditions should work.
**However**, I’m not sure how does this actually interact with NUTS (for plain HMC it seems all good), as NUTS only call `steps` with `n_steps=1` and the trajectory building relies on its on doubling tree algorithm.

jonasmac16:
An update from above code(fixing an error), the tempering could be conceptually done like this:
```
n_steps =  5

α = 1.02
[...]

PR #96, opened 2019-09-11
Implement tempering in leapfrog integrator

xukai92:
This PR
- Removes some old comments
- Resolves https://github.com/TuringLang/AdvancedHMC.jl/issues/52

### DIFF
--- a/src/AdvancedHMC.jl
+++ b/src/AdvancedHMC.jl
@@ -7,16 +7,11 @@ using LinearAlgebra: Symmetric, UpperTriangular, mul!, ldiv!, dot, I, diag, chol
 using StatsFuns: logaddexp
 using LazyArrays: BroadcastArray
 using Random: GLOBAL_RNG, AbstractRNG
-using ProgressMeter
+using ProgressMeter: Progress, next!
 using Parameters: @unpack, reconstruct
 using ArgCheck: @argcheck
 
 import StatsBase: sample
-
-# Notations
-# d - dimension of sampling sapce
-# π(θ) - target distribution
-# r - momentum variable
 
 include("adaptation/Adaptation.jl")
 export UnitEuclideanMetric, DiagEuclideanMetric, DenseEuclideanMetric
--- a/src/AdvancedHMC.jl
+++ b/src/AdvancedHMC.jl
@@ -22,7 +22,7 @@ import .Adaptation: adapt!
 include("hamiltonian.jl")
 export Hamiltonian
 include("integrator.jl")
-export Leapfrog
+export Leapfrog, JitteredLeapfrog, TemperedLeapfrog
 include("trajectory.jl")
 export StaticTrajectory, find_good_eps, HMCDA, NUTS, Slice, Multinomial, ClassicNoUTurn, GeneralisedNoUTurn
 
--- a/src/AdvancedHMC.jl
+++ b/src/AdvancedHMC.jl
@@ -7,11 +7,17 @@ using LinearAlgebra: Symmetric, UpperTriangular, mul!, ldiv!, dot, I, diag, chol
 using StatsFuns: logaddexp
 using LazyArrays: BroadcastArray
 using Random: GLOBAL_RNG, AbstractRNG
-using ProgressMeter: Progress, next!
+using ProgressMeter
 using Parameters: @unpack, reconstruct
 using ArgCheck: @argcheck
 
 import StatsBase: sample
+
+# Notations	
+# ℓπ: log density of the target distribution
+# ∂ℓπ∂θ: gradient of the log density of the target distribution
+# θ: position variables / model parameters
+# r: momentum variables
 
 include("adaptation/Adaptation.jl")
 export UnitEuclideanMetric, DiagEuclideanMetric, DenseEuclideanMetric
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -8,7 +8,7 @@ abstract type AbstractIntegrator end
 abstract type AbstractLeapfrog{T} <: AbstractIntegrator end
 
 jitter(::AbstractRNG, ::AbstractLeapfrog, ϵ) = ϵ
-temper(lf::AbstractLeapfrog, r, ::NamedTuple{(:i, :is_first),Tuple{Int64,Bool}}, ::Int) = r
+temper(lf::AbstractLeapfrog, r, ::NamedTuple{(:i, :is_half),Tuple{Int64,Bool}}, ::Int) = r
 
 function step(
     rng::AbstractRNG,
@@ -26,7 +26,7 @@ function step(
     @unpack value, gradient = ∂H∂θ(h, θ)
     for i = 1:n_steps
         # Tempering
-        r = temper(lf, r, (i=i, is_first=true), n_steps)
+        r = temper(lf, r, (i=i, is_half=true), n_steps)
         # Take a half leapfrog step for momentum variable
         r = r - ϵ / 2 * gradient
         # Take a full leapfrog step for position variable
@@ -36,7 +36,7 @@ function step(
         @unpack value, gradient = ∂H∂θ(h, θ)
         r = r - ϵ / 2 * gradient
         # Tempering
-        r = temper(lf, r, (i=i, is_first=false), n_steps)
+        r = temper(lf, r, (i=i, is_half=false), n_steps)
         # Create a new phase point by caching the logdensity and gradient
         z = phasepoint(h, θ, r; ℓπ=DualValue(value, gradient))
         !isfinite(z) && break
@@ -79,14 +79,13 @@ function Base.show(io::IO, l::TemperedLeapfrog)
 end
 
 """
-    temper(lf::TemperedLeapfrog, r, step::NamedTuple{(:i, :is_first),Tuple{Int64,Bool}}, n_steps::Int)
+    temper(lf::TemperedLeapfrog, r, step::NamedTuple{(:i, :is_half),Tuple{Int64,Bool}}, n_steps::Int)
 
 Tempering step. `step` is a named tuple with 
 - `i` being the current leapfrog iteration and 
-- `is_first` indicating whether or not it's the first half momentum/tempering step
-
+- `is_half` indicating whether or not it's (the first) half momentum/tempering step
 """
-function temper(lf::TemperedLeapfrog, r, step::NamedTuple{(:i, :is_first),Tuple{Int64,Bool}}, n_steps::Int)
-    i_temper = 2(step.i - 1) + 1 + step.is_first    # counter for half temper steps
+function temper(lf::TemperedLeapfrog, r, step::NamedTuple{(:i, :is_half),Tuple{Int64,Bool}}, n_steps::Int)
+    i_temper = 2(step.i - 1) + 1 + step.is_half    # counter for half temper steps
     return i_temper <= n_steps ? r * sqrt(lf.α) : r / sqrt(lf.α)
 end
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -7,19 +7,19 @@ abstract type AbstractIntegrator end
 
 struct Leapfrog{T<:AbstractFloat} <: AbstractIntegrator
     ϵ       ::  T
-    jitter  :: T
+    jitter  ::  T
 end
 Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round(l.ϵ; sigdigits=3)))")
 
-Leapfrog(ϵ::T) where {T<:AbstractFloat} = Leapfrog(ϵ, zero(T))
+Leapfrog(ϵ::T; jitter::T=zero(T)) where {T<:AbstractFloat} = Leapfrog{T}(ϵ, jitter)
 
 function step(
-    lf::Leapfrog{F},
+    lf::Leapfrog{T},
     h::Hamiltonian,
     z::PhasePoint,
     n_steps::Int=1;
     fwd::Bool=n_steps > 0 # simulate hamiltonian backward when n_steps < 0,
-) where {F<:AbstractFloat,T<:Real}
+) where {T<:AbstractFloat}
     @unpack θ, r = z
     ϵ = fwd ? lf.ϵ : -lf.ϵ
     # Jitter step size; ref: https://github.com/stan-dev/stan/blob/1bb054027b01326e66ec610e95ef9b2a60aa6bec/src/stan/mcmc/hmc/base_hmc.hpp#L177-L178
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -8,7 +8,7 @@ abstract type AbstractIntegrator end
 abstract type AbstractLeapfrog{T} <: AbstractIntegrator end
 
 jitter(::AbstractRNG, ::AbstractLeapfrog, ϵ) = ϵ
-temper(lf::AbstractLeapfrog, r, i::Tuple{Int,Bool}, n_steps::Int) = r
+temper(lf::AbstractLeapfrog, r, ::NamedTuple{(:i, :is_first),Tuple{Int64,Bool}}, ::Int) = r
 
 function step(
     rng::AbstractRNG,
@@ -26,7 +26,7 @@ function step(
     @unpack value, gradient = ∂H∂θ(h, θ)
     for i = 1:n_steps
         # Tempering
-        r = temper(lf, r, (i,false), n_steps)
+        r = temper(lf, r, (i=i, is_first=true), n_steps)
         # Take a half leapfrog step for momentum variable
         r = r - ϵ / 2 * gradient
         # Take a full leapfrog step for position variable
@@ -36,7 +36,7 @@ function step(
         @unpack value, gradient = ∂H∂θ(h, θ)
         r = r - ϵ / 2 * gradient
         # Tempering
-        r = temper(lf, r, (i,true), n_steps)
+        r = temper(lf, r, (i=i, is_first=false), n_steps)
         # Create a new phase point by caching the logdensity and gradient
         z = phasepoint(h, θ, r; ℓπ=DualValue(value, gradient))
         !isfinite(z) && break
@@ -78,12 +78,15 @@ function Base.show(io::IO, l::TemperedLeapfrog)
     print(io, "TemperedLeapfrog(ϵ=$(round(l.ϵ; sigdigits=3)), α=$(round(l.α; sigdigits=3)))")
 end
 
-function temper(lf::TemperedLeapfrog, r, i::Tuple{Int,Bool}, n_steps::Int)
-    if i[2] == true # first half of leapfrog
-        # `ceil` includes mid if `n_steps` is odd, e.g. `<= ceil(5 / 2)` => `<= 3`
-        return i[1] <= ceil(Int, n_steps / 2) ? r * sqrt(lf.α) : r / sqrt(lf.α)
-    else           # second half of leapfrog
-        # `floor` excludes mid if `n_steps` is odd, e.g. `<= floor(5 / 2)` => `<= 2`
-        return i[1] <= floor(Int, n_steps / 2) ? r * sqrt(lf.α) : r / sqrt(lf.α)
-    end
+"""
+    temper(lf::TemperedLeapfrog, r, step::NamedTuple{(:i, :is_first),Tuple{Int64,Bool}}, n_steps::Int)
+
+Tempering step. `step` is a named tuple with 
+- `i` being the current leapfrog iteration and 
+- `is_first` indicating whether or not it's the first half momentum/tempering step
+
+"""
+function temper(lf::TemperedLeapfrog, r, step::NamedTuple{(:i, :is_first),Tuple{Int64,Bool}}, n_steps::Int)
+    i_temper = 2(step.i - 1) + 1 + step.is_first    # counter for half temper steps
+    return i_temper <= n_steps ? r * sqrt(lf.α) : r / sqrt(lf.α)
 end
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -8,10 +8,11 @@ abstract type AbstractIntegrator end
 struct Leapfrog{T<:AbstractFloat} <: AbstractIntegrator
     ϵ       ::  T
     jitter  ::  T
+    α       ::  T
 end
 Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round(l.ϵ; sigdigits=3)))")
 
-Leapfrog(ϵ::T; jitter::T=zero(T)) where {T<:AbstractFloat} = Leapfrog{T}(ϵ, jitter)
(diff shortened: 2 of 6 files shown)


src/integrator.jl
```diff
@@ -13,7 +13,6 @@ Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round(l.ϵ; sigdigits=
 
 Leapfrog(ϵ::T) where {T<:AbstractFloat} = Leapfrog(ϵ, zero(T))
 
-# TODO: double check the function below to see if it is type stable or not
 function step(
```
xukai92: @yebai Regarding the leapfrog step implementation, we discussed before that the current one is less optimized. It seems that combing L30 and L34 can give nontrivial improvement. I checked it via the following example.

```julia
julia> using BenchmarkTools

julia> x = rand(100);

julia> y = rand(100);

julia> eps = 0.01;

julia> f1() = x - eps * y
f1 (generic function with 1 method)

julia> f2() = (z = x - eps / 2 * y; z = z - eps / 2 * y)
f2 (generic function with 1 method)

julia> @benchmark f1()
BenchmarkTools.Trial: 
  memory estimate:  1.75 KiB
  allocs estimate:  2
  --------------
  minimum time:     140.348 ns (0.00% GC)
  median time:      189.408 ns (0.00% GC)
  mean time:        223.946 ns (8.04% GC)
  maximum time:     70.844 μs (99.36% GC)
  --------------
  samples:          10000
  evals/sample:     451

julia> @benchmark f2()
BenchmarkTools.Trial: 
  memory estimate:  3.53 KiB
  allocs estimate:  6
  --------------
  minimum time:     290.158 ns (0.00% GC)
  median time:      325.403 ns (0.00% GC)
  mean time:        400.899 ns (9.83% GC)
  maximum time:     156.173 μs (99.57% GC)
  --------------
  samples:          10000
  evals/sample:     196
```

yebai:
Thanks, @xukai92. I'll take a look soon.

src/AdvancedHMC.jl
```diff
@@ -7,17 +7,12 @@ using LinearAlgebra: Symmetric, UpperTriangular, mul!, ldiv!, dot, I, diag, chol
...
 using LazyArrays: BroadcastArray
 using Random: GLOBAL_RNG, AbstractRNG
-using ProgressMeter
+using ProgressMeter: Progress, next!
 using Parameters: @unpack, reconstruct
 using ArgCheck: @argcheck
 
 import StatsBase: sample
 
-# Notations
```
yebai: I would keep these notations although it is not complete. We can keep improving it.

src/integrator.jl
yebai: It is hard to tell whether this is a significant improvement without comparison to the total sampling time: as far as I know, The time for `leapfrog` might be a tiny fraction of total computation time, while most computation is spent on computing the `logpdf` and its gradients. I suggest that we keep `leapfrog` as-is` for now and change it only when more concrete benchmarking is performed.

src/AdvancedHMC.jl
xukai92: OK I can add them back **with the updated names**.

xukai92:
> The code looks correct to me. Maybe we can refactor the implementation a bit, e.g. by adding some subtypes of `AbstractIntegrator`: `JitteredLeapfrog`, `TemperedLeapfrog`, `Leapfrog`. We can then overload the `step` function for these different `Integrator` types. This is helpful for readability and avoiding misuse since we can't really use the jittering trick together with tempering.

I also thought about this by I feel it's hard to achieve so without repeating many codes. I will give a try to see how it goes though.

yebai:
It’s probably ok to repeat the code here, since the leapfrog function is quite short.

xukai92:
So the code will look something like below

```julia
####
#### Numerical methods for simulating Hamiltonian trajectory.
####


abstract type AbstractIntegrator end

struct Leapfrog{T<:AbstractFloat} <: AbstractIntegrator
    ϵ   ::  T
end
Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round(l.ϵ; sigdigits=3)))")

function Leapfrog(ϵ::T; jitter::T=zero(T), α::T=one(T)) where {T<:AbstractFloat}
    if jitter == zero(T) && α == one(T)
        return Leapfrog(ϵ)
    elseif jitter != zero(T) && α == one(T)
        return JitteredLeapfrog(ϵ, jitter)
    elseif jitter == zero(T) && α != one(T)
        return TemperedLeapfrog(ϵ, α)
    else
        error("Leapfrog integrator with both jittering and tempering is not supported")
    end
end

function step(
    lf::Leapfrog{T},
    h::Hamiltonian,
    z::PhasePoint,
    n_steps::Int=1;
    fwd::Bool=n_steps > 0 # simulate hamiltonian backward when n_steps < 0,
) where {T<:AbstractFloat}
    ϵ = fwd ? lf.ϵ : -lf.ϵ
    n_steps = abs(n_steps)

    @unpack θ, r = z
    @unpack value, gradient = ∂H∂θ(h, θ)
    for i = 1:n_steps
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
        ∇r = ∂H∂r(h, r)
        θ = θ + ϵ * ∇r  # take a full leapfrog step for position variable
        @unpack value, gradient = ∂H∂θ(h, θ)
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
        # Create a new phase point by caching the logdensity and gradient
        z = phasepoint(h, θ, r; ℓπ=DualValue(value, gradient))
        !isfinite(z) && break
    end
    return z
end

struct JitteredLeapfrog{T<:AbstractFloat} <: AbstractIntegrator
    ϵ       ::  T
    jitter  ::  T
end
function Base.show(io::IO, l::JitteredLeapfrog)
    print(io, "JitteredLeapfrog(ϵ=$(round(l.ϵ; sigdigits=3)), jitter=$(round(l.jitter; sigdigits=3)))")
end

function step(
    lf::JitteredLeapfrog{T},
    h::Hamiltonian,
    z::PhasePoint,
    n_steps::Int=1;
    fwd::Bool=n_steps > 0 # simulate hamiltonian backward when n_steps < 0,
) where {T<:AbstractFloat}
    ϵ = fwd ? lf.ϵ : -lf.ϵ
    n_steps = abs(n_steps)

    # Jitter step size; ref: https://github.com/stan-dev/stan/blob/1bb054027b01326e66ec610e95ef9b2a60aa6bec/src/stan/mcmc/hmc/base_hmc.hpp#L177-L178
    !iszero(lf.jitter) && (ϵ *= (1 + lf.jitter * (2 * rand() - 1)))

    @unpack θ, r = z
    @unpack value, gradient = ∂H∂θ(h, θ)
    for i = 1:n_steps
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
        ∇r = ∂H∂r(h, r)
        θ = θ + ϵ * ∇r  # take a full leapfrog step for position variable
        @unpack value, gradient = ∂H∂θ(h, θ)
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
        # Create a new phase point by caching the logdensity and gradient
        z = phasepoint(h, θ, r; ℓπ=DualValue(value, gradient))
        !isfinite(z) && break
    end
    return z
end

struct TemperedLeapfrog{T<:AbstractFloat} <: AbstractIntegrator
    ϵ       ::  T
    α       ::  T
end
function Base.show(io::IO, l::TemperedLeapfrog)
    print(io, "TemperedLeapfrog(ϵ=$(round(l.ϵ; sigdigits=3)), α=$(round(l.α; sigdigits=3)))")
end

function step(
    lf::TemperedLeapfrog{T},
    h::Hamiltonian,
    z::PhasePoint,
    n_steps::Int=1;
    fwd::Bool=n_steps > 0 # simulate hamiltonian backward when n_steps < 0,
) where {T<:AbstractFloat}
    ϵ = fwd ? lf.ϵ : -lf.ϵ
    n_steps = abs(n_steps)

    @unpack θ, r = z
    @unpack value, gradient = ∂H∂θ(h, θ)
    for i = 1:n_steps
        # Tempering; `ceil` includes mid if `n_steps` is odd, e.g. `<= ceil(5 / 2)` => `<= 3` 
        !isone(sqrt(lf.α)) && (r = i <= ceil(Int, n_steps / 2) ? r * sqrt(lf.α) : r / sqrt(lf.α))
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
        ∇r = ∂H∂r(h, r)
        θ = θ + ϵ * ∇r  # take a full leapfrog step for position variable
        @unpack value, gradient = ∂H∂θ(h, θ)
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
        # Tempering; `floor` excludes mid if `n_steps` is odd, e.g. `<= floor(5 / 2)` => `<= 2` 
        !isone(sqrt(lf.α)) && (r = i <= floor(Int, n_steps / 2) ? r * sqrt(lf.α) : r / sqrt(lf.α))
        # Create a new phase point by caching the logdensity and gradient
        z = phasepoint(h, θ, r; ℓπ=DualValue(value, gradient))
        !isfinite(z) && break
    end
    return z
end
```

A lot of code repeating here:
- Prepare stage
```julia
    ϵ = fwd ? lf.ϵ : -lf.ϵ
    n_steps = abs(n_steps)

    # Optionally jittering

    @unpack θ, r = z
    @unpack value, gradient = ∂H∂θ(h, θ)
```
- The `for` loop

We cannot simply wrap the whole `for` loop into a function because the tempered version behaves differently.
- the single leapfrog step

We can also extract the single leapfrog step out, i.e. define a function for 
```julia
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
        ∇r = ∂H∂r(h, r)
        θ = θ + ϵ * ∇r  # take a full leapfrog step for position variable
        @unpack value, gradient = ∂H∂θ(h, θ)
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
```
But I feel this also reduce the readbility at this function need to return gradient to be used for the next iteration.

To be honest, I think it's less readble than the old version.

Another way to tackle this is to introduce two functions `jitter` and `temper` which might do nothing depends on their types.

```julia
```abstract type AbstractIntegrator end

struct Leapfrog{T<:AbstractFloat} <: AbstractIntegrator
    ϵ       ::  T
    jitter  ::  T
    α       ::  T
end
Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round(l.ϵ; sigdigits=3)))")

Leapfrog(ϵ::T; jitter::T=zero(T), α::T=one(T)) where {T<:AbstractFloat} = Leapfrog{T}(ϵ, jitter, α)

function step(
    lf::Leapfrog{T},
    h::Hamiltonian,
    z::PhasePoint,
    n_steps::Int=1;
    fwd::Bool=n_steps > 0 # simulate hamiltonian backward when n_steps < 0,
) where {T<:AbstractFloat}
    @unpack θ, r = z
    ϵ = fwd ? lf.ϵ : -lf.ϵ
    sqrtα = sqrt(lf.α)
    n_steps = abs(n_steps)
    # Jitter step size; ref: https://github.com/stan-dev/stan/blob/1bb054027b01326e66ec610e95ef9b2a60aa6bec/src/stan/mcmc/hmc/base_hmc.hpp#L177-L178
    ϵ = jitter(lf, ϵ)

    @unpack value, gradient = ∂H∂θ(h, θ)
    for i = 1:n_steps
        r = temper1(lf, r, i, n_steps)
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
        ∇r = ∂H∂r(h, r)
        θ = θ + ϵ * ∇r  # take a full leapfrog step for position variable
        @unpack value, gradient = ∂H∂θ(h, θ)
        r = r - ϵ / 2 * gradient    # take a half leapfrog step for momentum variable
        r = temper2(lf, i, n_steps)
        # Create a new phase point by caching the logdensity and gradient
        z = phasepoint(h, θ, r; ℓπ=DualValue(value, gradient))
        !isfinite(z) && break
    end
    return z
end
```

How do you think?

yebai:
Introducing two new functions `jittering` and `tempering` sounds good. Also, you can try `@generated` functions - https://docs.julialang.org/en/v1/manual/metaprogramming/index.html#An-advanced-example-1.

xukai92:
> Introducing two new functions jittering and tempering sounds good. 

I will try this approach then.

> Alternatively, you can try @generated functions - https://docs.julialang.org/en/v1/manual/metaprogramming/index.html#An-advanced-example-1.

I don't feel we need it here 
1. If it's for stability, the original version doesn't meet any type stability issue.
2. If it's for avoid writing too many codes, I feel with `@ generated` the code actually becomes harder to read and harder to extend.

src/integrator.jl
```diff
@@ -5,37 +5,76 @@
...
+struct TemperedLeapfrog{T<:AbstractFloat} <: AbstractLeapfrog{T}
+    ϵ       ::  T
+    α       ::  T
+end
+
+function Base.show(io::IO, l::TemperedLeapfrog)
+    print(io, "TemperedLeapfrog(ϵ=$(round(l.ϵ; sigdigits=3)), α=$(round(l.α; sigdigits=3)))")
+end
+
+function temper_first(lf::TemperedLeapfrog, r, i::Int, n_steps::Int)
```
yebai: Maybe merge `temper_first` and `temper_second` into `termper`? Since their function signature is the same, and we can distinguish multiplication and division by inspecting `i` and `n`.

src/sampler.jl
```diff
@@ -53,13 +53,13 @@ Progress meter update with all trajectory stats, iteration number and metric sho
 function pm_next!(pm, stat::NamedTuple, i::Int, metric::AbstractMetric)
     # Add current iteration and mass matrix
     stat = (iterations=i, stat..., mass_matrix=metric)
-    ProgressMeter.next!(pm; showvalues=[tuple(s...) for s in pairs(stat)])
+    next!(pm; showvalues=[tuple(s...) for s in pairs(stat)])
```
yebai: I prefer keeping the package name prefix because it helps to recognise where a function is from.

src/integrator.jl
```diff
@@ -5,37 +5,76 @@
...
+    ϵ       ::  T
+    jitter  ::  T
+end
+
+function Base.show(io::IO, l::JitteredLeapfrog)
+    print(io, "JitteredLeapfrog(ϵ=$(round(l.ϵ; sigdigits=3)), jitter=$(round(l.jitter; sigdigits=3)))")
+end
+
+# Jitter step size; ref: https://github.com/stan-dev/stan/blob/1bb054027b01326e66ec610e95ef9b2a60aa6bec/src/stan/mcmc/hmc/base_hmc.hpp#L177-L178
+jitter(lf::JitteredLeapfrog, ϵ) = ϵ * (1 + lf.jitter * (2 * rand() - 1))
```
yebai: Consider adding a `RNG` argument here since it relies on random generator.

yebai:
Hi Kai, the PR looks correct to me - I left some minor comments above.

src/integrator.jl
xukai92: I'd like to keep the current syntax for now. If we merge them we just get a function with a if statement. Plus the calling is less readable (one cannot tell what does `true` mean in `temper(lf, r, i, n_steps ,true)` directly).

src/sampler.jl
xukai92: Sounds good.

src/integrator.jl
xukai92: Good point!

xukai92: The currently `step` doesn't pass RNG so there are two options here.
1. Pass RNG to `step`
2. Add RNG to `JitteredLeapfrog`
How do you think?

yebai: Sounds good!

src/integrator.jl
yebai: That's true. Maybe move the odd-checking logic for leapfrog steps out of `temper`? E.g. 

```julia

temper(...)
# some stuff
!isodd(n_steps) && temper(...)
```

This way the `temper` function will also be simplified (not more odd-checking code).

yebai: The code above has a bug. It adds an extra multiplication to moment if the number of leapfrog steps is odd. But maybe we can do the following 

```julia
!isodd(n_steps) && temper(...)
# some stuff
!isodd(n_steps) && temper(...)
```

This won't do any change to the moment for the `mid` leapfrog step if `n_step` is odd.

xukai92: By `isodd` do you mean `< div(n_steps, 2)` and `> div(n_steps, 2)`?

xukai92:
```
!isodd(n_steps) && temper(...)
# some stuff
!isodd(n_steps) && temper(...)
``
Doesn't quite make sense to me. We always do this two `temper` steps regardless if `n_steps` is odd. The logic about the odd number thing is to implement Neal's tempering scheme in which the middle lf step do multiplication first then division. (Other lf steps either do 2 multiplication or 2 division).

Comment (yebai) [reply src/integrator.jl]: A slightly better way is to define

```julia

@enum TemperedState TemperStep1 TemperStep2
function temper(lf::TemperedLeapfrog, r, i::Int, n_steps::Int, s::TemperedState) end

# Then, 
temper(..., TemperStep1)
# some stuff
temper(..., TemperStep2)
```

yebai: > By isodd do you mean < div(n_steps, 2) and > div(n_steps, 2)?

Julia has a built-in function `isodd`, e.g. `isodd(n_steps)`. 

> Doesn't quite make sense to me. We always do this two `temper` steps regardless if `n_steps` is odd. The logic about the odd number thing is to implement Neal's tempering scheme in which the middle lf step do multiplication first then division. (Other lf steps either do 2 multiplication or 2 division). 

Yes, there are different variants of the tempering trick, e.g. do nothing for the mid-step if `n_step` is odd.

xukai92: OK I see it. I can live with this but the current implementation also handles `isodd(n_steps)=true` correctly for Neal's version, what do we want to change?

Also even with the `isodd` function, we still need to have something like `temper_first` and `temper_second` or `@enum TemperedState TemperStep1 TemperStep2` to resolve which half it is. In fact, I don't like `@enum` here because we cannot dispatch over that, and this means we sill need a branch in that function, right?

I cannot see why
```julia
# Then, 
temper(..., TemperStep1)
# some stuff
temper(..., TemperStep2)
```
with
```julia
function temper(..., t)
    if t == TemperStep1
        ...
    elseif t == TemperStep2
        ...
    end
end
```
is over
```julia
# Then, 
temper_first(...)
# some stuff
temper_second(...)
```

xukai92: I think a bit more and feel your proposal of tempering (do nothing to odd) might be neater. I also think this scheme is correct. I will give a try tomorrow.

xukai92: @yebai I actually realize using iterators here would lead to much nicer code. Please check the update.

src/integrator.jl
```diff
@@ -5,37 +5,87 @@
...
+end
+
+function Base.show(io::IO, l::TemperedLeapfrog)
+    print(io, "TemperedLeapfrog(ϵ=$(round(l.ϵ; sigdigits=3)), α=$(round(l.α; sigdigits=3)))")
+end
+
+function get_tempering_schedule(lf::TemperedLeapfrog, n_steps)
+    up(r) = r * sqrt(lf.α)
+    down(r) = r / sqrt(lf.α)
+    return (i == 0 ? r -> r : i <= n_steps ? up : down for i in 0:2n_steps)
```
xukai92: I have to introduce an unused starting state to make the interface easy.

src/integrator.jl
```diff
@@ -5,37 +5,87 @@
...
-    ϵ       ::  T
-    jitter  :: T
-end
-Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round(l.ϵ; sigdigits=3)))")
+abstract type AbstractLeapfrog{T} <: AbstractIntegrator end
 
-Leapfrog(ϵ::T) where {T<:AbstractFloat} = Leapfrog(ϵ, zero(T))
+jitter(::AbstractRNG, ::AbstractLeapfrog, ϵ) = ϵ
+function get_tempering_schedule(lf::AbstractLeapfrog, n_steps)
+    return Iterators.repeated(r -> r, 1 + 2n_steps)
```
xukai92: I have to introduce an unused starting state to make the interface easy.

src/integrator.jl
```diff
@@ -5,37 +5,87 @@
...
-    # Jitter step size; ref: https://github.com/stan-dev/stan/blob/1bb054027b01326e66ec610e95ef9b2a60aa6bec/src/stan/mcmc/hmc/base_hmc.hpp#L177-L178
-    iszero(lf.jitter) || (ϵ *= (1 + lf.jitter * (2 * rand() - 1)))
+    ϵ = jitter(rng, lf, ϵ)
+    tempering_schedule = get_tempering_schedule(lf, n_steps)
 
+    @unpack θ, r = z
     @unpack value, gradient = ∂H∂θ(h, θ)
-    for i = 1:abs(n_steps)
-        r = r - ϵ / 2 * gradient # take a half leapfrog step for momentum variable
+    _, temper_state = iterate(tempering_schedule)
```
xukai92: Here is the unused starting state.

src/integrator.jl
yebai: Hi Kai, I feel the new code is a bit harder to read than before, so I did a bit refactoring of the tempering code without functionality change. It's basically changing the argument `i::Int` into ` tuple `(i::Int, ::Bool)`, where the bool value indicates first or second half of the leapfrog step.

yebai:
Tests pass on my machine - the current Travis CI is failing due to weird issues with the test coverage package.

src/integrator.jl
xukai92: Fair enough. 

The code you wrote seems to be buggy. `false` was passed to `temper` in the first call but `if i[2] == true # first half of leapfrog` was written in `temper`, which means `i[2]` was used to indicate if the step is the first half tempering step, right?

In fact, I don't think using a tuple is readable 
- For the `step` function, passing `(i,false)` is mysterious - what does `false` mean here?
- For the `temper` function - `i[2] == true` means nothing to someone is reading only that function. 

I updated the current version to use a named tuple, fix and simplify the logic in `temper`.
The trick of `floor` and `ceil` I used was a bit mysterious TBH. In fact, the logic could be much simpler (as in my previous version):

Let `i` being the current leapfrog iteration counter and let `is_first` being whether or not it's the first step in the current leapfrog iteration. Then `i_temper`, the counter for half tempering step, is simply `2(i - 1) + 1 + is_first` and the logic for tempering is simply
- `if i_temper <= n_steps`, temper up
- `else`, temper down

This deals with both even and odd `n_steps`.

yebai: Looks good. Maybe rename is_first to is_half? Since it indicates half and full leapfrog step here.

xukai92: Done

Issue #98, opened 2019-09-13
Initial mass matrix is reset to identity when n_adapts>0


andreasnoack:
I'm using the example from the README but I'm passing a custom mass matrix. I have added a bit of `sleep` to the gradient to slow things down. Notice that the mass matrix is immediately set to the identity. Hence, if you have a good initial estimate of the mass matrix, it appears that you'd currently have to set `n_adapts=0.`

```julia
### Define the target distribution and its gradient
using Distributions: logpdf, MvNormal
using DiffResults: GradientResult, value, gradient
using ForwardDiff: gradient!

D = 10
target = MvNormal(zeros(D), ones(D))
ℓπ(θ) = logpdf(target, θ)

function ∂ℓπ∂θ(θ)
    sleep(0.0001)
    res = GradientResult(θ)
    gradient!(res, ℓπ, θ)
    return (value(res), gradient(res))
end

### Build up a HMC sampler to draw samples
using AdvancedHMC

# Sampling parameter settings
n_samples, n_adapts = 10_000, 2_000

# Draw a random starting points
θ_init = randn(D)

# Define metric space, Hamiltonian, sampling method and adaptor
metric = DiagEuclideanMetric(fill(0.5, D))
h = Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
int = Leapfrog(find_good_eps(h, θ_init))
prop = NUTS{AdvancedHMC.Multinomial,GeneralisedNoUTurn}(int)
adaptor = StanHMCAdaptor(n_adapts, Preconditioner(metric), NesterovDualAveraging(0.8, int.ϵ))

# Draw samples via simulating Hamiltonian dynamics
# - `samples` will store the samples
# - `stats` will store statistics for each sample
samples, stats = sample(h, prop, θ_init, n_samples, adaptor, n_adapts; progress=true)
```

xukai92:
Good catch. The current code handles this incorrectly. I will push some fixes soon.

xukai92:
Fixed by https://github.com/TuringLang/AdvancedHMC.jl/pull/100

PR #100, opened 2019-09-13
Init. adaptor via metric

xukai92:
Fix issue https://github.com/TuringLang/AdvancedHMC.jl/issues/98

### DIFF
--- a/src/sampler.jl
+++ b/src/sampler.jl
@@ -37,9 +37,6 @@ function adapt!(
     θ::AbstractVector{T},
     α::T
 ) where {T<:Real}
-    if i == 1
-        setM⁻¹!(adaptor, h.metric.M⁻¹)
-    end
     isadapted = false
     if i <= n_adapts
         adapt!(adaptor, θ, α)
--- a/src/sampler.jl
+++ b/src/sampler.jl
@@ -37,6 +37,9 @@ function adapt!(
     θ::AbstractVector{T},
     α::T
 ) where {T<:Real}
+    if i == 1
+        setM⁻¹!(adaptor, h.metric.M⁻¹)
+    end
     isadapted = false
     if i <= n_adapts
         adapt!(adaptor, θ, α)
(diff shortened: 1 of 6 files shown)


test/adaptation/precond.jl
```diff
@@ -112,43 +109,56 @@ end
...
+                @test res.adaptor.pc.var ≈ diag(Σ) rtol=0.2
+                
+                res = runnuts(ℓπ, ∂ℓπ∂θ, DenseEuclideanMetric(D))
+                @test res.adaptor.pc.covar ≈ Σ rtol=0.25
+            end
         end
+
     end
 
+    @testset "Initialisation adaptor by metric" begin
```
xukai92: Only this test is new. Other changes are just moving around codes.

src/sampler.jl
```diff
@@ -37,6 +37,9 @@ function adapt!(
     θ::AbstractVector{T},
     α::T
 ) where {T<:Real}
+    if i == 1
```
yebai: The code looks correct to me. Perhaps we can avoid introducing an the `setM⁻¹!` function? E.g we can initialise `adaptor` with the user-specified mass matrix when it is constructed, instead of setting it in the sampling step? 

```julia
n_adapts, D = 2_000, 10
mass_init = fill(0.5, D)

# initialize mass matrix when `adaptor` is created
metric = DiagEuclideanMetric(mass_init) 
adaptor = StanHMCAdaptor(n_adapts, Preconditioner(metric), NesterovDualAveraging(0.8, int.ϵ))
```

This would make the code slightly cleaner IMO.

xukai92: This is actually a better design! I will try to change the code for this!

xukai92: As a side note, this requires the user to use `Preconditioner(metric)` to create the preconditioner. But this is fine as this is the most common way to create a preconditioner.

andreasnoack:
Thanks for fixing this so promptly.

Issue #35, opened 2019-03-28
Samples during adaption phase


xukai92:
Maybe we should make the `sample` function not return those samples by default (and support an optional keyword for it)?

yebai:
yes, but let's consider fix these issues altogether in https://github.com/TuringLang/Turing.jl/issues/723

yebai:
Is this issue resolved in AHMC now?

xukai92:
AHMC currently doesn't deal with dropping adaption phase samples at all. This is done in Turing side. I'm not sure if we want to support it here at all? How do you think?

xukai92:
We can also implement it here. So Turing side can just droping adaption samples via passing a keyword argument to AHMC. I just don't know which one is better (this or the current one).

yebai:
Maybe  we can create a keyword argument in AHMC, the use it optionally in other packages like Turing.

cpfiffer:
I think Turing should maintain it's current behavior by dropping the samples at the end of sampling. If it's done strictly on the AHMC side, then the NUTS/HMC code won't work so well in Gibbs. 

It is a good idea to include a keyword for AHMC though, if only to give users more options for raw AHMC.

xukai92:
Also, if AHMC knows those samples will be dropped, it doesn't have to store them at all, which could be a potential performance improvement.

xukai92:
> ... by dropping the samples at the end of sampling

Maybe Turing side should not keep the samples during adaptation at all if they were to be dropped in the end? @cpfiffer

cpfiffer:
That's really a design choice. You could move the adaptation stuff to `sample_init!` if you wanted to. A question I'd have is related to Gibbs: Do you need to run the other component samplers while HMC is running, or can it adapt independently? If it can adapt independently then we can just put the adaptation phase in `sample_init!` and everything's cool. Alternatively, if they need to be run together, we'd have to modify the `Gibbs` code to run the adaptation cycle ahead of time with all the component samplers.

yebai:
>  question I'd have is related to Gibbs: Do you need to run the other component samplers while HMC is running, or can it adapt independently? If it can adapt independently then we can just put the adaptation phase in sample_init! and everything's cool.

It's likely this won't work well, because it'll only adapt a local covariance conditioned on variables not updated by HMC. 

> Maybe Turing side should not keep the samples during adaptation at all if they were to be dropped in the end? @cpfiffer

[...]

PR #108, opened 2019-10-01
Add kw to drop warmup samples

xukai92:


yebai:
Looks good to me. 

@cpfiffer Will this affect the Gibbs sampler, as you mentioned in #35 ?

xukai92:
Also I notice TemperedLeapfrog with adaptation is more likely to fail the tests. Do you think tempering should actually be disabled during adaptation phase?

yebai:
> Also I notice TemperedLeapfrog with adaptation is more likely to fail the tests. Do you think tempering should actually be disabled during adaptation phase?


Which sampler are you using?

xukai92:
Static HMC; see https://travis-ci.org/TuringLang/AdvancedHMC.jl/jobs/592055365#L377.

I already disable HMCDA with TemperedLeapfrog (https://github.com/TuringLang/AdvancedHMC.jl/blob/master/test/hmc.jl#L41) as this combination seems to be extremly unstable.

yebai:
It is likely tempering won't play nicely with NUTS yet, due to multiple calls to `step` in each dynamic trajectory.

yebai:
Happy to disable tempering during adaption for all cases.

PR #111, opened 2019-10-09
Fix effective depth

xukai92:
If the current `max_depth` is set to `n` it actually simulates a tree with maximum depth `n+1`. This PR fixes this offset.


andreasnoack:
Thanks for fixing this so promply. Would be great if you could make a bugfix release as well.

Issue #60, opened 2019-04-27
Move SGLD and SGHMC from Turing.jl


xukai92:

xukai92:
https://github.com/TuringLang/Turing.jl/issues/769

sivapvarma:
For implementing SGHMC and SGLD here, I am still at a loss as to how to query the mini batch gradient of posterior. Any pointer or some documentation about this repo would be helpful.

xukai92:
I think for the scope of this package, we should assume the density and gradient function returns that of a mini batch.

Janssena:
Heya, just wondering about the state of the SG samplers? 

I noticed that SGHMC / SGLD samplers exist in Turing, but details on the implementation and how data is mini-batched is missing?

PR #113, opened 2019-10-13
[WIP] add SGHMC, SGLD trajectories

sivapvarma:
Just placeholders for now, still getting a feeling for how things are organized in AHMC, at the same time wanted to get started.

Goal is to fix Issue #60 .

Comments are welcome.

### DIFF
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -723,6 +723,7 @@ struct SGHMC{
 } <: AbstractTrajectory{I}
     integrator      :: I
     n_steps         :: Int  # number of samples
+    batch_size      :: Int  # no of data points in minibatch for gradient estimate
     η               :: F    # learning rate
     α               :: F    # momentum decay
 end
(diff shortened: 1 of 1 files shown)


src/trajectory.jl
```diff
@@ -601,3 +601,76 @@ function update(
...
+## Stochastic Gradient Hamilton Samplers
+##
+
+###
+### Stochastic Gradient Hamiltonian Monte Carlo sampler.
+###
+"""
+Stochastic Gradient HMC with fixed number of steps.
+"""
+struct SGHMC{
```
xukai92: Need to be a subtype of `AbstractTrajectory`

src/trajectory.jl
```diff
@@ -601,3 +601,76 @@ function update(
...
+    α               :: F    # momentum decay
+end
+
+function transition(
+    rng::AbstractRNG,
+    τ::SGHMC,
+    h::Hamiltonian,
+    z::PhasePoint
+) where {T<:Real}
+    z′ = step(rng, τ.integrator, h, z, τ.n_steps)
```
xukai92: We probably only need to change this line to implmenet SGHMC.

src/trajectory.jl
```diff
@@ -601,3 +601,76 @@ function update(
...
+    ϵ               :: F    # constant scale factor of the learning rate
+end
+
+function transition(
+    rng::AbstractRNG,
+    τ::SGLD,
+    h::Hamiltonian,
+    z::PhasePoint
+) where {T<:Real}
+    z′ = step(rng, τ.integrator, h, z, τ.n_steps)
```
xukai92: We probably only need to change this line to implmenet SGLD.

xukai92:
I left some comments on how to proceed.

xukai92:
The commit history is messed up. Please rebase and force-push.

xukai92:
> I am still stuck about how we compute the stochastic gradients. My main source of confusion is that there are many AD frameworks supported by Turing ( ForwardDiff, ReverseDiff, Zygote, Tracker). I know Zygote is the way forward. But I keep getting lost in how all of them are handled by Turing. Any pointers to AD documentation for Turing would help.

I think the design here is Turing/AD-agonistic. All we need to assume is that `Hamiltonian.∂ℓπ∂θ` returns a tuple of value and gradient. 

> Furthermore, we need to compute gradients on minibatches, so it is still more details of how the interface should be designed.

I guess there are two options:

1. Assume the gradient is scaled correctly by users when implementing `∂ℓπ∂θ` or
2. Include the batch size (M) and whole dataset size (N) in `SGLD`/`SGHMC` and we scale it inside AHMC

Issue #99, opened 2019-09-13
Better names for sampling methods from trees


xukai92:
See https://github.com/TuringLang/AdvancedHMC.jl/commit/26aabb59c819d6b238aa70b80b692fdb51462934#commitcomment-34976303

xukai92:
Related discussion: https://github.com/TuringLang/AdvancedHMC.jl/pull/91#issuecomment-522636114

yebai:
Maybe we can remove `Slice` and `Multinomial` from exports? These exports are only useful for expert-users, so we might not want to export anyway.

xukai92:
I prefer to rename them for several reasons. 
- Users have to use `AdvancedHMC.Slice` if we don't export them.
- AdvancedHMC IS for "expert-users".

Some options:
1. Go for the simplicity: `Multinomial` -> `Multi`
2. Go for self-contained: `TrajectoryMultinomialSampler` and `TrajectorySliceSampler`

Any thoughts?

xukai92:
Option 3:

Define `@enum TrajectorySampler slice multinomial` and somehow dispatch over them. (Might be tricky?)

yebai:
I mean in most cases we will use the default NUTS configuration, so maybe it’s ok to remove Slice/Multinomial from exports. Also, the user can still explicitly import these names if needed.

yebai:
we can consider create a sugar syntax, e.g. NUTS(...) will create a NUTS sampler with Multinomial sampler.

yebai:
Alternatively, we can shorten `SliceTrajectorySampler` to `SliceTS`, similarly `MultinomialTS` and `MetropolisTS`.

xukai92:
I like this `???TS` proposal! Let's do this.

yebai:
Closed by #114.

PR #114, opened 2019-10-23
Rename trajector sampling structs

xukai92:
Better names for sampling methods from trees #99

### DIFF
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -26,82 +26,8 @@ Hamiltonian dynamics numerical simulation trajectories.
 """
 abstract type AbstractTrajectory{I<:AbstractIntegrator} <: AbstractProposal end
 
-"""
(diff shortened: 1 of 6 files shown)


src/trajectory.jl
```diff
@@ -134,67 +60,141 @@ end
...
 
 mh_accept(
     rng::AbstractRNG,
-    s::Multinomial,
-    s′::Multinomial
+    s::MultinomialTS,
+    s′::MultinomialTS
 ) = rand(rng) < min(1, exp(s′.ℓw - s.ℓw))
 
+"""
```
yebai: Is this new block just a copy and paste?

xukai92: Yes. I just moved the code to the very begining (before static trajectires). The reason is we later would like to reuse these struct for static trajecotries as well.

PR #120, opened 2019-10-31
Fix issues on 32bit

andreasnoack:
First I'll see if I can reproduce here and then I'll add the fix.


andreasnoack:
Why is CI not enabled for PRs?

PR #121, opened 2019-10-31
Fix issues on 32bit

andreasnoack:
Continuation of #120

### DIFF
--- a/src/adaptation/precond.jl
+++ b/src/adaptation/precond.jl
@@ -57,9 +57,9 @@ end
 
 # https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/var_adaptation.hpp
 function get_var(wv::WelfordVar{T, AT}) where {T<:AbstractFloat, AT<:AbstractVector{T}}
-    n, M = wv.n, wv.M
+    n, M = T(wv.n), wv.M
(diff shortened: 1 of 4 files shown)


andreasnoack:
Any idea what is going on with the preconditioner tests here. Why are they so much off on 32 bit?

xukai92:
Thanks for the PR. I suspect the failed test might be related to the term we added to stablise the online estimation at https://github.com/TuringLang/AdvancedHMC.jl/blob/master/src/adaptation/precond.jl#L62. I will take a closer look tomorrow.

andreasnoack:
Indeed there was integer overflow in that function on 32 bit. Test are now passing on 32 bit.

src/adaptation/precond.jl
```diff
@@ -57,9 +57,9 @@ end
 
 # https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/var_adaptation.hpp
 function get_var(wv::WelfordVar{T, AT}) where {T<:AbstractFloat, AT<:AbstractVector{T}}
-    n, M = wv.n, wv.M
+    n, M = T(wv.n), wv.M
```
xukai92: Thanks for the fix. Looks correct to me.

xukai92:
Will merge it in 12 hours.

Issue #123, opened 2019-11-02
Add more statistics


sethaxen:
PyMC3 has a comprehensive set of NUTS statistics [here](https://docs.pymc.io/api/inference.html#pymc3.step_methods.hmc.nuts.NUTS), some of which are missing from AdvancedHMC and would be nice to have returned when applicable. Here are the AdvancedHMC (=>PyMC3) ones that currently match up:
- step_size => step_size_bar (no jitter)
- n_steps => tree_size
- acceptance_rate => mean_tree_accept
- log_density => model_logp
- hamiltonian_energy => energy
- tree_depth => depth
- numerical_error => diverging (I think these technically aren't the same)

AdvancedHMC also has:
- is_accept

PyMC3 also has:
- energy_change: The difference in energy between the start and the end of the trajectory. For a perfect integrator this would always be zero.
- max_energy_change: The maximum difference in energy along the whole trajectory.
depth: The depth of the tree that was used to generate this sample
- tune: This is True, if step size adaptation was turned on when this sample was generated.
- step_size: The step size used for this sample.

The proposal is to add the following stats:
- hamiltonian_energy_error => energy_change
- max_hamiltonian_energy_error => max_energy_change
- is_adapt => tune
- step_size => step_size
- step_size_bar => step_size_bar

The latter two involving a renaming.

xukai92:
Looks good to me. Should be straight forward to add. Do you want to make a PR for this?

sethaxen:
Certainly!

sethaxen:
Fixed by #124

PR #124, opened 2019-11-03
Add more statistics and move jitter to transition

sethaxen:
See #123 for discussion

Adds the following statistics:
- [x] `hamiltonian_energy_error`
- [x] `max_hamiltonian_energy_error`
- [x] `is_adapt`
- [x] `nom_step_size`

To-do:
- [x] test that stats are returned, in particular that `is_adapt` starts as `true` and switches to `false` after adaptation is over
- [x] refactor code around `jitter` so that the step size used is accessible for stats. I'd appreciate input on this.

I'll submit a PR to Turing so that the new stats fields show up in `internals` instead of `parameters`. Looks like Turing doesn't use AdvancedHMC's `sample` function, so I'll need to add an `is_adapt` over there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ metric = DiagEuclideanMetric(D)
 h = Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
 int = Leapfrog(find_good_eps(h, θ_init))
 prop = NUTS{MultinomialTS,GeneralisedNoUTurn}(int)
-adaptor = StanHMCAdaptor(n_adapts, Preconditioner(metric), NesterovDualAveraging(0.8, int.ϵ))
+adaptor = StanHMCAdaptor(n_adapts, Preconditioner(metric), NesterovDualAveraging(0.8, int))
 
 # Draw samples via simulating Hamiltonian dynamics
 # - `samples` will store the samples
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -67,9 +67,9 @@ Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round(l.ϵ; sigdigits=
 ### Jittering
 
 mutable struct JitteredLeapfrog{T<:AbstractFloat} <: AbstractLeapfrog{T}
-    ϵ0      ::  T
+    ϵ0      ::  T  # nominal (unjittered) step size
     jitter  ::  T
-    ϵ       ::  T
+    ϵ       ::  T  # current (jittered) step size
 end
 
 JitteredLeapfrog(ϵ0, jitter) = JitteredLeapfrog(ϵ0, jitter, ϵ0)
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -17,7 +17,7 @@ nom_step_size(i::AbstractIntegrator) = step_size(i)
 abstract type AbstractLeapfrog{T} <: AbstractIntegrator end
 
 step_size(lf::AbstractLeapfrog) = lf.ϵ
-jitter!(::AbstractRNG, lf::AbstractLeapfrog) = lf
+jitter(::AbstractRNG, lf::AbstractLeapfrog) = lf
 temper(lf::AbstractLeapfrog, r, ::NamedTuple{(:i, :is_half),<:Tuple{Integer,Bool}}, ::Int) = r
 stat(lf::AbstractLeapfrog) = (step_size=step_size(lf), nom_step_size=nom_step_size(lf))
 
@@ -30,7 +30,6 @@ function step(
     fwd::Bool=n_steps > 0   # simulate hamiltonian backward when n_steps < 0,
 ) where {T<:AbstractFloat}
     n_steps = abs(n_steps)  # to support `n_steps < 0` cases
-    jitter!(rng, lf)
     ϵ = fwd ? step_size(lf) : -step_size(lf)
 
     @unpack θ, r = z
@@ -66,7 +65,7 @@ Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round(l.ϵ; sigdigits=
 
 ### Jittering
 
-mutable struct JitteredLeapfrog{T<:AbstractFloat} <: AbstractLeapfrog{T}
+struct JitteredLeapfrog{T<:AbstractFloat} <: AbstractLeapfrog{T}
     ϵ0      ::  T  # nominal (unjittered) step size
     jitter  ::  T
     ϵ       ::  T  # current (jittered) step size
@@ -81,9 +80,9 @@ end
 nom_step_size(lf::JitteredLeapfrog) = lf.ϵ0
 
 # Jitter step size; ref: https://github.com/stan-dev/stan/blob/1bb054027b01326e66ec610e95ef9b2a60aa6bec/src/stan/mcmc/hmc/base_hmc.hpp#L177-L178
-function jitter!(rng::AbstractRNG, lf::JitteredLeapfrog)
-    lf.ϵ = lf.ϵ0 * (1 + lf.jitter * (2 * rand(rng) - 1))
-    return lf
+function jitter(rng::AbstractRNG, lf::JitteredLeapfrog)
+    ϵ = lf.ϵ0 * (1 + lf.jitter * (2 * rand(rng) - 1))
+    return reconstruct(lf, ϵ=ϵ)
 end
 
 ### Tempering
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -12,7 +12,22 @@ abstract type AbstractIntegrator end
 NesterovDualAveraging(δ::AbstractFloat, i::AbstractIntegrator) = NesterovDualAveraging(δ, nom_step_size(i))
 
 stat(::AbstractIntegrator) = NamedTuple()
+
+"""
+    nom_step_size(::AbstractIntegrator)
+
+Get the nominal integration step size. The current integration step size may
+differ from this, for example if the step size is jittered. Nominal step size is
+usually used in adaptation.
+"""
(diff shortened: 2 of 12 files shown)


sethaxen:
I'm not certain why Travis isn't detecting this PR

xukai92:
Just see if reopening it will trigger Travis.

sethaxen:
Thanks! Any advice on how to handle `jitter`?

sethaxen:
What if the integrator `step`s returned a `Transition` instead of a `PhasePoint`? That would enable other experimental integrators to also return stats we don't know about.

xukai92:
I prefer not because `Transition` is designed to store a transition from a MCMC kernel.

How about the following
1. define `stats(::AbstractLeapfrog) = ...` （similar to L10 and L11 for `jitter` and `temper`)
    - This can then be used in `transition` to grab info from integrator
2. add another filed `ϵ0` to `JitteredLeapfrog` and make it mutable
3. change `jitter` to `jitter!` to "refresh" `ϵ` using `ϵ0`

How do you think?

sethaxen:
That's a convenient way to do it. I implemented it.

I misunderstood what PyMC3's `step_size_bar` is. It's an internal parameter during the adaptation phase. `step_size` in PyMC3 and Stan indicates the nominal (unjittered) step size, and they don't output the jittered one. Still, I think that would be useful to have in stats. I think I'll keep the nominal step size as `step_size` and for `JitteredLeapfrog` only output `step_size_jittered`.

Adaptation uses the nominal step size, so I'll also add a `nominal_step_size` method. It also seems that a common pattern is `NesterovDualAveraging(0.8, integrator.ϵ)`. I'm also planning to add something like
```julia
NesterovDualAveraging(δ::AbstractFloat, i::AbstractIntegrator) = NesterovDualAveraging(δ, nominal_step_size(i))
```

sethaxen:
> > Adaptation uses the nominal step size, so I'll also add a nominal_step_size method.
> 
> Can you explain a bit more? I don't quite catch this point.

So when one sets the step size, it's really setting a "nominal" step size, since if one jitters for example, the step size can change from step to step. It's the nominal step size that should get adapted (this is how Stan does it). You'd also want to use the nominal step size for HMCDA. If there's a more descriptive term than "nominal", I'm happy to use it; that's just the term Stan uses.

sethaxen:
One more point tangential to this PR is that the way Stan and (I think) PyMC3 implement HMC, the step size is only jittered at the beginning of the transition, not at every step in the transition. As currently implemented, `JitteredLeapfrog` is going to jitter every step.

src/integrator.jl
```diff
@@ -59,17 +66,25 @@ Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round(l.ϵ; sigdigits=
 
 ### Jittering
 
-struct JitteredLeapfrog{T<:AbstractFloat} <: AbstractLeapfrog{T}
-    ϵ       ::  T
+mutable struct JitteredLeapfrog{T<:AbstractFloat} <: AbstractLeapfrog{T}
+    ϵ0      ::  T
```
xukai92: Let's add a comment here to indicate this is the nominal one.

src/trajectory.jl
```diff
@@ -149,22 +151,25 @@ function transition(
     h::Hamiltonian,
     z::PhasePoint
 ) where {T<:Real}
+    H0 = energy(z)
     z′ = step(rng, τ.integrator, h, z, τ.n_steps)
     # Accept via MH criteria
-    is_accept, α = mh_accept_ratio(rng, energy(z), energy(z′))
+    is_accept, α = mh_accept_ratio(rng, H0, energy(z′))
```
xukai92: Why do we change this again? Seems unnecessary.

src/trajectory.jl
```diff
@@ -149,22 +151,25 @@ function transition(
...
-    return Transition(z, stat)
+    H = energy(z)
+    tstat = merge(
+        (n_steps=τ.n_steps,
+         is_accept=is_accept,
+         acceptance_rate=α,
+         log_density=z.ℓπ.value,
+         hamiltonian_energy=H,
+         hamiltonian_energy_error=H - H0),
+        stat(τ.integrator)
```
xukai92: Alignment.

src/trajectory.jl
sethaxen: We need H0 down below to compute `hamiltonian_energy_error`, so we cache it above and reuse it here.

src/trajectory.jl
sethaxen: I think alignment here is correct. It's aligned with the open parenthesis of the previous item passed to `merge`.

xukai92:
> One more point tangential to this PR is that the way Stan and (I think) PyMC3 implement HMC, the step size is only jittered at the beginning of the transition, not at every step in the transition. As currently implemented, JitteredLeapfrog is going to jitter every step.

You are right. It's the case for NUTS but HMC and HMCDA are fine as there is only a single call for the integrator. But let's also do jitter one time for a transition by moving it to the beginning of `transition`. It's fine now as we only have leapfrog integrator and we can worry about generalise it later.

Also I think what we could improve is that we don't need make `JitteredLeapfrog` mutable anymore because for `jitter` we can just return a new leapfrog and use that within the `transition` function. So it's acutally has a nice side effect :)

src/trajectory.jl
xukai92: Oh OK I see you intention. It's OK to keep as it but actually `energy(z)` caches the computation already as the energy is computed when `z` is constructed.

src/trajectory.jl
xukai92: Then I guess the other one is mis-aligned. I just notice they are different but didn't look carefully.

src/trajectory.jl
```diff
@@ -441,17 +455,21 @@ function transition(
...
+        (n_steps=tree.nα,
+         is_accept=true,
+         acceptance_rate=tree.sum_α / tree.nα,
+         log_density=zcand.ℓπ.value,
+         hamiltonian_energy=H,
+         hamiltonian_energy_error=H - H0,
+         max_hamiltonian_energy_error=tree.ΔH_max,
+         tree_depth=j,
+         numerical_error=termination.numerical),
+         stat(τ.integrator)
```
xukai92: This one.

xukai92: Maybe consider move the `),` from the previous line to a new line?

sethaxen:
> But let's also do jitter one time for a transition by moving it to the beginning of `transition`.

> Also I think what we could improve is that we don't need make `JitteredLeapfrog` mutable anymore because for `jitter` we can just return a new leapfrog and use that within the `transition` function. So it's acutally has a nice side effect :)

So just to be clear, you're recommending 1) making `JitteredLeapfrog` immutable again and renaming `jitter!` back to `jitter` 2) removing `jitter` from `step`, and 3) adding `jitter` to the beginning of `transition`, returning a new `JitteredLeapfrog` to be used within that transition. Did I get that right?

xukai92:
Yes exactly!

src/trajectory.jl
sethaxen: Got it. I figured just for generality it would be nice to keep it this way so that this function isn't dependent on `energy`'s implementation details.

sethaxen:
Okay, unless there's anything else, I think this is ready to merge.

README.md
```diff
@@ -43,7 +43,7 @@ metric = DiagEuclideanMetric(D)
 h = Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
 int = Leapfrog(find_good_eps(h, θ_init))
 prop = NUTS{MultinomialTS,GeneralisedNoUTurn}(int)
-adaptor = StanHMCAdaptor(n_adapts, Preconditioner(metric), NesterovDualAveraging(0.8, int.ϵ))
+adaptor = StanHMCAdaptor(n_adapts, Preconditioner(metric), NesterovDualAveraging(0.8, int))
```
yebai: This looks like a breaking change. If so, maybe consider adding a special `NesterovDualAveraging` constructor to maintain backward compatibility, and print a warning message?

src/trajectory.jl
```diff
@@ -16,6 +16,8 @@ struct Transition{P<:PhasePoint, NT<:NamedTuple}
     stat    ::  NT
 end
 
+stat(t::Transition) = t.stat
```
yebai: Consider renaming `stat`  to `stats`?

src/trajectory.jl
```diff
@@ -185,8 +193,9 @@ function transition(
     z::PhasePoint
 ) where {T<:Real}
     # Create the corresponding static τ
-    n_steps = max(1, floor(Int, τ.λ / τ.integrator.ϵ))
-    static_τ = StaticTrajectory(τ.integrator, n_steps)
+    n_steps = max(1, floor(Int, τ.λ / nom_step_size(τ.integrator)))
+    integrator = jitter(rng, τ.integrator)
```
xukai92: This `jitter` call is duplicate. As the later `transition` call already jitters the step size.

README.md
xukai92: I think the original construct is still there. It's a conveint wrapper for it.

src/trajectory.jl
sethaxen: I think the benefit of `stat` over `stats` is consistency with how the terms are used within the package. `stats` is currently being used for a vector of things like what `stat()` produces. I don't much care though.

README.md
sethaxen: Yes @xukai92's right, this won't break anything.

src/trajectory.jl
xukai92: Yes I do use the convention that `stats` is for a list of `stat`.

xukai92:
PR looks good to me now!

xukai92:
I will merge it in 2 hours if no objection.

Issue #118, opened 2019-10-24
Add NUTS diagnostics 


trappmartin:
Several people on Slack asked for better divergence diagnostics of NUTS.

Kusti Skyten wrote:
Divergent transitions https://mc-stan.org/bayesplot/articles/visual-mcmc-diagnostics.html#divergent-transitions are useful for choosing between equivalent formulations of a model with a better geometry for the posterior. Here's all the diagnostics in DHMC: https://github.com/tpapp/DynamicHMC.jl/blob/master/src/diagnostics.jl. AHMC seems to only have EBFMI (haven't looked that carefully at the code though).

Seth Axen wrote:
I also remember the output of AHMC's EBFMI very different from e.g. ArviZ. I haven't looked at it in detail yet though.

jonasmac16:
#50

jonasmac16:
#50

xukai92:
I believe the cited Stan article or Sec. 6.2 of Betancourt  corresond to the `numerical_error` filed of elements in `stats`

sethaxen:
I confirmed that there is a problem with the implementation of `EBFMI`. Its values are huge (it's supposed to be between 0 and 1 IIR, with lower than 0.3 being considered quite bad). It's an easy fix though. The numerator is supposed to be a mean, so it just needs to be divided by `length(Es) - 1`.

xukai92:
Yes you are right. I should not have used `var` in the denumerator actually... Should be a summation too. Nice catch!

xukai92:
In other words, it should be off by `1 / length(Es)` (not `1 / (length(Es) - 1)`).

sethaxen:
The numerator is a summation over `n=length(Es)-1`, while the denominator is a summation over `n+1`. `var` in Julia defaults to `correction=true`, which would introduce a factor of `n`, so no, I think `var` is appropriate. FWIW, this is how `var` in R works and that's what RStan's implementation uses.

xukai92:
Good point. I will make a PR to fix it.

sethaxen:
Is `numerical_error` now completely equivalent to an indicator of divergent transition?

xukai92:
If an `Inf` or `NaN` is treated as an infinte density then yes.

PR #125, opened 2019-11-03
Fix EBFMI bug

xukai92:
Address https://github.com/TuringLang/AdvancedHMC.jl/issues/118#issuecomment-549121503


sethaxen:
Any idea what's causing those errors? I'm seeing them also on #124, and I don't think here or there they have anything to do with the changes made.

xukai92:
It seems that after https://github.com/TuringLang/AdvancedHMC.jl/pull/120 is merged the preconditioner is less numerical stable... I will look into why. Let's merge this PR first anyway.

xukai92:
I mean #121

PR #127, opened 2019-11-05
Allow more versions for dependencies

andreasnoack:
In particular we need to allow `StatsBase` 0.32 but I think it would be good to loosen the other ones as well.


xukai92:
Thanks for the PR.

PR #130, opened 2019-11-08
Bugfix for tempered Leapfrog

xukai92:


src/integrator.jl
```diff
@@ -119,6 +119,6 @@ Tempering step. `step` is a named tuple with
 - `is_half` indicating whether or not it's (the first) half momentum/tempering step
 """
 function temper(lf::TemperedLeapfrog, r, step::NamedTuple{(:i, :is_half),<:Tuple{Integer,Bool}}, n_steps::Int)
-    i_temper = 2(step.i - 1) + 1 + step.is_half    # counter for half temper steps
+    i_temper = 2(step.i - 1) + 1 + !step.is_half    # counter for half temper steps
```
xukai92: Quite dumb bug. I was thinking maybe we should use the other appraoch in which we generate the tempering schedule before all leapfrog steps and simply apply them. With this appraoch we can simply "print out" the tempering schedule. And this appraoch is more general as the tempering schedule could be more flexible.

yebai: we can probably do a dry run and print out the tempering variables without changing the code.

test/integrator.jl
```diff
@@ -42,12 +42,30 @@ end
...
+@testset "temper" begin
+    αsqrt = 2.0
+    lf = TemperedLeapfrog(ϵ, αsqrt ^ 2)
+    r = ones(5)
+    r1 = AdvancedHMC.temper(lf, r, (i=1, is_half=true), 3)
+    r2 = AdvancedHMC.temper(lf, r, (i=1, is_half=false), 3)
+    r3 = AdvancedHMC.temper(lf, r, (i=2, is_half=true), 3)
+    r4 = AdvancedHMC.temper(lf, r, (i=2, is_half=false), 3)
+    r5 = AdvancedHMC.temper(lf, r, (i=3, is_half=true), 3)
+    r6 = AdvancedHMC.temper(lf, r, (i=4, is_half=false), 3)
```
yebai: Is `i=4` intended here? Or should it be `i=3`?

src/integrator.jl
xukai92: We could. But how about using a speficic tempering schedule. E.g. allow user to provides any sequnece of alphas, e.g. TemperedLeapfrog(ϵ, [1.2,1.1,1.05,1.025])?

test/integrator.jl
xukai92: It's a typo. I will fix it and also add a boundary check for temper.

src/integrator.jl
yebai: Yeah, but maybe keep the current code for now, and change it when we find a good motivating use case.

PR #132, opened 2019-11-09
Allow latest version of StatsFuns and prepare for bugfix release

andreasnoack:
Because StatsFuns didn't have bound on SpecialFunctions in version 0.8, some users might currently experience breakage if StatsFuns upgrade is hold back.


xukai92:
Thanks for the PR!

Issue #110, opened 2019-10-04
Improve numerical error message clarity


yebai:
The current numerical error message looks like
```julia
┌ Warning: The current proposal will be rejected due to numerical error(s).
│   isfiniteθ = true
│   isfiniter = false
│   isfiniteℓπ = false
│   isfiniteℓκ = false
```

Maybe the following is slightly more neat and clearer?

```julia

┌ Warning: The current proposal will be rejected due to numerical error(s).
│   isfinite.((θ,r,ℓπ,ℓκ)) = (true, false, false, false)
```

xukai92:
Related issue https://github.com/TuringLang/Turing.jl/issues/974#issuecomment-555395341

yebai:
Closed by #136

PR #136, opened 2019-12-01
Improve numerical error message

xukai92:
Addressed https://github.com/TuringLang/AdvancedHMC.jl/issues/110 and fix a function signature for Julia 1.3


xukai92:
I will merge this when CI passes.

Issue #103, opened 2019-09-20
Unify `Trajectory` type for static, dynamic HMC samplers. 


yebai:
We have the following nice composable interface for `NUTS` at the moment

```julia
p = NUTS{Multinomial,GeneralisedNoUTurn}(int)
```
where `NUTS` is a subtype of `DynamicTrajectory`. Maybe we can also unify `StaticTrajectory` and `DynamicTrajectory` via introducing a `Static` type and  a `MetropolisTS` trajectory sampler (with the final point as an MH proposal), i.e.

```julia
struct Static{T<:Integer} <: AbstractTerminationCriterion
    n_leapfrog::T
end

struct MetropolisTS <: AbstractTrajectorySampler end

# dynamic HMC with Multinomial sampling and generalised No-U-Turn. 
nuts = Trajectory{MultinomialTS,GeneralisedNoUTurn}(int)

# static HMC with a Multinomial sampling
hmc_Multinomial = Trajectory{MultinomialTS,Static}(int)

# standard HMC with an MH correction step
hmc= Trajectory{MetropolisTS,Static}(int)

```

We can also use integration time instead of a number of leapfrog steps in the `Static` type. This allows us to use `Trajectory` to construct all common HMC samplers including HMC, HMCDA and NUTS.

xukai92:
I will do https://github.com/TuringLang/AdvancedHMC.jl/issues/102 first and see how to unify them later.

xukai92:
After working on the spiral idea for a bit I realise that what we really need to do is treat the transition kernel as a seperate type.
```julia
sturct Kernel{T<:AbstractTrajectory,S<:AbstractTrajectorySampler,F<:AbstractRefreshment}
t::T
s::S
f::F
end
```

The current `NUTS{MultinomialTS,GeneralisedNoUTurn}` should be `Kernel(NUTS{GeneralisedNoUTurn}, MultinomialTS())` where `f` by default is set to be the completely refreshment.

By explicitly dispatch over `transition()` over `Kernel`, we can also the compelte control of a complete transition.

PS: We can also provide a composite kernel somehow. This is useful to combine a multinomial step + a MH step. Or maybe it's overcomplicated and we could introduce a specific type for this.

yebai:
Closed by https://github.com/TuringLang/AdvancedHMC.jl/pull/245

PR #142, opened 2019-12-06
RFC static to support MultinomialTS

xukai92:
Support of `HMC{MultinomialTS}` (#102)


src/Kernel/static.jl
```diff
@@ -0,0 +1,101 @@
...
+    #       We can also copy `z′ and avoid mutating the original `z′`. But this is
+    #       not efficient and immutability of `z′` is not important in this local scope.
+    return z′
+end
+
+### Last from trajecory
+
+samplecand(::AbstractRNG, τ::HMC{LastTS}, h, z) = step(τ.integrator, h, z, τ.n_steps)
+
+### Multinomial sampling from trajecory
```
xukai92: @yebai The main change is here.

src/integrator.jl
```diff
@@ -69,6 +75,36 @@ function step(
     end
     return z
 end
+# TODO: merge step and steps using Task
+function steps(
```
xukai92: @yebai Another main change is here.

xukai92:
@yebai I pointed out the two main changes. Other changes are mostly rename of files and moving around codes.

src/Kernel/static.jl
xukai92: (and the function below)

src/integrator.jl
xukai92: This implementation is pretty naive now. I guess we can unify `step` and `steps` by implementing the leapfrog using `Task`. But I'd leave it for another PR.

yebai: I'm slightly confused here. It does not seem to add any new functionality. The old `step` function can handle both 1-step and multiple-step leapfrog integration.

src/sampler.jl
```diff
@@ -15,7 +15,7 @@ function sample_init(
 end
 
 # A step is a momentum refreshment plus a transition
-function step(
+function transit(
```
yebai: Here it does not seem to add anything by changing the function name.  I would keep the old function name if there are no strong reasons. The function name `step` is more intuitive than `transit` -- `transit` sounds a bit mysterious, I think. Plus, this is a breaking change, although currently only depreciated.

src/AdvancedHMC.jl
```diff
@@ -29,7 +29,7 @@ end
 # θ: position variables / model parameters
 # r: momentum variables
 
-include("adaptation/Adaptation.jl")
+include("Adaptation/Adaptation.jl")
```
yebai: Maybe use a lower case name for a directory, which is easier to navigate on terminals.

src/integrator.jl
xukai92: This function is to **return** all points during the integration.

src/sampler.jl
xukai92: Sure. I can revert this.

xukai92:
There is some weird thing happening to the folder name on git. I will try to fix them before merging.

src/AdvancedHMC.jl
```diff
@@ -42,7 +42,7 @@ export Hamiltonian
 include("integrator.jl")
 export Leapfrog, JitteredLeapfrog, TemperedLeapfrog
 include("trajectory.jl")
-export StaticTrajectory, find_good_eps, HMCDA, NUTS, SliceTS, MultinomialTS, ClassicNoUTurn, GeneralisedNoUTurn
+export HMC, HMCDA, NUTS, LastTS, SliceTS, MultinomialTS, ClassicNoUTurn, GeneralisedNoUTurn, find_good_eps
```
yebai: `LastTS` is a bit vague, maybe consider renaming to `EndPointTS`?

src/adaptation/stan_adaption.jl
```diff
@@ -107,5 +107,6 @@ function adapt!(
     end
 end
 
-# Deprecated constructor
+# Deprecation
```
yebai: Minor comment: perhaps "Deprecated functions" is better here.

src/integrator.jl
```diff
@@ -36,12 +36,24 @@ jitter(::Union{AbstractRNG,AbstractVector{<:AbstractRNG}}, lf::AbstractLeapfrog)
...
+    return θ, r, value, gradient
+end
+
 function step(
     lf::AbstractLeapfrog{T},
     h::Hamiltonian,
     z::PhasePoint,
     n_steps::Int=1;
-    fwd::Bool=n_steps > 0   # simulate hamiltonian backward when n_steps < 0,
+    fwd::Bool=n_steps > 0,  # simulate hamiltonian backward when n_steps < 0,
```
yebai: why adding a comma after the last argument?

src/integrator.jl
yebai: I see. Maybe we can support this via adding a keyword argument to the original function, e.g. 

```julia
function step(
    lf::AbstractLeapfrog{T},
    h::Hamiltonian,
    z::PhasePoint,
    n_steps::Int=1;
    fwd::Bool=n_steps > 0,  # simulate hamiltonian backward when n_steps < 0,
    zs::Union{Vector{PhasePoint}, PhasePoint} = z # change to Vector{PhasePoint}(undef, n_steps) otherwise.
) where {T<:AbstractScalarOrVec{<:AbstractFloat}}
    n_steps = abs(n_steps)  # to support `n_steps < 0` cases
    ϵ = fwd ? step_size(lf) : -step_size(lf)
    ϵ = ϵ'

    @unpack θ, r = z
    @unpack value, gradient = ∂H∂θ(h, θ)

    for i = 1:n_steps
        # Tempering
        r = temper(lf, r, (i=i, is_half=true), n_steps)
        # Single leapfrog step
        θ, r, value, gradient = step(lf, ϵ, h, θ, r, value, gradient)
        # Tempering
        r = temper(lf, r, (i=i, is_half=false), n_steps)
        # Create a new phase point by caching the logdensity and gradient
        z = phasepoint(h, θ, r; ℓπ=DualValue(value, gradient))
        !isfinite(z) && break

      if zs  isa Vector 
        zs[i] = z
      else 
        zs = z
      end

    end
    return zs
end
```

This would avoid the clutter of introducing two additional `step` functions.

xukai92:
> Here a more informative name is Trajectory, I think. See also: #103

I agree. I think it should be `StaticTrajectory`. For the same reason, `NUTS` should not be existi, but just `DynamicTrajectory` with the type of stoping criteria specified. The current change is just to make static ones consistent with dynamic ones but I will have another PR to change them together.

src/integrator.jl
xukai92: Just personal formatting style. Always having a comma makes it easy to add or remove args. I can change it if you think it's confusing.

src/integrator.jl
xukai92: Using `zs` to name a single point is confusing. I have an implementation which is nice. I will push it later.

src/integrator.jl
yebai: sure, sounds good to remove it.

src/sampler.jl
xukai92: OK so I reverted it. But again it would be nice to have a seperate name so that `step` for leapfrog is namly different from `step` for MCMC transition.

src/integrator.jl
xukai92: Apparently the channel approach causes some unexpected errors which I cannot find the root. I will implement your version first for now.

xukai92:
@yebai Are you happy to merge this?

yebai:
> I agree. I think it should be StaticTrajectory. For the same reason, NUTS should not be existi, but just DynamicTrajectory with the type of stoping criteria specified. The current change is just to make static ones consistent with dynamic ones but I will have another PR to change them together.

Yes, ready to merge. Before merging, perhaps revert the changes to `StaticTrajectory` before doing your proposed refactoring? This is because the current PR introduce a temporary breaking change, which might be confusing.

xukai92:
> Before merging, perhaps revert the changes to

Sure. I will do it.

PR #146, opened 2019-12-14
WIP: Setup DiffEq common interface integrator

ChrisRackauckas:
This is just a very quick implementation. Tests and everything are needed, but this should be a good guide to anyone. I set it up so it just defaults to velocity verlet, but since it's setup with the DynamicalODEProblem any symplectic integrator will work. I think I grabbed the step size and everything correctly. It would probably be a good idea to do this through Requires.jl.


src/integrator.jl
```diff
@@ -76,6 +76,41 @@ function step(
     res
 end
 
+struct DiffEqIntegrator{A}
+  alg::A
+  DiffEqIntegrator(alg=VelocityVerlet()) = new(alg)
```
xukai92: Should be `DiffEqIntegrator(alg::A=VelocityVerlet()) where {A} = new{A}(alg)`

xukai92:
Thanks Chris. This is very helpful. Can you fix the syntax error that fails loading of AHMC? I will merge this PR then and work from here.

Project.toml
```diff
@@ -7,6 +7,7 @@ ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
 InplaceOps = "505f98c9-085e-5b2c-8e89-488be7bf1f34"
 LazyArrays = "5078a376-72f3-5289-bfd5-ec5146d43c02"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
+OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
```
yebai: Loading `OrdinaryDiffEq` optionally using `Requires` sounds useful here, since we want to keep AHMC dependency footprint low.

src/AdvancedHMC.jl
```diff
@@ -48,4 +48,6 @@ include("diagnosis.jl")
 include("sampler.jl")
 export sample
 
+include("init.jl")
```
xukai92:
```suggestion
using Requires
include("init.jl")
```

Project.toml
```diff
@@ -22,6 +22,7 @@ Parameters = "0.10, 0.11, 0.12"
 ProgressMeter = "1"
 StatsBase = "0.31, 0.32"
 StatsFuns = "0.8, 0.9"
+Requires = "1"
```
xukai92: We also need `Requires = "ae029012-a4dd-5104-9daa-d747884805df"` in `[deps]`

xukai92:
@yebai I think it's OK to merge this and we can work from this PR.

PR #152, opened 2020-01-02
Work around compiler bug - Julia issue 34232

mohamed82008:
This PR works around the bug reported in https://github.com/JuliaLang/julia/issues/34232. I don't quite understand why there is a problem to begin with, but with this PR, everything works fine.

### DIFF
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -654,7 +654,7 @@ function mh_accept_ratio(
     return accept, α
 end
 function mh_accept_ratio(
-    rng::AbstractVector{<:AbstractRNG},
+    rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
     Horiginal::AbstractVector{<:T},
     Hproposal::AbstractVector{<:T},
 ) where {T<:AbstractFloat}
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -641,19 +641,25 @@ find_good_eps(
     max_n_iters::Int=100
 ) where {T<:AbstractFloat} = find_good_eps(GLOBAL_RNG, h, θ; max_n_iters=max_n_iters)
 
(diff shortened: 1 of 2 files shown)


mohamed82008:
If tests pass, let's merge and release this as soon as possible to continue my work on DynamicPPL.

mohamed82008:
The test failure seems to be a seed issue.

src/trajectory.jl
```diff
@@ -641,21 +641,29 @@ find_good_eps(
...
 """
 Perform MH acceptance based on energy, i.e. negative log probability.
 """
 function mh_accept_ratio(
-    rng::Union{AbstractRNG,AbstractVector{<:AbstractRNG}},
-    Horiginal::AbstractScalarOrVec{<:T},
-    Hproposal::AbstractScalarOrVec{<:T}
+    rng::AbstractRNG,
+    Horiginal::T,
+    Hproposal::T,
```
yebai: This looks good to me. 

@xukai92  I think the old implementation has some type ambiguity, because it does not guarantee variables `rnd`, `Horiginal` and `Hproposal` to be scalar or vector at the same time. The type inconsistency among these variables might be the cause of the compiler error (JuliaLang/julia#34232).

mohamed82008: I think it's more subtle than that. For one, the error didn't show up before when DynamicPPL was not a thing. Also, dispatch is not entirely the issue as I show in `mh2` in JuliaLang/julia#34232.

mohamed82008: Also, the function actually runs correctly even with the compiler errors there, and no run-time ambiguity error is thrown, so...it's most likely not an everyday ambiguity error.

mohamed82008: Btw, this error is only on Julia 1.0 and 1.1. Julia 1.2 and 1.3 work fine.

PR #161, opened 2020-01-13
Improve utility

xukai92:
Add default support for ForwardDiff and Zygote when constructing `Hamiltonian`


src/init.jl
```diff
@@ -48,4 +48,84 @@ function __init__()
...
+            res, back = Zygote.pullback(ℓπ, θ)
+            return res, back(fill(1, size(θ)...))[1]
+        end
+
+        # NOTE
+        # It's important for Zygote to come after ForwardDiff otherwise
+        # the function def below will be override by ForwardDiff's def.
+        # This is probably because Zygote is dependent on ForwardDiff and
+        # they will be also available at the same time.
+        function Hamiltonian(metric, ℓπ)
```
devmotion: That seems wrong since it will override ForwardDiff's definition - so if both Zygote and ForwardDiff are loaded, it's never possible to actually use the constructor for ForwardDiff?

src/init.jl
```diff
@@ -48,4 +48,84 @@ function __init__()
...
+        import .Zygote
+
+        function ∂ℓπ∂θ_zygote(ℓπ, θ::AbstractVector)
+            res, back = Zygote.pullback(ℓπ, θ)
+            return res[1], back(1)[1]
+        end
+
+        function ∂ℓπ∂θ_zygote(ℓπ, θ::AbstractMatrix)
+            res, back = Zygote.pullback(ℓπ, θ)
+            return res, back(fill(1, size(θ)...))[1]
```
devmotion:
```suggestion
            return res, back(ones(Int, size(θ)))[1]
```

src/init.jl
```diff
@@ -48,4 +48,84 @@ function __init__()
...
+        # end
+
+        function Hamiltonian(metric, ℓπ)
+            ∂ℓπ∂θ(θ::AbstractVecOrMat) = ∂ℓπ∂θ_forwarddiff(ℓπ, θ)
+            return Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
+        end
+
+    end
+
+    @require Zygote = "e88e6eb3-aa80-5325-afca-941959d7151f" begin
```
devmotion: I guess one could just depend on [ZygoteRules](https://github.com/FluxML/ZygoteRules.jl) since `Zygote.pullback` is actually imported from `ZygoteRules.pullback`, so `@require` should not be needed.

In general, I'm wondering if it would be more convenient to define something along the lines of
```julia
struct LogPDFWrapper{Tlogπ,T∂logπ∂θ}
    ℓπ::Tlogπ
    ∂ℓπ∂θ::T∂logπ∂θ
end
```
with all the required functions to evaluate the log pdf, it's gradient, and the gradients of the kinetic energy. This could replace the two existing fields in `Hamiltonian`. An advantage (or rather the main motivation) of this approach would be that it allows to define
```julia
struct ForwardDiffLogPDFWrapper{Tlogπ}
    ℓπ::Tlogπ
end
```
(maybe even with an additional cache) and
```julia
struct ZygoteLogPDFWrapper{Tlogπ}
    ℓπ::Tlogπ
end
```
that can be used in `Hamiltonian` instead of a custom `LogPDFWrapper`. For these different structs it would be easy to define different implementations of, e.g., `∂ℓπ∂θ(wrapper, θ::AbstractVecOrMat)` such as
```julia
∂ℓπ∂θ(wrapper::LogPDFWrapper, θ::AbstractVecOrMat) = wrapper.∂ℓπ∂θ(θ)

function ∂ℓπ∂θ(wrapper::ForwardDiffPDFWrapper, θ::AbstractVector)
    res = DiffResults.GradientResult(θ)
    ForwardDiff.gradient!(res, wrapper.ℓπ, θ)
    return DiffResults.value(res), DiffResults.gradient(res)
end

function ∂ℓπ∂θ(wrapper::ForwardDiffLogPDFWrapper, θ::AbstractMatrix)
    jacob = similar(θ)
    res = DiffResults.JacobianResult(similar(θ, size(θ, 2)), jacob)
    ForwardDiff.jacobian!(res, wrapper.ℓπ, θ)
    jacob_full = DiffResults.jacobian(res)

    d, n = size(jacob)
    for i in 1:n
        jacob[:,i] = jacob_full[i,1+(i-1)*d:i*d]
    end
    return DiffResults.value(res), jacob
end

function ∂ℓπ∂θ(wrapper::ZygoteLogPDFWrapper, θ::AbstractVector)
    res, back = ZygoteRules.pullback(wrapper.ℓπ, θ)
    return res[1], back(1)[1]
end

function ∂ℓπ∂θ(wrapper::ZygoteLogPDFWrapper, θ::AbstractMatrix)
    res, back = ZygoteRules.pullback(wrapper.ℓπ, θ)
    return res, back(ones(Int, size(θ)))[1]
end
```
In the implementation of `Hamiltonian` one would then just call `∂ℓπ∂θ(hamiltonian.logpdfwrapper, θ)` and don't care about the type of wrapper. Similarly, with this approach it should be easy to exploit the special structure of the wrappers (and caches etc) for other functions that are evaluated when running HMC.

Depending on the purpose and use of `Hamiltonian` (I'm not very familiar with this package :stuck_out_tongue: ) one could also move this design up to the level of `Hamiltonian`s and define `Hamiltonian`, `ForwardDiffHamiltonian` and `ZygoteHamiltonian` that behave like a `Hamiltonian` with the corresponding wrappers proposed above.

README.md
```diff
@@ -46,13 +38,10 @@ n_samples, n_adapts = 12_000, 2_000
 
 # Define metric space, Hamiltonian, sampling method and adaptor
 metric = DiagEuclideanMetric(D)
-h = Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
+h = Hamiltonian(metric, ℓπ) # Hamiltonian(metric, ℓπ, ∂ℓπ∂θ) for hand-coded gradient ∂ℓπ∂θ
```
yebai: I suggest we drop some nonstandard acronyms, 

```suggestion
hamiltonian = Hamiltonian(metric, ℓπ) # Or, Hamiltonian(metric, ℓπ, ∂ℓπ∂θ) for hand-coded gradient ∂ℓπ∂θ
```

README.md
```diff
@@ -46,13 +38,10 @@ n_samples, n_adapts = 12_000, 2_000
...
 int = Leapfrog(find_good_eps(h, θ_init))
-prop = NUTS{MultinomialTS,GeneralisedNoUTurn}(int)
-adaptor = StanHMCAdaptor(
-    Preconditioner(metric), 
-    NesterovDualAveraging(0.8, int)
-)
+prop = NUTS{MultinomialTS, GeneralisedNoUTurn}(int)
```
yebai:
```suggestion
proposal = NUTS{MultinomialTS, GeneralisedNoUTurn}(int)
```

README.md
```diff
@@ -46,13 +38,10 @@ n_samples, n_adapts = 12_000, 2_000
 
 # Define metric space, Hamiltonian, sampling method and adaptor
 metric = DiagEuclideanMetric(D)
-h = Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
+h = Hamiltonian(metric, ℓπ) # Hamiltonian(metric, ℓπ, ∂ℓπ∂θ) for hand-coded gradient ∂ℓπ∂θ
 int = Leapfrog(find_good_eps(h, θ_init))
```
yebai:
```suggestion
integrator = Leapfrog(find_good_eps(h, θ_init))
```

src/adaptation/stepsize.jl
```diff
@@ -125,7 +125,7 @@ function adapt_stepsize!(da::NesterovDualAveraging{T}, α::AbstractScalarOrVec{<
     # TODO: we might want to remove this when all other numerical issues are correctly handelled
     if any(isnan.(ϵ)) || any(isinf.(ϵ))
         @warn "Incorrect ϵ = $ϵ; ϵ_previous = $(da.state.ϵ) is used instead."
-        # TODO: this revert is buggy for batch mode
+        # FIXME: this revert is buggy for batch mode
```
yebai: Maybe create an issue to track this.

src/sampler.jl
```diff
@@ -154,7 +152,8 @@ function sample(
         tstat = merge(tstat, (is_adapt=isadapted,))
         # Update progress meter
         if progress
-            pm_next!(pm, tstat, i, h.metric)
+            # Do include current iteration and mass matrix
+            pm_next!(pm, (iterations=i, tstat..., mass_matrix=h.metric))
```
yebai: this is cleaner than the current implementation!

test/demo.jl
```diff
@@ -24,13 +16,10 @@ n_samples, n_adapts = 12_000, 2_000
 
 # Define metric space, Hamiltonian, sampling method and adaptor
 metric = DiagEuclideanMetric(D)
-h = Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
+h = Hamiltonian(metric, ℓπ) # Hamiltonian(metric, ℓπ, ∂ℓπ∂θ) for hand-coded gradient ∂ℓπ∂θ
```
yebai: similarly, pls consider dropping these nonstandard acronyms.

src/init.jl
xukai92: > I guess one could just depend on ZygoteRules since Zygote.pullback is actually imported from ZygoteRules.pullback, so @require should not be needed.

I don't want to depend on ZygoteRules either. But I guess this will solve the issue of order.

---

Thanks for the suggestions. I would prefer not complicate the interfaces here, i.e. introducing `ForwardDiffHamiltonian`, `ZygoteHamiltonian`, etc.

The purpose of this PR is to introduce some untilities for users who don't want to construct the gradient function. In fact, if they want to explicit use some backend, they can also use the original signature.

I have another idea which can keep the interface simpel while allow user to switch AD (when both Zygote and ForwardDiff is loaded). It also captures ambiguity. 

```julia
using AdvancedHMC, Zygote, ForwardDiff
...
Hamiltonian(metric, ℓπ)
```
gives
```
MethodError: Hamiltonian(metric, ℓπ) is ambiguous because both Zygote and ForwardDiff is loaded. Please use Hamiltonian(metric, ℓπ, ZygoteAD) or Hamiltonian(metric, ℓπ, ForwardDiffAD) explictly.
```

When none of them is loaded, it will give
```
ERROR: No AD backend is loaded. Please load Zygote or ForwardDiff before calling Hamiltonian(metric, ℓπ).
```

I will push my commits in a bit.

test/demo.jl
xukai92: Fixed.

src/init.jl
```diff
@@ -95,17 +99,22 @@ function __init__()
         #     return v, g
         # end
 
-        function Hamiltonian(metric, ℓπ)
+        function Hamiltonian(metric::AbstractMetric, ℓπ, ::Union{ForwardDiffAD, Type{ForwardDiffAD}})
```
devmotion: I would decide on either a type or an instance (see https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-confusion-about-whether-something-is-an-instance-or-a-type-1). According to the style guide preferably an instance, I guess.

src/init.jl
```diff
@@ -116,16 +125,13 @@ function __init__()
             return res, back(ones(Int, size(θ)))[1]
         end
 
-        # NOTE
-        # It's important for Zygote to come after ForwardDiff otherwise
-        # the function def below will be override by ForwardDiff's def.
-        # This is probably because Zygote is dependent on ForwardDiff and
-        # they will be also available at the same time.
-        function Hamiltonian(metric, ℓπ)
+        function Hamiltonian(metric::AbstractMetric, ℓπ, ::Union{ZygoteAD, Type{ZygoteAD}})
```
devmotion: Same here.

src/init.jl
```diff
@@ -95,17 +99,22 @@ function __init__()
...
         # end
 
-        function Hamiltonian(metric, ℓπ)
+        function Hamiltonian(metric::AbstractMetric, ℓπ, ::Union{ForwardDiffAD, Type{ForwardDiffAD}})
             ∂ℓπ∂θ(θ::AbstractVecOrMat) = ∂ℓπ∂θ_forwarddiff(ℓπ, θ)
             return Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
-        end
+        end        
+
+        export ForwardDiffAD
```
devmotion: ForwardDiffAD exists in Turing:
https://github.com/TuringLang/Turing.jl/blob/aad1aee6b66095edaadcc9eb05129cd882609a25/src/core/ad.jl#L28
So either it should not be exported or it should be imported in Turing, I guess.

src/init.jl
```diff
@@ -116,16 +125,13 @@ function __init__()
...
-        # the function def below will be override by ForwardDiff's def.
-        # This is probably because Zygote is dependent on ForwardDiff and
-        # they will be also available at the same time.
-        function Hamiltonian(metric, ℓπ)
+        function Hamiltonian(metric::AbstractMetric, ℓπ, ::Union{ZygoteAD, Type{ZygoteAD}})
             ∂ℓπ∂θ(θ::AbstractVecOrMat) = ∂ℓπ∂θ_zygote(ℓπ, θ)
             return Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
         end
 
+        export ZygoteAD
```
devmotion: Same problem as above, ZygoteAD will be defined in Turing as well:
https://github.com/TuringLang/Turing.jl/blob/93f1ff5891185eace12ec529c22dc2367a1a7946/src/core/ad.jl#L35

xukai92:
@devmotion Here is a summary of current behaviour of AD utilites:

1. No AD
```julia
using AdvancedHMC
hamiltonian = Hamiltonian(UnitEuclideanMetric(1), x -> x^2)
```
will give
```julia
ERROR: MethodError: no method matching Hamiltonian(metric::AbstractMetric, ℓπ) because no backend is loaded. Please load an AD package (Zygote or ForwardDiff) first.
```

2. One AD
```julia
using AdvancedHMC, ForwardDiff
hamiltonian = Hamiltonian(UnitEuclideanMetric(1), x -> x^2)
```
will work fine but
```julia
using AdvancedHMC, Zygote
hamiltonian = Hamiltonian(UnitEuclideanMetric(1), x -> x^2)
```
still thorws
```julia
ERROR: MethodError: Hamiltonian(metric::AbstractMetric, ℓπ) is ambiguous because multiple AD pakcages are available (ForwardDiff and Zygote). Please use AD explictly. Candidates:
  Hamiltonian(metric, ℓπ, ForwardDiffAD())
  Hamiltonian(metric, ℓπ, ZygoteAD())
```
as `using Zygote` alsways use `ForwardDiff`. I didn't end up using the `ZygoteRules` trick because it's pretty weird to ask people do `using ZygoteRules`. Although it would be nice we can fix this (i.e. if user only does `using Zygote`, we use it as default), the recommanded way is do the explict call - this PR is just utility.

3. Multiple AD
```julia
using AdvancedHMC, ForwardDiff, Zygote
hamiltonian = Hamiltonian(UnitEuclideanMetric(1), x -> x^2)
```
will give
```julia
ERROR: MethodError: Hamiltonian(metric::AbstractMetric, ℓπ) is ambiguous because multiple AD pakcages are available (ForwardDiff and Zygote). Please use AD explictly. Candidates:
  Hamiltonian(metric, ℓπ, ForwardDiffAD())
  Hamiltonian(metric, ℓπ, ZygoteAD())
```
In this case `Hamiltonian(metric, ℓπ, ForwardDiffAD())` and `Hamiltonian(metric, ℓπ, ZygoteAD())` will work fine.

src/init.jl
xukai92: AHMC is not dependent on Turing. This PR is not internal. It's utilites for users who wants to interact with AHMC directly instead of using it through Turing.

src/init.jl
xukai92: I will fix this.

src/init.jl
xukai92: There is no change for Turing because the AD control is explicty there when constructing `Hamiltonian`.

xukai92: Oh sorry I just got your point of not exporting. Let me think of it.

devmotion: Sure, I see why you want to export it and that Turing does not use the convenience constructor. I was just wondering if the singletons of the same name that exist in Turing might lead to problems.

yebai:
> In this case `Hamiltonian(metric, ℓπ, ForwardDiffAD())` and `Hamiltonian(metric, ℓπ, ZygoteAD())` will work fine.

Maybe just use the module name to select AD backends, e.g. 

- `Hamiltonian(metric, ℓπ, ForwardDiff)`
- `Hamiltonian(metric, ℓπ, Zygote)`

This avoids introducing unnecessary type names and exports.

src/AdvancedHMC.jl
```diff
@@ -48,6 +48,26 @@ include("diagnosis.jl")
 include("sampler.jl")
 export sample
 
+### AD utilities
```
yebai: Maybe consider moving this block into `contrib/ad.jl` as well.

xukai92:
> > In this case `Hamiltonian(metric, ℓπ, ForwardDiffAD())` and `Hamiltonian(metric, ℓπ, ZygoteAD())` will work fine.
> 
> Maybe just use the module name to select AD backends, e.g.
> 
> * `Hamiltonian(metric, ℓπ, ForwardDiff)`
> * `Hamiltonian(metric, ℓπ, Zygote)`
> 
> This avoids introducing unnecessary type names and exports.

David also pointed it. I'm fixing it now.

src/AdvancedHMC.jl
xukai92: Sure. I will move them.

src/adaptation/stepsize.jl
xukai92: https://github.com/TuringLang/AdvancedHMC.jl/issues/162

PR #166, opened 2020-01-18
Small improvements

xukai92:


src/trajectory.jl
```diff
@@ -676,7 +677,7 @@ function mh_accept_ratio(
     Horiginal::T,
     Hproposal::T,
 ) where {T <: AbstractFloat}
-    α = min(1.0, exp(Horiginal - Hproposal))
+    α = min(one(T), exp(Horiginal - Hproposal))
     accept = rand(rng) < α
```
xukai92: @yebai I think this is a bug. We should have generate a vector of uniform R.V.s otherwise the chains are explicitly coupled. Do you agree?

yebai: yes,  this should be a vector of IID uniform RVs. I thought `rng` is a vector of random streams, which should return a vector right?

src/trajectory.jl
```diff
@@ -205,13 +205,14 @@ function accept_phasepoint!(z::T, z′::T, is_accept::Bool) where {T<:PhasePoint
...
-        z′.ℓκ.value[(!).(is_accept)] = z.ℓκ.value[(!).(is_accept)]
-        z′.ℓκ.gradient[:,(!).(is_accept)] = z.ℓκ.gradient[:,(!).(is_accept)]
+    is_reject = (!).(is_accept)
+    if any(is_reject)
+        z′.θ[:,is_reject] = z.θ[:,is_reject]
+        z′.r[:,is_reject] = z.r[:,is_reject]
+        z′.ℓπ.value[is_reject] = z.ℓπ.value[is_reject]
+        z′.ℓπ.gradient[:,is_reject] = z.ℓπ.gradient[:,is_reject]
+        z′.ℓκ.value[is_reject] = z.ℓκ.value[is_reject]
+        z′.ℓκ.gradient[:,is_reject] = z.ℓκ.gradient[:,is_reject]
```
yebai: nice improvement!

src/trajectory.jl
```diff
@@ -686,7 +687,8 @@ function mh_accept_ratio(
     Horiginal::AbstractVector{<:T},
     Hproposal::AbstractVector{<:T},
 ) where {T<:AbstractFloat}
-    α = min.(1.0, exp.(Horiginal .- Hproposal))
+    α = min.(one(T), exp.(Horiginal .- Hproposal))
+    # FIXME: is this a bug?
     accept = _rand(rng) .< α
```
yebai: similarly, this should generate a vector of IID uniform RVs.

src/trajectory.jl
xukai92: This one does not. The function signature is `rng::AbstractRNG`.

src/trajectory.jl
xukai92: Only when `rng isa AbstractVector{<:AbstractRNG}}`.

src/trajectory.jl
yebai: Shall we fix this from this PR?

src/trajectory.jl
yebai: Right.

xukai92: It's actually this line (`rand(rng::AbstractRNG)` for `Horiginal ::AbstractVector{<: AbstractFloat}`) is buggy. Fixed now.

xukai92:
Ready to merge when CI passes.

xukai92:
I will merge this in 12 hours.

yebai:
Don't' merge yet -  there is still a small bug around I think. I'll take a look later.

xukai92:
@yebai Where is the possible bug you have in mind?

src/trajectory.jl
```diff
@@ -686,13 +687,13 @@ function mh_accept_ratio(
...
-    α = min.(1.0, exp.(Horiginal .- Hproposal))
-    accept = _rand(rng) .< α
+    α = min.(one(T), exp.(Horiginal .- Hproposal))
+    accept = _rand(rng, length(Horiginal)) .< α
     return accept, α
 end
 
-_rand(rng::AbstractRNG) = rand(rng)
-_rand(rng::AbstractVector{<:AbstractRNG}) = rand.(rng)
+_rand(rng::AbstractRNG, n_chains::Int) = rand(rng, n_chains)
```
yebai: This line draws `n_chains` uniform variables from a shared RNG, although the other parts of the sampling process use independent RNGs correctly. This shared RNG for acceptance steps leads to a subtle coupling of vectorised chains, which might cause bias.

yebai: Maybe force the user to pass a vector of `RNG`s here, and throw an error if not.

yebai:
> @yebai Where is the possible bug you have in mind?

@xukai92  Pls see comments above.

src/trajectory.jl
xukai92: > This shared RNG for acceptance steps leads to a subtle coupling of vectorised chains, which might cause bias.

I actually discussed with Charles on this and in the end he convinced me that it probably doesn't matter. His argument in brief is below
- What we worry here is the coupling of accepting or rejecting caused by using a shared random seed. That is, e.g., for two chains, `u_1` and `u_2` are coupled in `[u_1, u_2] = rand(2)`.
- This should not be a concern for pseudo random number generators as numbers drawn should be i.i.d (by assuming perfect randomness) anyway.
- If this is a concern, we should also worry that, say, some dimensions of the momentum variables drawn are coupled with some other; but we don't.

Maybe we should get some evidence that this really happens before forcing users to do so?

yebai: Can you add a comment in the code stating this subtle coupling in case we want to revisit this in the future?  Once that is done I'm ready to merge this PR as is.

xukai92: Just added.

PR #180, opened 2020-02-21
bump compat entry for ArgCheck

simeonschaub:
ArgCheck 2.0 is only breaking in how it works together with OptionalArgChecks, so that shouldn't affect this package at all. Might be worth considering to setup CompatHelper for this repo, which would automatically create PRs in cases like this.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "AdvancedHMC"
 uuid = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
-version = "0.2.20"
+version = "0.2.21"
 
 [deps]
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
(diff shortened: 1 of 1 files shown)


yebai:
Thanks @simeonschaub and @andreasnoack!

simeonschaub:
Would it be possible to tag a new release as well?

Project.toml
```diff
@@ -1,6 +1,6 @@
 name = "AdvancedHMC"
 uuid = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
-version = "0.2.20"
+version = "0.2.21"
```
yebai: @JuliaRegistrator register

andreasnoack: I think you'd have to do it as a commit comment. It would be great if you could use this one https://github.com/TuringLang/AdvancedHMC.jl/commit/20d5350fc6bc60fde6c5cb31243d222ab11bacb7 as it would also include the LazyArrays change

PR #181, opened 2020-02-26
Allow LazyArrays version 0.15

andreasnoack:


andreasnoack:
I'll suggest that we enable CompatHelper to automate this process.

yebai:
> I'll suggest that we enable CompatHelper to automate this process.

Somehow, we didn't enable compat helper for this repo. But this is fixed by https://github.com/TuringLang/AdvancedHMC.jl/pull/182

PR #185, opened 2020-03-04
Improving vectorized mode 

xukai92:
This PR improves vectorized mode
- [x] Compatibility of `MultinomialTS`
- [x] Revert of phase points
- [x] Bugfixes for some utilites like `show`


src/trajectory.jl
```diff
@@ -227,22 +227,52 @@ samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z) = step(τ.integrator, h,
...
     i = 0
     while cp < u
-        cp += p[i +=1]
+        cp += p[i+=1]
     end
-    return xs[max(i, 1)]
+    return max(i, 1)
+end
+
+function randcat(rng, P_unorm::AbstractMatrix{T}) where {T}
```
yebai:
```suggestion
function randcat(rng::AbstractRNG, P_unorm::AbstractMatrix{T}) where {T}
```

src/trajectory.jl
```diff
@@ -688,18 +718,15 @@ function mh_accept_ratio(
     Hproposal::AbstractVector{<:T},
 ) where {T<:AbstractFloat}
     α = min.(one(T), exp.(Horiginal .- Hproposal))
-    accept = _rand(rng, length(Horiginal)) .< α
+    # NOTE: There is a chance that sharing the RNG over multiple
+    #       chains for accepting / rejecting might couple
+    #       the chains. We need to revisit this more rigirously 
+    #       in the future. See discussions at 
+    #       https://github.com/TuringLang/AdvancedHMC.jl/pull/166#pullrequestreview-367216534
+    accept = rand(rng, T, length(Horiginal)) .< α
```
yebai: This is potentially confusing since `rnd` in the argument has a type `rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}}`. We can simplify by enforcing a constraint that the user has to pass a vector for vectorised HMC, and a scalar otherwise. Any reason why we want to keep support for `rng` being an `AbstractRNG` type in vectorised HMC?

src/trajectory.jl
```diff
@@ -227,22 +227,52 @@ samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z) = step(τ.integrator, h,
...
 function samplecand(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
     zs = step(τ.integrator, h, z, τ.n_steps; res=[z for _ in 1:abs(τ.n_steps)])
     ℓws = -energy.(zs)
+    if eltype(ℓws) <: AbstractVector
+        ℓws = hcat(ℓws...)
+    end
     ℓws = ℓws .- maximum(ℓws)
-    p_unorm = exp.(ℓws)
-    return randcat(rng, zs, p_unorm / sum(p_unorm))
+    prob_unorm = exp.(ℓws)
```
yebai: Maybe switch to `ℓprob_unorm` here, as the log scale is numerically more stable, particularly when some candidates have a tiny weight while others have a large one.

src/trajectory.jl
```diff
@@ -227,22 +227,52 @@ samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z) = step(τ.integrator, h,
...
+    u = rand(rng, T, size(P, 1))
+    C = cumsum(P; dims=2)
+    is = convert.(Int, vec(sum(C .< u; dims=2)))
+    return max.(is, 1)
+end
+
+randcat(rng::AbstractRNG, xs, p::AbstractVector) = xs[randcat(rng, p)]
+
+# xs is in the form of Vector{Vector} and has shape [n_steps][]
+function randcat(rng, xs, P_unorm::AbstractMatrix)
```
yebai: Similiar here.
```suggestion
function randcat(rng, zs::VectorOrMat{PhasePoint}, P_unorm::AbstractMatrix)
```

src/trajectory.jl
```diff
@@ -227,22 +227,52 @@ samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z) = step(τ.integrator, h,
...
+
+function randcat(rng, P_unorm::AbstractMatrix{T}) where {T}
+    P = P_unorm ./ sum(P_unorm; dims=2)
+    u = rand(rng, T, size(P, 1))
+    C = cumsum(P; dims=2)
+    is = convert.(Int, vec(sum(C .< u; dims=2)))
+    return max.(is, 1)
+end
+
+randcat(rng::AbstractRNG, xs, p::AbstractVector) = xs[randcat(rng, p)]
```
yebai: Maybe change `xs` into `zs` for consistency? Also adding type annotation makes it more readable here.
```suggestion
randcat(rng::AbstractRNG, zs::Vector{PhasePoint}, p::AbstractVector) = xs[randcat(rng, p)]
```

src/trajectory.jl
xukai92: Yes but it would be `Union{AbstractRNG, AbstractVector{<:AbstractRNG}}`

src/trajectory.jl
xukai92: One reason is for to be compatible with `Random.seed!`, `Random.GLOBAL_RNG` is reset. In this case, `rng = GLOBAL_RNG` controls all randomness. Also I think it makes sense to control the randomness using a single RNG in general.

src/trajectory.jl
xukai92: Good point.  Let me change all to log domain.

src/trajectory.jl
xukai92: `zs` is always a vector of phase points. In vectorized mode, a single `z` stores all phase points for all chains for a particular step. I think this is a bit conterintuitive and I was trying to figure out a way to improve this. In particular, maybe we should all multiple steps to be sotred in a single `PhasePoint`. This would make a low of mutiple-chain/multiple-step operations easier. But I think it worth another PR with more discussions.

yebai: > In particular, maybe we should all multiple steps to be sotred in a single PhasePoint. 

I'm not sure, saving more than one (position, momentum) pairs in a PhasePoint would make PhasePoint and leapfrog related code complicated. Maybe there is a better approach.

xukai92: It will certainly make the code more complicated. But I think any solution have to do so, no matter in which manner, to achive high perfomrance.

xukai92:
@yebai Anything else we want to include in this PR?

src/trajectory.jl
yebai: That's true - it might still be good for consistency to always keep things in the log scale to avoid accidental mistreatments. What do you think?

src/trajectory.jl
yebai: Perhaps we can keep this discussion for future PRs.

src/trajectory.jl
```diff
@@ -227,8 +227,8 @@ samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z) = step(τ.integrator, h,
 
 ### Multinomial sampling from trajecory
 
-function randcat(rng::AbstractRNG, p_unorm::AbstractVector{T}) where {T}
-    p = p_unorm ./ sum(p_unorm)
+function randcat(rng::AbstractRNG, up::AbstractVector{T}) where {T}
```
yebai: Maybe rename `up` to `unnormlised_prob` for readability? `up` is really hard to parse...

src/trajectory.jl
```diff
@@ -238,30 +238,30 @@ function randcat(rng::AbstractRNG, p_unorm::AbstractVector{T}) where {T}
...
     return max.(is, 1)
 end
 
-randcat(rng::AbstractRNG, xs, p::AbstractVector) = xs[randcat(rng, p)]
+randcat(rng::AbstractRNG, zs::AbstractVector{<:PhasePoint}, p::AbstractVector) = zs[randcat(rng, p)]
 
-# xs is in the form of Vector{Vector} and has shape [n_steps][]
-function randcat(rng, xs, P_unorm::AbstractMatrix)
-    x = similar(first(xs))
+# zs is in the form of Vector{PhasePoint{Matrix}} and has shape [n_steps][dim, n_chains]
```
yebai:
```suggestion
# zs is in the form of Vector{Matrix{PhasePoint}} and has shape [n_steps][dim, n_chains]
```

src/trajectory.jl
yebai: > Also I think it makes sense to control the randomness using a single RNG in general.

Can we automatically generate random seeds for `n_chains` using a separate global RNG stream?  This way all randomness is controlled by a single global RNG, but still, keep RNG independent for separate chains.

src/trajectory.jl
xukai92: Sounds good.

src/trajectory.jl
xukai92: That's a good solution!

src/trajectory.jl
xukai92: Yes I think it would make the code clear. Let's assume `randcat` takes an unnormalised log probability then.

xukai92: If so I guess we should renmae `randcat` to something else that indicates it takes probabilities in log space.

Another question is do we want to keep probability in log space even within `randcat`? If so it would requires change all accumation of probabities to take use of `logaddexp`, which I assume would be slower than simply doing addition or `cumsum`.

src/trajectory.jl
xukai92: I took a look into how to remove all places with `Union{AbstractRNG, AbstractVector{<:AbstractRNG}}` and some of them requires some works on the metrics space part. I think this is out of the scope of this PR and worth a seperate one because it actually influences the whole code space. I created an issue (https://github.com/TuringLang/AdvancedHMC.jl/issues/186) to track this.

src/trajectory.jl
yebai: It's ok to work with ordinary probabilities in `randcat` I think. We can revisit this later if it causes numerical instabilities.

src/trajectory.jl
yebai: Sounds good to me.

src/trajectory.jl
xukai92: Cool. I renamed the function which takes log probabilities as `randcatlog` to indicate the input is in log domain.

src/AdvancedHMC.jl
```diff
@@ -26,7 +26,7 @@ function Base.randn(rng::AbstractVector{<:AbstractRNG}, T, dim::Int, n_chains::I
     return cat(randn.(rng, T, dim)...; dims=2)
 end
 
-randcatlog(rng::AbstractRNG, unnorm_ℓp::AbstractVector{T}) where {T} =
+randcatlog(rng::AbstractRNG, unnorm_ℓp::AbstractVector) =
```
yebai: `randcatlog` is not more readable than `randcat`... 
Maybe a docstring explaining that `randcat` assumes log probabilities is what we need here.

xukai92: I think this information should be reflected by the name. Basically `randcatlog` and `randcat` are two different functions. Even with a docstring it would make the code confusing when reading something like `randcat(ℓp)` but for `randcatlog(ℓp)` at least we would then realise what it is doing from the naming.

xukai92: Maybe there is a better name than `randcatlog`

xukai92:
This PR is ready to merge once the naming of `randcat` is resolved. I think we should have something which reflects the probability in log domain, maybe `randlogcat` is better than `randcatlog`?

yebai:
> This PR is ready to merge once the naming of randcat is resolved. I think we should have something which reflects the probability in log domain, maybe randlogcat is better than randcatlog?


Maybe `randcat_logp`?

xukai92:
Sounds good. I just made the change.

PR #188, opened 2020-03-26
Merge Preconditioner into Welford estimators

### LINKED ISSUE #174: Merge metric and preconditoner

xukai92:
Resolves #174 with the main changes below

- `Preconditioner` and related types are merged to Welford estimators
  - ~The immutable and mutable fields of Welford estimators are seperated as in `NesterovDualAveraging`~
- Users are now encouraged to use `MassMatrixAdaptor(::Metric)` and `StepSizeAdaptor(::Float, ::Integrator)`
  - The old methods are deprecated.

The related codes are also moved around a bit.

- `precond.jl` is renamed to `massmatrix.jl`
- Codes related to metric are now back to it's own `metric.jl`
- Compositional adaptors are now all in `composite.jl`


xukai92:
For reviewing purpose, I temperoraly renamed `massmatrix.jl` back to `precond.jl` so that GitHub can show the diff.

src/adaptation/Adaptation.jl
```diff
@@ -1,75 +1,59 @@
...
-adapt!(
-    adaptor::T,
-    θ::AbstractVecOrMat{<:AbstractFloat},
-    α::AbstractScalarOrVec{<:AbstractFloat}
-) where {T<:AbstractAdaptor} = error("`adapt!(adaptor::$T, θ::AbstractVecOrMat{<:AbstractFloat}, α::AbstractScalarOrVec{<:AbstractFloat})` is not implemented.")
-reset!(adaptor::T) where {T<:AbstractAdaptor} = error("`reset!(adaptor::$T)` is not implemented.")
-initialize!(adaptor::T, n_adapts::Int) where {T<:AbstractAdaptor} = throw(MethodError(initialize!, adaptor, n_adapts))
-finalize!(adaptor::T) where {T<:AbstractAdaptor} = error("`finalize!(adaptor::$T)` is not implemented.")
+function getM⁻¹ end
+function getϵ end
```
devmotion: Are users supposed to work with or call these functions? IMO names of user-facing functions should not include fancy unicode characters. Additionally, a name such as `stepsize` or `getstepsize` would be more descriptive.

If it's not user-facing, it doesn't matter that much, but still more descriptive names might be easier to understand :man_shrugging:

src/adaptation/Adaptation.jl
```diff
@@ -1,75 +1,59 @@
...
 
-# TODO: implement consensus adaptor
+StepSizeAdaptor(δ::AbstractFloat, i::AbstractIntegrator) = 
+    NesterovDualAveraging(δ, nom_step_size(i))
 
-##
-## Compositional adaptor
-## TODO: generalise this to a list of adaptors
-##
+@deprecate NesterovDualAveraging(δ, i::AbstractIntegrator) StepSizeAdaptor(δ, i)
```
devmotion: I'd suggest collecting all deprecations at the end of the file. IIRC that's recommended, and IMO in that way it is easier to spot the deprecated functions and remove them in one of the upcoming releases.

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
 #### Robust online (co-)variance estimators.
 ####
 
-## Variance estimator
+abstract type MassMatrixAdaptor <: AbstractAdaptor end
 
-abstract type VarEstimator{T} end
+initialize!(adaptor::T, n_adapts::Int) where {T<:MassMatrixAdaptor} = nothing
+finalize!(adaptor::T) where {T<:MassMatrixAdaptor} = nothing
```
devmotion: I'd suggest removing the static type parameters:

```suggestion
initialize!(::MassMatrixAdaptor, ::Int) = nothing
finalize!(::MassMatrixAdaptor) = nothing
```

src/adaptation/composite.jl
```diff
@@ -70,12 +99,10 @@ function StanHMCAdaptor(
     init_buffer::Int=75,
     term_buffer::Int=50,
     window_size::Int=25
-) where {M<:AbstractPreconditioner}
+) where {M<:MassMatrixAdaptor}
```
devmotion: I'd suggest removing the static type parameter and just writing `pc::MassMatrixAdaptor`, similar to `ssa:StepSizeAdaptor`.

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
 #### Robust online (co-)variance estimators.
 ####
 
-## Variance estimator
+abstract type MassMatrixAdaptor <: AbstractAdaptor end
```
devmotion: It seems all subtypes of `MassMatrixAdaptor` have a type parameter `T` - would it make sense to include it already to `MassMatrixAdaptor` or even `AbstractAdaptor`?

src/adaptation/Adaptation.jl
```diff
@@ -1,75 +1,59 @@
...
 
-struct NaiveHMCAdaptor{M<:AbstractPreconditioner, Tssa<:StepSizeAdaptor} <: AbstractAdaptor
-    pc  :: M
-    ssa :: Tssa
-end
-Base.show(io::IO, a::NaiveHMCAdaptor) = print(io, "NaiveHMCAdaptor(pc=$(a.pc), ssa=$(a.ssa))")
+MassMatrixAdaptor(m::UnitEuclideanMetric{T}) where {T} = 
+    UnitMassMatrix(T)
```
devmotion: IMO
```suggestion
    UnitMassMatrix{T}()
```
feels more natural if `T` is just a type parameter.

src/adaptation/Adaptation.jl
```diff
@@ -1,75 +1,59 @@
...
+MassMatrixAdaptor(m::DiagEuclideanMetric{T}) where {T} = 
+    WelfordVar(T, size(m); var=copy(m.M⁻¹))
```
devmotion: Same here, I guess you could directly write
```suggestion
    WelfordVar{T}(size(m); var=copy(m.M⁻¹))
```

src/adaptation/Adaptation.jl
```diff
@@ -1,75 +1,59 @@
...
+MassMatrixAdaptor(m::DenseEuclideanMetric{T}) where {T} = 
+    WelfordCov(T, size(m); cov=copy(m.M⁻¹))
```
devmotion: And here as well
```suggestion
    WelfordCov{T}(size(m); cov=copy(m.M⁻¹))
```

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
-    nc.n += 1
-    push!(nc.S, s)
-end
+Base.show(io::IO, ::UnitMassMatrix) = print(io, "UnitMassMatrix")
 
-function reset!(nc::NaiveVar{T}) where {T}
-    nc.n = 0
-    nc.S = T()
-end
+UnitMassMatrix(::Type{T}=Float64) where {T} = UnitMassMatrix{T}()
```
devmotion: As suggested above, IMO it would be more natural to define
```suggestion
UnitMassMatrix() = UnitMassMatrix{Float64}()
```
and just call `UnitMassMatrix{T}()` explicitly if possible.

devmotion: IMO that makes it clearer that `T` is the actual type parameter.

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
+)
+    resize!(ve, θ)
+    push!(ve, θ)
+    if ve.n >= ve.n_min && is_update
+        ve.var .= getest(ve)
+    end
+end
+
+# NOTE: this naive variance estimator is used only in testing
+mutable struct NaiveVar{E<:AbstractVecOrMat{<:AbstractFloat},T<:AbstractVector{E}} <: VarEstimator{T}
```
devmotion: Does it have to be mutable?

devmotion: I saw you increase `n` below, but `n` just seems to be the length of the vector `S`. I'd suggest just calling `length(S)` if needed and not saving `n`.

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
-    nc.n += 1
-    push!(nc.S, s)
-end
+NaiveVar(sz::Tuple{Vararg{Int}}; kwargs...) = NaiveVar(Float64, sz; kwargs...)
 
-function get_cov(nc::NaiveCov)
-    @assert nc.n >= 2 "Cannot get covariance with only one sample"
-    return Statistics.cov(nc.S)
+function Base.push!(nv::NaiveVar, s::AbstractVecOrMat)
+    nv.n += 1
```
devmotion: As mentioned above, if possible I'd remove `n`:
```suggestion
```

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
+    push!(nv.S, s)
 end
 
-function reset!(nc::NaiveCov{T}) where {T}
-    nc.n = 0
-    nc.S = T()
+function reset!(nv::NaiveVar{T}) where {T}
+    nv.n = 0
```
devmotion: Same here:
```suggestion
```

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
-    return Statistics.cov(nc.S)
+function Base.push!(nv::NaiveVar, s::AbstractVecOrMat)
+    nv.n += 1
+    push!(nv.S, s)
 end
 
-function reset!(nc::NaiveCov{T}) where {T}
-    nc.n = 0
-    nc.S = T()
+function reset!(nv::NaiveVar{T}) where {T}
```
devmotion:
```suggestion
function reset!(nv::NaiveVar)
```

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
+    nv.n = 0
+    nv.S = T()
```
devmotion: It seems `nv` could be immutable (see above) if you use
```suggestion
    resize!(nv.S, 0)
```

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
 end
 
-function add_sample!(wc::WelfordCov, s::AbstractVector)
-    wc.n += 1
-    @unpack δ, μ, n, M = wc
-    δ .= s .- μ
-    μ .+= δ ./ n
-    M .+= (s .- μ) .* δ'
-end
+Base.show(io::IO, ::WelfordVar) = print(io, "WelfordVar")
```
devmotion: Maybe include some (important) fields here?

src/adaptation/precond.jl
devmotion: I'd also suggest making the element type a type parameter rather (or in addition to) `E`:
```suggestion
struct NaiveVar{S<:AbstractFloat,T<:AbstractVector{<:AbstractVecOrMat{S}}} <: VarEstimator{T}
```
Maybe it would also be more important to know the element type in `VarEstimator`? Then it should rather be `VarEstimator{S}`.

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
+
+# NOTE: this naive variance estimator is used only in testing
+mutable struct NaiveVar{E<:AbstractVecOrMat{<:AbstractFloat},T<:AbstractVector{E}} <: VarEstimator{T}
     n :: Int
     S :: T
 end
 
-NaiveCov(::Type{T}, sz::Union{Tuple{Int}}) where {T<:AbstractFloat} = NaiveCov(0, Vector{Vector{T}}())
+NaiveVar(::Type{T}, sz::Tuple{Int}) where {T<:AbstractFloat} = NaiveVar(0, Vector{Vector{T}}())
+NaiveVar(::Type{T}, sz::Tuple{Int,Int}) where {T<:AbstractFloat} = NaiveVar(0, Vector{Matrix{T}}())
```
devmotion: If the element type `T` would be a type parameter instead of/in addition to `E` and if `n` is removed, it would be natural to write
```suggestion
NaiveVar{T}(sz::Tuple{Int}) where {T<:AbstractFloat} = NaiveVar(Vector{Vector{T}}())
NaiveVar{T}(sz::Tuple{Int,Int}) where {T<:AbstractFloat} = NaiveVar(Vector{Matrix{T}}())
```

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
 
-function add_sample!(nc::NaiveCov, s::AbstractVector)
-    nc.n += 1
-    push!(nc.S, s)
-end
+NaiveVar(sz::Tuple{Vararg{Int}}; kwargs...) = NaiveVar(Float64, sz; kwargs...)
```
devmotion: And similarly here
```suggestion
NaiveVar(sz::Tuple{Vararg{Int}}) = NaiveVar{Float64}(sz)
```

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
-    WelfordCov(0, zeros(T, d), zeros(T, d, d), zeros(T, d))
-end
-
-function reset!(wc::WelfordCov{T}) where {T<:AbstractFloat}
-    wc.n = 0
-    wc.μ .= zero(T)
-    wc.M .= zero(T)
+# Ref： https://github.com/stan-dev/math/blob/develop/stan/math/prim/mat/fun/welford_var_estimator.hpp
+mutable struct WelfordVar{T<:AbstractVecOrMat{<:AbstractFloat}} <: VarEstimator{T}
+    n     :: Int
```
devmotion: Again, couldn't this be changed to
```suggestion
struct WelfordVar{S<:AbstractFloat,T<:AbstractVecOrMat{S}} <: VarEstimator{T}
```
?

If you need a consistent way for accessing the number of samples for different variance estimators, I'd suggest defining a function `nsamples(::WelfordVar) = ...` and so on instead of relying on a field `n`.

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
-# Ref: https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/covar_adaptation.hpp
-function get_cov(wc::WelfordCov{T}) where {T<:AbstractFloat}
-    n, M = T(wc.n), wc.M
-    @assert n >= 2 "Cannot get variance with only one sample"
-    return (n / ((n + 5) * (n - 1))) .* M + T(1e-3) * (5 / (n + 5)) * LinearAlgebra.I
+function WelfordVar(
+    ::Type{T}, sz::Union{Tuple{Int},Tuple{Int,Int}}; 
+    n_min::Int=10, var=ones(T, sz)
+) where {T<:AbstractFloat}
+    return WelfordVar(0, n_min, zeros(T, sz), zeros(T, sz), zeros(T, sz), var)
```
devmotion: Again, without `n` and with different type parameters the natural form would be
```suggestion
function WelfordVar{T}(
	sz::Union{Tuple{Int},Tuple{Int,Int}}; 
    n_min::Int=10, var=ones(T, sz)
) where {T<:AbstractFloat}
    return WelfordVar(n_min, zeros(T, sz), zeros(T, sz), zeros(T, sz), var)
```

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
 end
 
-###
-### Preconditioning matrix adaptors.
-###
+WelfordVar(sz::Tuple{Vararg{Int}}; kwargs...) = WelfordVar(Float64, sz; kwargs...)
```
devmotion: And
```suggestion
WelfordVar(sz::Union{Tuple{Int},Tuple{Int,Int}}; kwargs...) = WelfordVar{Float64}(sz; kwargs...)
```
here

src/adaptation/Adaptation.jl
```diff
@@ -1,75 +1,59 @@
 module Adaptation
+export Adaptation
```
devmotion: IMO it would simplify the import and export statements if the adaptation implementation is not in a separate module.

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
 )
-    resize!(dpc, θ)
-    add_sample!(dpc.ve, θ)
-    if dpc.ve.n >= dpc.n_min && is_update
-        # Can be made inplace
-        dpc.var .= get_var(dpc.ve)
+    resize!(ce, θ)
+    push!(ce, θ)
+    if ce.n >= ce.n_min && is_update
+        ce.cov .= getest(ce)
```
devmotion: It seems it's basically the same logic for the variance and the covariance estimators - maybe just define a function `update!(estimator)` that updates the estimators, so in one case `ve.var` and in the other case `ce.cov`. Then one could just have one `adapt!` method for both of them.

src/adaptation/precond.jl
```diff
@@ -2,389 +2,227 @@
...
-    sz::Tuple{Int};
-    n_min::Int=10,
-    covar=LinearAlgebra.diagm(0 => ones(T, sz))
-) where {T}
-    ce = WelfordCov(T, sz)
-    # TODO: take use of the line below when we have an interface to set which covariance estimator to use
-    # ce = NaiveCov(T)
-    return DensePreconditioner(n_min, ce, covar)
+# NOTE: This naive covariance estimator is used only in testing.
+mutable struct NaiveCov{T<:AbstractVector{<:AbstractVector{<:AbstractFloat}}} <: CovEstimator{T}
```
devmotion: Same comments as above, basically.

devmotion:
I haven't checked everything yet but I left some comments. Some things became a bit clearer to me while reading through your code, so I guess you can just ignore some of the first comments :smile: 

Many comments are focused around the setup of the variance and covariance estimators. If possible, I'd suggest removing the field `n` (and maybe `n_min`) and replacing them with a function `nsamples(estimator)` and `nmin(estimator)` if possible. Moreover, I'd suggest only using vectors as fields if the `resize!` functionality is required (if a matrix `theta` is provided one would work with the flattened version `vec(theta)`). The advantage would be that the estimators could be immutable structs, in which the arrays are just updated and resized.

Additionally, if one refactors the update steps `ve.var .= getest(ve)` and `ce.cov .= getest(ce)` into separate `update!` functions, it would be possible to just use a single `adapt!` function for the variance and covariance estimators, I guess.

src/adaptation/Adaptation.jl
xukai92: I agree. Originally I was trying to implement these adaptations s.t. they might be used for another library in the future. However, we are moving towards the opposite direction - making adaption more tied to AdvancedHMC.jl now. So I think it's fair to not make them as a seperate module any more.

src/adaptation/Adaptation.jl
xukai92: They are not exposed to users.

src/adaptation/Adaptation.jl
xukai92: Sounds good!

src/adaptation/Adaptation.jl
xukai92: Good point.

src/adaptation/precond.jl
xukai92: > I saw you increase n below, but n just seems to be the length of the vector S. I'd suggest just calling length(S) if needed and not saving n.

Yes it's only for naive var/covar estimator though. For Welford, `n` is not accessible. In fact naive estimators are not used but only for test purposes. But I can do your suggestion anyway.

xukai92:
Hmmm. It seems that my reply on Saturaday was not sent out!? Thanks a lot for the review. I've adapted most of the suggestions you made. I will take a pass on fixing the remaining now.

The thing for `n` in Welford estimators is that we have no access to the number of samples from the statistics we track. The main advatange of Welford is that we don't need to store all samples but we need to keep track `n` - the number of samples we collect so far. So I don't think we can remove it. For `n_min`, it's again not type specific but instances specific. Users could instantiate two different estimators with different `n_min`.

The sperate `update!` suggestion is cool. Thanks again :D

src/adaptation/precond.jl
xukai92: It seems that for naive estimaters, there is no real need of accessing the number of samples. So I just remove that from naive ones and keep whatever we have for Welford.

src/adaptation/Adaptation.jl
xukai92: Let's keep it as a module for now then.

src/integrator.jl
```diff
@@ -9,8 +9,6 @@
 
 abstract type AbstractIntegrator end
 
-NesterovDualAveraging(δ::AbstractFloat, i::AbstractIntegrator) = NesterovDualAveraging(δ, nom_step_size(i))
```
yebai: why removing this function?

src/adaptation/precond.jl
```diff
@@ -2,389 +2,211 @@
...
-end
+reset!(nv::NaiveVar) = resize!(nv.S, 0)
 
-function add_sample!(wc::WelfordCov, s::AbstractVector)
-    wc.n += 1
-    @unpack δ, μ, n, M = wc
-    δ .= s .- μ
-    μ .+= δ ./ n
-    M .+= (s .- μ) .* δ'
+function getest(nv::NaiveVar)
```
yebai: `getest` can be easily confused with `get_test`, unfortunately. Why not keep `get_var` and `get_cov` instead?

src/adaptation/composite.jl
```diff
@@ -48,12 +46,43 @@ function Base.show(io::IO, state::StanHMCAdaptorState)
...
 end
 
-################
-### Adaptors ###
-################
+### Composite adaptors
+
+abstract type CompositeAdaptor <: AbstractAdaptor end
+
+getM⁻¹(ca::CompositeAdaptor) = getM⁻¹(ca.pc)
```
xukai92: @yebai This is the only intention to introduce `CompositeAdaptor`. These functions was definted for `StanHMCAdaptor` and `NaiveHMCAdaptor` seperately. Please let me know if you think it's fine to introduce this in this PR. I'm fine to revert it as well if you think it's a bit of an over-kill for now.

src/adaptation/precond.jl
```diff
@@ -2,389 +2,211 @@
...
+    ::AbstractVecOrMat{<:AbstractFloat},
+    ::AbstractScalarOrVec{<:AbstractFloat},
+    is_update::Bool=true
+) = nothing
 
-## Covariance estimator
+## Diagonal mass matrix adaptor
 
-abstract type CovEstimator{T} end
+abstract type VarEstimator{T} <: MassMatrixAdaptor end
```
yebai: Maybe consider rename `VarEstimator` into `DiagMatrixAdaptor`.

src/metric.jl
```diff
@@ -0,0 +1,109 @@
...
+DenseEuclideanMetric(D::Int) = DenseEuclideanMetric(Float64, D)
+DenseEuclideanMetric(::Type{T}, sz::Tuple{Int}) where {T} = DenseEuclideanMetric(Matrix{T}(I, first(sz), first(sz)))
+DenseEuclideanMetric(sz::Tuple{Int}) = DenseEuclideanMetric(Float64, sz)
+
+renew(ue::DenseEuclideanMetric, M⁻¹) = DenseEuclideanMetric(M⁻¹)
+
+Base.size(e::DenseEuclideanMetric, dim...) = size(e._temp, dim...)
+Base.show(io::IO, dem::DenseEuclideanMetric) = print(io, "DenseEuclideanMetric(diag=$(_string_M⁻¹(dem.M⁻¹)))")
+
+# `rand` functions for `metric` types.
```
yebai: Ideally, `rand` is defined for `Kinect` energy type instead of `metric` types. But we can keep this as is in this PR and fix it in a separate effort.

src/sampler.jl
```diff
@@ -1,3 +1,28 @@
...
+function Parameters.reconstruct(h::Hamiltonian, adaptor::Union{MassMatrixAdaptor, CompositeAdaptor})
+    metric = renew(h.metric, getM⁻¹(adaptor))
+    return reconstruct(h, metric=metric)
+end
+
+Parameters.reconstruct(τ::AbstractProposal, ::AbstractAdaptor) = τ
+function Parameters.reconstruct(τ::AbstractProposal, adaptor::Union{StepSizeAdaptor, CompositeAdaptor})
+    # FIXME: this does not support change type of `ϵ` (e.g. Float to Vector)
+    # FIXME: this is buggy for `JitteredLeapfrog`
+    integrator = reconstruct(τ.integrator, ϵ=getϵ(adaptor))
```
yebai: Maybe add `Parameters` prefix here for consistency.

src/sampler.jl
```diff
@@ -1,3 +1,28 @@
...
+    return reconstruct(τ, integrator=integrator)
```
yebai: Similar here.

src/sampler.jl
```diff
@@ -1,3 +1,28 @@
...
+end
+
+function resize(h::Hamiltonian, θ::AbstractVecOrMat{T}) where {T<:AbstractFloat}
+    metric = h.metric
+    if size(metric) != size(θ)
+        metric = typeof(metric)(size(θ))
+        h = reconstruct(h, metric=metric)
```
yebai: Similar here.

src/adaptation/composite.jl
yebai: I find it a bit overkill and makes things slightly harder to follow here. Maybe revert it.

src/AdvancedHMC.jl
```diff
@@ -1,74 +1,49 @@
...
+
 include("integrator.jl")
 export Leapfrog, JitteredLeapfrog, TemperedLeapfrog
+
 include("trajectory.jl")
-export StaticTrajectory, HMCDA, NUTS, EndPointTS, SliceTS, MultinomialTS, ClassicNoUTurn, GeneralisedNoUTurn, find_good_eps
+export EndPointTS, SliceTS, MultinomialTS, 
+       StaticTrajectory, HMCDA, NUTS, 
+       ClassicNoUTurn, GeneralisedNoUTurn, 
+       find_good_eps
```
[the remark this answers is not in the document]
xukai92: Can do!

src/adaptation/composite.jl
xukai92: No problem for me.

src/integrator.jl
xukai92: It's moved to `adaptation.jl`. The intention is to keep all functions that are only for the purpose of user interface into the same place.

src/adaptation/precond.jl
xukai92: Just for consistency. Calling `getest(::VarEstimator)` or `getest(::CovEstimator)` is self-explianed on what is expected to return. If the issues is about confusing against `get_test`, we can do `get_estimation`. I don't mind revert to the old names as well.

src/metric.jl
xukai92: Sure. Just a side note, these codes are not changed but only moved out from `precond.jl`.

src/sampler.jl
xukai92: I think a better way to resolve this consistency of usage is by `import reconstruct` explictly in the beginning of AdvancedHMC. So that we know all the functions we extend. How do you think?

src/adaptation/Adaptation.jl
```diff
@@ -61,15 +53,40 @@ end
...
+export NaiveHMCAdaptor, StanHMCAdaptor
+
+# User helpers for AdvancedHMC
+
+using ..AdvancedHMC: 
+    AbstractIntegrator, nom_step_size,
+    AbstractMetric, UnitEuclideanMetric, DiagEuclideanMetric, DenseEuclideanMetric
+
+StepSizeAdaptor(δ::AbstractFloat, i::AbstractIntegrator) = 
+    NesterovDualAveraging(δ, nom_step_size(i))
```
yebai, line 66:
```suggestion
    StepSizeAdaptor(δ, nom_step_size(i))
```
I feel this should be in the `integrator.jl` file instead of here. Since 

1. We don't want `Adaption` to depend on HMC code for modularity. I think we discussed this a while ago.  
2.`StepSizeAdaptor` is exported by `Adaption` anyway, so it is ok to extend it.

src/adaptation/Adaptation.jl
```diff
@@ -61,15 +53,40 @@ end
...
+
+# Deprecations
 
-export adapt!, initialize!, finalize!, getϵ, getM⁻¹, reset!, renew,
-       NesterovDualAveraging,
-       UnitPreconditioner, DiagPreconditioner, DensePreconditioner,
-       AbstractMetric, UnitEuclideanMetric, DiagEuclideanMetric, DenseEuclideanMetric,
-       Preconditioner, NaiveHMCAdaptor, StanHMCAdaptor
+@deprecate StanHMCAdaptor(n_adapts, pc, ssa) initialize!(StanHMCAdaptor(pc, ssa), n_adapts)
+@deprecate NesterovDualAveraging(δ, i::AbstractIntegrator) StepSizeAdaptor(δ, i)
```
yebai, line 89: Perhaps move all the deprecation statements into a separate file.

src/adaptation/precond.jl
yebai: `get_estimation` sounds good.

src/sampler.jl
yebai: Either way works, although `Parameters.reconstruct` is a bit more explicit about `reconstruct` is not an AHMC function.

src/integrator.jl
yebai: See new comment in `adaption.jl`.

src/sampler.jl
xukai92: I agree and that's why it is written as it ATM. The convention is
- For functions to extend, we use `Module.fn` explictly
- For functions to use (those indicated by `using Module: fn1, fn2 ...` in the beginning of AHMC), we drop `Module`

However, as you point here, it leads to this inconsistency. I would prefer explicity `import` to solve it, just so it is consistency with explict `using`.

yebai: Sounds good.

src/adaptation/Adaptation.jl
xukai92: I'm OK with moving them out of `adaptation.jl` to dropthis dependency but I think they should instead go to `AdvancedHMC.jl` or even a seperate file. 

My reason is that, L65 - L91 are _helper functions for adaptors_ to for easy initialization and it makes sense to group them together.

- If we put `StepSizeAdaptor` to `integrator.jl`, we should also put `MassMatrixAdaptor` to `metric.jl`. I think it's too messy and hard to find these functions.
- When I develop `integrator.jl`, I don't really care any about how adaptors would be created.

src/adaptation/Adaptation.jl
xukai92: Can I ask why? I think the current way is fine. Functions are still located in whatever file they are in based on their functionatliy, and I can easily find all deprecated ones by searching for the keyword `@deprecate`.

xukai92: The search results is meaningful in the sense that it tells me what are the deprecated functions actually pointing to.

xukai92:
Also I will rename `precond.jl` to `massmatrix.jl` before merging if it sounds good to everyone. It is currently not renamed for the purpose of showing diff in GitHub for reviewing.

src/adaptation/Adaptation.jl
yebai: Only `StepSizeAdaptor(δ, nom_step_size(i))` should be in `integrator.jl` since it is for Jittered leapforg integrator and depends on `nom_step_size`. Other stuff can stay as is for now.

src/adaptation/Adaptation.jl
yebai: It's a minor comment. The purpose is to track all depreciated APIs from a central location, and regularly remove them when a new major/breaking release is to be released.

xukai92: As I said, it's quick easy to track all the deprecated functions by searching `@deprecate`, which is quite simple in both terminal and text editors. So I guess we don't really need it.

src/adaptation/Adaptation.jl
xukai92: I don't think it makes sense. First of all, `StepSizeAdaptor(δ, nom_step_size(i))` is for all integrators but not only for jittered leapforg. `nom_step_size` defined for all integrators - for nomral leapfrog, it's nominated step size is same as actual used ones. Secondly, `MassMatrixAdaptor` also depends on metric types. It doesn't make much difference whether things are depedend on a function or just the types IMO.

yebai: We would like to keep a one-way dependency between `Adaption` and `HMC` if possible. Before this PR, HMC depends on Adaption but not the other way in general. The `Metric` related code was part of `Adaption`'s preconditioner codes, and it might make sense to keep `Metric` part of adaption again. It is important to keep code as "simple" as possible moving forward.

yebai: At the moment, the name `metric` is a bit abused I think. We're using `metric` to refer to both mass matrices and distributions for momentum variables (Kinect energy). However, we can probably improve this in seprate PR.

xukai92: > We would like to keep a one-way dependency ...

As I said, I agree with the dependncy point. I'm proposing moving them to `AdvancedHMC.jl` or somewhere else, instead of putting them into seperate files.

> It is important to keep code as "simple" as possible moving forward.

I belive what I proposed make the code easier to manage. In my experience of this PR, keeping these interface functions in the same place is the most easiest way, but not putting them in different places.

My preference is 1 > 2 > 3 > 4

1. Keeping them in `Adaptation.jl`
2. Keeping them in a seperate file.
3. Keeping them in `AdvancedHMC.jl` right after the include of `Adaptation`
4. Keeping them in seperate files, i.e. `MassMatrixAdaptor ` into `metric.jl` and `StepSizeAdaptor` into `leapfrog.jl`

As 1 breaks the dependency point and 2 might make the repo structure more complex, I would prefer 3 instead. 4 really makes navigating the code difficult. 

When I started with this PR and try to find this definition `NesterovDualAveraging` without search, I clicked `stepsize.jl` and then `Adaptation.jl` and didn't find it, which was quite confusing. This is the reason I think the nature place to put them are not in AHMC but in Adaptation. Later after dealing with the mass matrix adaptors, I realise that these functions are just helper functions - they are actually not for functionality of either adaptation or leapfrog/metric. They are functions only for user friendly purpose. One can use AHMC without them as well,  e.g. by doing `NesterovDualAveraging(0.8, 0.01)`. But I think those function make it clear that what are the adaptor is adapting, .e.g `StepSizeAdaptor(0.8, integrator)`.

> The Metric related code was part of Adaption's preconditioner codes, and it might make sense to keep Metric part of adaption again.

I think what we did before makes sense only because `precond.jl` contains code for adaptors but not because it contains code for metrics.

yebai: Sure, let’s keep them in the file ‘AdvancedHMC.jl’ before we figure out a better way!

yebai:
It seems some tests are broken.

xukai92:
I fixed a x86 vs x64 difference bug. It passes in Travis but for some reason it still shows failure here.

xukai92:
I re-run the CI and it shows correctly now.

README.md
```diff
@@ -5,49 +5,54 @@
...
-- We presented AdvancedHMC.jl at [AABI](http://approximateinference.org/) 2019 in Vancouver, Canada. ([pdf](https://openreview.net/forum?id=rJgzckn4tH))
+- We presented a paper for AdvancedHMC.jl at [AABI](http://approximateinference.org/) 2019 in Vancouver, Canada. ([abs](http://proceedings.mlr.press/v118/xu20a.html), [pdf](http://proceedings.mlr.press/v118/xu20a/xu20a.pdf), [OpenReview](https://openreview.net/forum?id=rJgzckn4tH))
 - We presented a poster for AdvancedHMC.jl at [StanCon 2019](https://mc-stan.org/events/stancon2019Cambridge/) in Cambridge, UK. ([pdf](https://github.com/TuringLang/AdvancedHMC.jl/files/3730367/StanCon-AHMC.pdf))
 
 **API CHANGES**
+- [v0.2.22] Three functions are renamed.
+  - `Preconditioner(metric::AbstractMetric)` -> `MassMatrixAdaptor(metric)` and 
+  - `NesterovDualAveraging(δ, integrator::AbstractIntegrator)` -> `StepSizeAdaptor(δ, integrator)`
+  - `find_good_eps` -> `find_good_stepsize`
 - [v0.2.15] `n_adapts` is not needed to construct `StanHMCAdaptor`; the old constructor is deprecated.
```
yebai, line 22:
```suggestion
- [v0.2.15] `n_adapts` is no longer needed to construct `StanHMCAdaptor`; the old constructor is deprecated.
```

README.md
```diff
@@ -5,49 +5,54 @@
...
-- [v0.2.8] Two exported types are renamed: `Multinomial` -> `MultinomialTS` and `Slice` -> `SliceTS`.
+- [v0.2.8] Two types are renamed.
```
yebai, line 23:
```suggestion
- [v0.2.8] Two Hamiltonian trajectory sampling methods are renamed to avoid a name clash with Distributions.
```

README.md
```diff
@@ -57,15 +62,15 @@ samples, stats = sample(hamiltonian, proposal, initial_θ, n_samples, adaptor, n
 
 ## API and supported HMC algorithms
 
-An important design goal of `AdvancedHMC.jl` is to be modular, and support algorithmic research on HMC. 
+An important design goal of AdvancedHMC.jl is to be modular, and support algorithmic research on HMC.
```
yebai, line 65:
```suggestion
An important design goal of AdvancedHMC.jl is modularity; we would like to support algorithmic research on HMC.
```

README.md
```diff
@@ -57,15 +62,15 @@ samples, stats = sample(hamiltonian, proposal, initial_θ, n_samples, adaptor, n
...
 This modularity means that different HMC variants can be easily constructed by composing various components, such as preconditioning metric (i.e. mass matrix), leapfrog integrators,  trajectories (static or dynamic), and adaption schemes etc. 
 The minimal example above can be modified to suit particular inference problems by picking components from the list below.
 
 ### Preconditioning matrix (`metric`)
```
yebai, line 69:
```suggestion
### Hamiltonian mass matrix (`metric`)
```

README.md
```diff
@@ -79,57 +84,59 @@ where `ϵ` is the step size of leapfrog integration.
...
-- Original NUTS with slice sampling: `NUTS{SliceTS,ClassicNoUTurn}(int)`
-- Generalised NUTS with slice sampling: `NUTS{SliceTS,GeneralisedNoUTurn}(int)`
-- Original NUTS with multinomial sampling: `NUTS{MultinomialTS,ClassicNoUTurn}(int)`
-- Generalised NUTS with multinomial sampling: `NUTS{MultinomialTS,GeneralisedNoUTurn}(int)`
-
-where `int` is the integrator used.
+- Static HMC with a fixed number of steps (`n_steps`): `StaticTrajectory(integrator, n_steps)`
+- HMC with a fixed total trajectory length (`trajectory_length`): `HMCDA(integrator, trajectory_length)` 
+- Original NUTS with slice sampling: `NUTS{SliceTS,ClassicNoUTurn}(integrator)`
+- Generalised NUTS with slice sampling: `NUTS{SliceTS,GeneralisedNoUTurn}(integrator)`
```
yebai, line 90:
```suggestion
- Generalised NUTS with slice sampling (Betancourt, M. (2017)): `NUTS{SliceTS,GeneralisedNoUTurn}(integrator)`
```

README.md
```diff
@@ -79,57 +84,59 @@ where `ϵ` is the step size of leapfrog integration.
...
+- Original NUTS with multinomial sampling: `NUTS{MultinomialTS,ClassicNoUTurn}(integrator)`
```
yebai, line 91:
```suggestion
- Original NUTS with multinomial sampling (Betancourt, M. (2017)): `NUTS{MultinomialTS,ClassicNoUTurn}(integrator)`
```

README.md
```diff
@@ -79,57 +84,59 @@ where `ϵ` is the step size of leapfrog integration.
...
+- Generalised NUTS with multinomial sampling: `NUTS{MultinomialTS,GeneralisedNoUTurn}(integrator)`
```
yebai, line 92:
```suggestion
- Generalised NUTS with multinomial sampling (Betancourt, M. (2017)): `NUTS{MultinomialTS,GeneralisedNoUTurn}(integrator)`
```

README.md
```diff
@@ -79,57 +84,59 @@ where `ϵ` is the step size of leapfrog integration.
...
 
 ### Adaptor (`adaptor`)
 
-- Preconditioning on metric space `metric`: `pc = Preconditioner(metric)`
-- Nesterov's dual averaging with target acceptance rate `δ` on integrator `int`: `da = NesterovDualAveraging(δ, int)`
-- Combine the two above *naively*: `NaiveHMCAdaptor(pc, da)`
-- Combine the first two using Stan's windowed adaptation: `StanHMCAdaptor(pc, da)`
+- Adapt the mass matrix of the metric space `metric`: `mma = MassMatrixAdaptor(metric)`
```
yebai, line 96:
```suggestion
- Adapt the mass matrix `metric` of the Hamiltonian dynamics: `mma = MassMatrixAdaptor(metric)`
```

README.md
```diff
@@ -79,57 +84,59 @@ where `ϵ` is the step size of leapfrog integration.
...
+  - This is lowered to `UnitMassMatrix`, `WelfordVar` or `WelfordCov` based on the type of the metric space.
```
yebai, line 97:
```suggestion
  - This is lowered to `UnitMassMatrix`, `WelfordVar` or `WelfordCov` based on the type of the mass matrix `metric`
```

README.md
```diff
@@ -79,57 +84,59 @@ where `ϵ` is the step size of leapfrog integration.
...
+- Adapt the step size of the integrator `integrator`: `ssa = StepSizeAdaptor(δ, integrator)`
```
yebai, line 98:
```suggestion
- Adapt the step size of the leapfrog integrator `integrator`: `ssa = StepSizeAdaptor(δ, integrator)`
```

README.md
```diff
@@ -79,57 +84,59 @@ where `ϵ` is the step size of leapfrog integration.
...
+  - It uses Nesterov's dual averaging with `δ` as the target acceptance rate.
+- Combine the two above *naively*: `NaiveHMCAdaptor(mma, ssa)`
+- Combine the first two using Stan's windowed adaptation: `StanHMCAdaptor(mma, ssa)`
 
 ### Gradients 
-`AdvancedHMC` supports both AD-based (`Zygote`, `Tracker` and `ForwardDiff`) and user-specified gradients. For the latter, simply replace `ForwardDiff` with `ℓπ_grad` in the ` Hamiltonian`  constructor, where the gradient function `π_grad` should return a tuple containing both the likelihood and its gradient. 
+`AdvancedHMC` supports both AD-based (`Zygote`, `Tracker` and `ForwardDiff`) and user-specified gradients. For the latter, simply replace `ForwardDiff` with `ℓπ_grad` in the `Hamiltonian`  constructor, where the gradient function `π_grad` should return a tuple containing both the likelihood and its gradient.
```
yebai, line 104:
```suggestion
`AdvancedHMC` supports both AD-based (`Zygote`, `Tracker` and `ForwardDiff`) and user-specified gradients. In order to use user-specified gradients, please replace `ForwardDiff` with `ℓπ_grad` in the `Hamiltonian`  constructor, where the gradient function `ℓπ_grad` should return a tuple containing both the log-likelihood and its gradient. 
```

README.md
```diff
@@ -79,57 +84,59 @@ where `ϵ` is the step size of leapfrog integration.
...
-    drop_warmup::Bool=false,
+    drop_warmup=false,
     verbose::Bool=true,
-    progress::Bool=false
+    progress::Bool=false,
 )
 ```
 
-Sample `n_samples` samples using the proposal `τ` under Hamiltonian `h`
+Sample `n_samples` samples using the proposal `τ` under the Hamiltonian `h`
```
yebai, line 125:
```suggestion
Draw `n_samples` samples using the proposal `τ` under the Hamiltonian system `h`
```

README.md
```diff
@@ -79,57 +84,59 @@ where `ϵ` is the step size of leapfrog integration.
...
 
-- The randomness is controlled by `rng`. 
-    - If `rng` is not provided, `GLOBAL_RNG` will be used.
+- The randomness is controlled by `rng`.
+  - If `rng` is not provided, `GLOBAL_RNG` will be used.
 - The initial point is given by `θ`.
 - The adaptor is set by `adaptor`, for which the default is no adaptation.
-    - It will perform `n_adapts` steps of adaptation, for which the default is the minimum of `1_000` and 10% of `n_samples`.
+  - It will perform `n_adapts` steps of adaptation, for which the default is the minimum of `1_000` and 10% of `n_samples`.
```
yebai, line 131:
```suggestion
  - It will perform `n_adapts` steps of adaptation, for which the default is `1_000` or 10% of `n_samples`, whichever is lower. 
```

README.md
yebai: @xukai92 For user-specified gradients, should the user define `π_grad`, or ``ℓπ_grad`?

yebai:
@xukai92  I did some minor edits to the README file. There is no changes to code. Also, there is a place where I'm not certain, please see comment above.

README.md
xukai92: It's `ℓπ_grad`. I think we had this typo since long time ago.

PR #190, opened 2020-03-31
[WIP] Improvements for vectorized mode

xukai92:
- `similar` now uses `zeros` instead of `similar` inside to avoid warnings from the numerical check in `PhasePoint`
- `similar(z::PhasePoint{<:Vector}, n::Int)` is added to initialize `PhasePoint{<:Matrix}` where the number of cols is `n`
- `randcat(::PhasePoint{<:Matrix})` is added accordingly


xukai92:
This is work-in-progress. I will post some benchmark nunbers of matrix-phasepoint against vector of phasepoints in a bit.

PR #191, opened 2020-04-09
Replace similar by zeros to avoid warning in PhasePoint

xukai92:
The old code uses `similar` to initialise the container for `PhasePoint`, which would lead warnings from the numeric check in the construction of `PhasePonit`. This PR fixes it.

### DIFF
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -23,7 +23,8 @@ struct DualValue{V<:AbstractScalarOrVec{<:AbstractFloat}, G<:AbstractVecOrMat{<:
     end
 end
 
-Base.similar(dv::DualValue{<:AbstractVector}) = DualValue(similar(dv.value), similar(dv.gradient))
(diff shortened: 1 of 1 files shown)


src/hamiltonian.jl
```diff
@@ -23,7 +23,8 @@ struct DualValue{V<:AbstractScalarOrVec{<:AbstractFloat}, G<:AbstractVecOrMat{<:
     end
 end
 
-Base.similar(dv::DualValue{<:AbstractVector}) = DualValue(similar(dv.value), similar(dv.gradient))
+Base.similar(dv::DualValue{<:AbstractVecOrMat{T}}, args...) where {T<:AbstractFloat} =
```
yebai, line 26: Do we really need the `argus..` argument? It does not to be used anywhere.

xukai92: it's a prepare for supporting `similar(z, 10)`, which will initialise a phase point with an extra dimension of length 10. But I guess I should introduce it later. Let me remove it for now.

Issue #198, opened 2020-05-22
Static trajectory multinomial HMC uses MH correction


bonStats:
In the transition function for static trajectories, it appears a Metropolis-Hastings correction is being applied:
https://github.com/TuringLang/AdvancedHMC.jl/blob/f31dc17ac6611dfcc255ce9b903e8d127e17cd23/src/trajectory.jl#L178
even when the trajectory is a `StaticTrajectory{MultinomialTS}`. 

As far as I can tell from https://arxiv.org/pdf/1701.02434.pdf#subsection.A.2 I think the MH correction isn't required when a multinomial sample of the trajectory is used.

Is there something in my understanding or the code that I'm missing?

xukai92:
Thanks for the issue. I think it's a bug. There shouldn't be any MH correction.

There is actually also another issue in how the trajectory is sampled for MultinomialTS, which I fixed in my local branch. I will create a PR to address both soon.

bonStats:
Fixed in https://github.com/TuringLang/AdvancedHMC.jl/pull/199

PR #199, opened 2020-05-23
Fix bugs in multinomial sampling

xukai92:
There are a few bugs in the multinomial trajectory sampling

1. There is an extra MH correction step (https://github.com/TuringLang/AdvancedHMC.jl/issues/198) that should be removed.
2. The way trajectory got sampled is wrong. The correct way is randomly expanding on both directions.
3. `randcat` for matrix has a typo that ended up shifting the samples by 1. This only influences the vectorized mode.

### DIFF
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -87,7 +87,13 @@ function step(
         else
             res = z
         end
-        !isfinite(z) && break
+        if !isfinite(z)
+            # Remove undef
+            if FullTraj
+                res = res[isassigned.(Ref(res), 1:n_steps)]
+            end
+            break
+        end
     end
     return res
 end
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -278,7 +278,7 @@ end
 # zs is in the form of Vector{PhasePoint{Matrix}} and has shape [n_steps][dim, n_chains]
 function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMatrix)
     z = similar(first(zs))
-    P = exp.(unnorm_ℓP .- logsumexp(unnorm_ℓP; dims=2)) # (n_chians, n_steps)
+    P = exp.(unnorm_ℓP .- logsumexp(unnorm_ℓP; dims=2)) # (n_chains, n_steps)
     is = randcat(rng, P')
     foreach(enumerate(is)) do (i_chain, i_step)
         zi = zs[i_step]
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -213,9 +213,7 @@ function transition(
 )
     H0 = energy(z)
     integrator = jitter(rng, τ.integrator)
-    z′ = samplecand(rng, τ, h, z)
-    # Are we going to accept the `z′` via MH criteria?
-    is_accept, α = mh_accept_ratio(rng, energy(z), energy(z′))
+    z′, is_accept, α = samplecand(rng, τ, h, z)
     # Do the actual accept / reject
     z = accept_phasepoint!(z, z′, is_accept)    # NOTE: this function changes `z′` in place in matrix-parallel mode
     # Reverse momentum variable to preserve reversibility
@@ -263,17 +261,26 @@ end
 
 ### Use end-point from trajectory as proposal 
 
-samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z) = step(τ.integrator, h, z, τ.n_steps)
+function samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z)
+    z′ = step(τ.integrator, h, z, τ.n_steps)
+    # Are we going to accept the `z′` via MH criteria?
+    is_accept, α = mh_accept_ratio(rng, energy(z), energy(z′))
+    return z′, is_accept, α
+end
 
 ### Multinomial sampling from trajectory
 
-randcat(rng::AbstractRNG, zs::AbstractVector{<:PhasePoint}, unnorm_ℓp::AbstractVector) = 
-    zs[randcat_logp(rng, unnorm_ℓp)]
+function randcat(rng::AbstractRNG, zs::AbstractVector{<:PhasePoint}, unnorm_ℓp::AbstractVector)
+    p = exp.(unnorm_ℓp .- logsumexp(unnorm_ℓp))
+    i = randcat(rng, p)
+    return zs[i]
+end
 
 # zs is in the form of Vector{PhasePoint{Matrix}} and has shape [n_steps][dim, n_chains]
 function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMatrix)
     z = similar(first(zs))
-    is = randcat_logp(rng, unnorm_ℓP)
+    P = exp.(unnorm_ℓP .- logsumexp(unnorm_ℓP; dims=2))
+    is = randcat(rng, P)
     foreach(enumerate(is)) do (i_chain, i_step)
         zi = zs[i_step]
         z.θ[:,i_chain] = zi.θ[:,i_chain]
@@ -298,7 +305,12 @@ function samplecand(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
         ℓws = hcat(ℓws...)
     end
     unnorm_ℓprob = ℓws
-    return randcat(rng, zs, unnorm_ℓprob)
+    z′ = randcat(rng, zs, unnorm_ℓprob)
+    Hs = -ℓws
+    # Computing adaptation statistics for dual averaging as done in NUTS
+    ΔH = Hs .- energy(z)
+    α = mean(exp.(min.(0, -ΔH)))
+    return z′, true, α
 end
 
 abstract type DynamicTrajectory{I<:AbstractIntegrator} <: AbstractTrajectory{I} end
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -263,7 +263,6 @@ end
 
 function sample_phasepoint(rng, τ::StaticTrajectory{EndPointTS}, h, z)
     z′ = step(τ.integrator, h, z, τ.n_steps)
-    # Are we going to accept the `z′` via MH criteria?
     is_accept, α = mh_accept_ratio(rng, energy(z), energy(z′))
     return z′, is_accept, α
 end
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -294,9 +294,9 @@ end
 
 function sample_phasepoint(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
     n_steps = abs(τ.n_steps)
-    # FIXME: deal with multi-chain generically.
-    #        Currently the direction of multiple chains are always coupled
-    n_steps_fwd = rand_sync(rng, 0:n_steps) 
+    # TODO: Deal with vectorized-mode generically.
+    #       Currently the direction of multiple chains are always coupled
+    n_steps_fwd = rand_coupled(rng, 0:n_steps) 
     zs_fwd = step(τ.integrator, h, z, n_steps_fwd; fwd=true, full_trajectory=Val(true))
     n_steps_bwd = n_steps - n_steps_fwd
     zs_bwd = step(τ.integrator, h, z, n_steps_bwd; fwd=false, full_trajectory=Val(true))
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -259,11 +259,7 @@ function accept_phasepoint!(z::T, z′::T, is_accept) where {T<:PhasePoint{<:Abs
     return z′
 end
 
-<<<<<<< HEAD
-### Use end-point from trajectory as proposal 
-=======
 ### Use end-point from the trajectory as a proposal and apply MH correction
->>>>>>> Apply suggestions from Hong
 
 function samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z)
     z′ = step(τ.integrator, h, z, τ.n_steps)
@@ -299,17 +295,10 @@ end
 
 function samplecand(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
     n_steps = abs(τ.n_steps)
-<<<<<<< HEAD
     n_steps_fwd = rand(0:n_steps) # FIXME: deal with multi-chain generically
     zs_fwd = step(τ.integrator, h, z, n_steps_fwd; fwd=true, full_trajectory=Val(true))
     n_steps_bwd = n_steps - n_steps_fwd
     zs_bwd = step(τ.integrator, h, z, n_steps_bwd; fwd=false, full_trajectory=Val(true))
-=======
-    n_steps_fwd = rand(0:n_steps) # FIXME: deal with multi-chain generically
-    zs_fwd = step(τ.integrator, h, z, n_fwd; fwd=true, res=[z for _ in 1:n_fwd])
-    n_steps_bwd = n_steps - n_steps_fwd
-    zs_bwd = step(τ.integrator, h, z, n_bwd; fwd=false, res=[z for _ in 1:n_bwd])
->>>>>>> Apply suggestions from Hong
     zs = vcat(reverse(zs_bwd)..., z, zs_fwd...)
     ℓws = -energy.(zs)
     if eltype(ℓws) <: AbstractVector
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -267,7 +267,8 @@ samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z) = step(τ.integrator, h,
 
 ### Multinomial sampling from trajectory
 
-randcat(rng::AbstractRNG, zs::AbstractVector{<:PhasePoint}, unnorm_ℓp::AbstractVector) = zs[randcat_logp(rng, unnorm_ℓp)]
+randcat(rng::AbstractRNG, zs::AbstractVector{<:PhasePoint}, unnorm_ℓp::AbstractVector) = 
+    zs[randcat_logp(rng, unnorm_ℓp)]
 
 # zs is in the form of Vector{PhasePoint{Matrix}} and has shape [n_steps][dim, n_chains]
 function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMatrix)
@@ -286,7 +287,12 @@ function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMat
 end
 
 function samplecand(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
-    zs = step(τ.integrator, h, z, τ.n_steps; full_trajectory = Val(true))
+    n_steps = abs(τ.n_steps)
+    n_fwd = rand(0:n_steps) # FIXME: deal with multi-chain generically
+    zs_fwd = step(τ.integrator, h, z, n_fwd; fwd=true, full_trajectory=Val(true))
+    n_bwd = n_steps - n_fwd
+    zs_bwd = step(τ.integrator, h, z, n_bwd; fwd=false, full_trajectory=Val(true))
+    zs = vcat(reverse(zs_bwd)..., z, zs_fwd...)
     ℓws = -energy.(zs)
     if eltype(ℓws) <: AbstractVector
         ℓws = hcat(ℓws...)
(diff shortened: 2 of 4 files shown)


src/metric.jl
```diff
@@ -1,6 +1,6 @@
 abstract type AbstractMetric end
 
-_string_M⁻¹(mat::AbstractMatrix, n_chars::Int=32) = _string_M⁻¹(LinearAlgebra.diag(mat), n_chars)
+_string_M⁻¹(mat::AbstractMatrix, n_chars::Int=32) = _string_M⁻¹(diag(mat), n_chars)
```
yebai, line 3: Minor comment: Any reason for reversing this change made in a previous PR?

src/trajectory.jl
```diff
@@ -173,9 +173,7 @@ function transition(
 )
     H0 = energy(z)
     integrator = jitter(rng, τ.integrator)
-    z′ = samplecand(rng, τ, h, z)
-    # Are we going to accept the `z′` via MH criteria?
-    is_accept, α = mh_accept_ratio(rng, energy(z), energy(z′))
+    z′, is_accept, α = samplecand(rng, τ, h, z)
```
yebai, line 176: Consider renaming `samplecand` to `sample_phasepoint`  for clarity.

src/trajectory.jl
```diff
@@ -223,16 +221,26 @@ end
 
 ### Use end-point from trajecory as proposal
```
yebai, line 222:
```suggestion
### Use end-point from the trajectory as a proposal and apply MH correction
```

src/trajectory.jl
```diff
@@ -223,16 +221,26 @@ end
 
 ### Use end-point from trajecory as proposal 
 
-samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z) = step(τ.integrator, h, z, τ.n_steps)
+function samplecand(rng, τ::StaticTrajectory{EndPointTS}, h, z)
+    z′ = step(τ.integrator, h, z, τ.n_steps)
+    # Are we going to accept the `z′` via MH criteria?
```
yebai, line 226: Consider removing this obsolete comment line.

src/trajectory.jl
```diff
@@ -246,13 +254,29 @@ function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMat
 end
 
 function samplecand(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
-    zs = step(τ.integrator, h, z, τ.n_steps; res=[z for _ in 1:abs(τ.n_steps)])
+    n_steps = abs(τ.n_steps)
+    n_fwd = rand(0:n_steps) # FIXME: deal with multi-chain generically
```
yebai, line 258:
```suggestion
    n_steps_fwd = rand(0:n_steps) # FIXME: deal with multi-chain generically
```

src/trajectory.jl
```diff
@@ -246,13 +254,29 @@ function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMat
...
+    zs_fwd = step(τ.integrator, h, z, n_fwd; fwd=true, res=[z for _ in 1:n_fwd])
+    n_bwd = n_steps - n_fwd
```
yebai, line 260:
```suggestion
    n_steps_bwd = n_steps - n_steps_fwd
```

src/trajectory.jl
```diff
@@ -246,13 +254,29 @@ function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMat
...
+    zs_bwd = step(τ.integrator, h, z, n_bwd; fwd=false, res=[z for _ in 1:n_bwd])
+    zs = vcat(reverse(zs_bwd)..., z, zs_fwd...)
     ℓws = -energy.(zs)
     if eltype(ℓws) <: AbstractVector
-        ℓws = hcat(ℓws...)
+        ℓws = cat(ℓws...; dims=2)
     end
     unnorm_ℓprob = ℓws
```
yebai, line 267: Consider removing unnecessary intermediate variable `ℓws` and use `unnorm_ℓprob` (or unnorm_ℓp for consistency) in all places.

src/trajectory.jl
```diff
@@ -246,13 +254,29 @@ function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMat
...
-    return randcat(rng, zs, unnorm_ℓprob)
+    z′ = randcat(rng, zs, unnorm_ℓprob)
+    # Computing adaptation statistics for dual averaging as done in NUTS
+    Hs = -ℓws
+    ΔH = Hs .- energy(z)
+    α = exp.(min.(0, -ΔH))  # this is a matrix for vectorized mode and a vector otherwise
+    α =
```
yebai, line 273: Consider `α = typeof(α) <: AbstractVector ? mean(α) : vec(mean(α; dims=2))` for clarity.

src/utilities.jl
```diff
@@ -31,17 +28,12 @@ function randcat(rng::AbstractRNG, p::AbstractVector{T}) where {T}
...
-) = randcat(rng, exp.(unnorm_ℓP .- logsumexp(unnorm_ℓP; dims=2)))
-
 function randcat(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}}, 
     P::AbstractMatrix{T}
 ) where {T}
     u = rand(rng, T, size(P, 1))
     C = cumsum(P; dims=2)
-    is = convert.(Int, vec(sum(C .< u; dims=2)))
+    is = convert.(Int, vec(sum(C .< u; dims=2))) .+ 1
```
yebai, line 37: Consider adding a comment explaining why adding 1 is needed here.

test/sampler-vec.jl
```diff
@@ -30,6 +30,9 @@ include("common.jl")
         n_chains = n_chains_list[i_test]
         metric = metricT((D, n_chains))
         h = Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
+        @test show(metric) == nothing
```
yebai, line 33: Consider rebasing with the master branch.

src/trajectory.jl
torfjelde: Or at least putting the if-statement on the same line as `=`, i.e. `α = if typeof(α) <: AbstractVector ...` (subjective opinion)

src/trajectory.jl
```diff
@@ -223,16 +221,26 @@ end
...
+    # Are we going to accept the `z′` via MH criteria?
+    is_accept, α = mh_accept_ratio(rng, energy(z), energy(z′))
+    return z′, is_accept, α
+end
 
 ### Multinomial sampling from trajecory
 
-randcat(rng::AbstractRNG, zs::AbstractVector{<:PhasePoint}, unnorm_ℓp::AbstractVector) = zs[randcat_logp(rng, unnorm_ℓp)]
+function randcat(rng::AbstractRNG, zs::AbstractVector{<:PhasePoint}, unnorm_ℓp::AbstractVector)
+    p = exp.(unnorm_ℓp .- logsumexp(unnorm_ℓp))
```
torfjelde, line 234: Maybe separate this out into it's own function, say, `unnorm_logprob_to_prob`, and call `randcat(rng, zs, unnorm_logprob_to_prob(unnorm_ℓp))` instead? Basically like `randcat_logp` but separated into two functions.
Right now there are some impls of `randcat` that expects unnormalized log-probs while some expect probs :confused:

src/metric.jl
xukai92: This one is not reverting. We previsouly decided to expliclty using functions from `LinearAlgebra`, e.g. `diag` so the variable `LinearAlgebra` is not defined now. This is a bug which was not catched. I'm happy to move it to seperate PR if you think this make the history clear.

src/trajectory.jl
xukai92: I think what you proposed is nicer than the one I reverted, and I see what's your concern about `randcat` sometimes expecte unormed. log. prob. but sometimes expect normed. prob. 

The reason I reverted the previous `randcat_logp` is that this function is only called once in thie `randcat` and not reused at all. This will be the same issue for `unnorm_logprob_to_prob`, and I would prefer we don't make extract this into a seperate function until this is really used more than one time.

Maybe a better way to handle all of this here is actually renaming this `randcat` to `randcat_weight`. Then there will be no ambiguity: `randcat` alwasy expected normalised probability, and there will not function that is only used once.

src/trajectory.jl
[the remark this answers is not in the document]
xukai92: Yes it's correct.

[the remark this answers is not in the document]
xukai92: Yes, right now in vectorized mode, the number of forward and backward are always coupled.

src/trajectory.jl
xukai92: Sure. I will take Hong's suggestion. What I currently have here is a very bad style!

src/trajectory.jl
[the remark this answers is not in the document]
xukai92: One thing to correct / be explict here: the trajectory has `n_steps + 1` *points* while has a length of `n_steps * step_size`; `zs` is an array with length of `n_steps + 1`.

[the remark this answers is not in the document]
torfjelde: By "length" I meant the _number of steps_ :+1: And (I'm being a bit pedantic here) I feel like using "length" about `n_steps * step_size` is slightly incorrect, since it's _not_ the length of the trajectory in, say, Euclidean distance as you'd normally interpret it, but rather it's the "interval of time" that you integrate, right?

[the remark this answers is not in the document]
yebai: Yes, that's integration time if we want to be pedantic.

src/trajectory.jl
yebai: `randcat_weight` is mysterious I think; consider renaming `randcat` to `rand_categorical` if possible. I hesitated a bit to raise this issue in my previous review because this is not related to this PR, and can be fixed in a separate effort.

src/trajectory.jl
```diff
@@ -246,13 +254,29 @@ function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMat
 end
 
 function samplecand(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
-    zs = step(τ.integrator, h, z, τ.n_steps; res=[z for _ in 1:abs(τ.n_steps)])
+    n_steps = abs(τ.n_steps)
+    n_steps_fwd = rand(0:n_steps) # FIXME: deal with multi-chain generically
+    zs_fwd = step(τ.integrator, h, z, n_fwd; fwd=true, res=[z for _ in 1:n_fwd])
```
yebai, line 259: we need to replace all references of `n_fwd` and `n_bwd` with `n_steps_fwd` and `n_steps_fwd`, respectively.

yebai:
Ready to merge once the tests pass.

xukai92:
> Ready to merge once the tests pass.

I haven't addressed all comments yet. Will ping you when it's ready.

xukai92:
Tests are passed now.

src/integrator.jl
```diff
@@ -87,7 +87,13 @@ function step(
         else
             res = z
         end
-        !isfinite(z) && break
+        if !isfinite(z)
+            # Remove undef
+            if FullTraj
+                res = res[isassigned.(Ref(res), 1:n_steps)]
```
xukai92, line 93: @yebai @torfjelde I encountered an issue today. Basically the change in https://github.com/TuringLang/AdvancedHMC.jl/pull/200 reveals another uncatched bug before: the ummatrued chain (due to `break` caused by numerical issues) now could contain `undef`, which should be correctly "removed".

xukai92: Before we were passing a vector of `z` (initial phase point), which silently increase the probabilities of using the initial phase point if there is a numerical issue.

torfjelde: Oooh, aight; nice catch!

xukai92:
@yebai Shall we merge this?

src/integrator.jl
yebai: Is `Ref` necessary here, why? 

Also, is the broadcasting for vectorised chains?

src/utilities.jl
```diff
@@ -16,10 +16,22 @@ function Base.randn(rng::AbstractVector{<:AbstractRNG}, T, dim::Int, n_chains::I
...
+# Helper functions to sync RNGs that ensures coupling works properly and reproducible
 
-randcat_logp(rng::AbstractRNG, unnorm_ℓp::AbstractVector) =
-    randcat(rng, exp.(unnorm_ℓp .- logsumexp(unnorm_ℓp)))
+rand_sync(rng::AbstractRNG, args...) = rand(rng, args...)
+
+# This function is needed to use only one RNG while a vector of RNGs are provided
+# in a way that RNGs are still synchronised. This is important if we want to use 
+# same RNGs to couple multiple chains.
+function rand_sync(rngs::AbstractVector{<:AbstractRNG}, args...)
```
yebai, line 26: Consider renaming `rand` into `rand_coupled` for clarity?  
```suggestion
function rand_coupled(rngs::AbstractVector{<:AbstractRNG}, args...)
```

src/utilities.jl
```diff
@@ -16,10 +16,22 @@ function Base.randn(rng::AbstractVector{<:AbstractRNG}, T, dim::Int, n_chains::I
     return cat(randn.(rng, T, dim)...; dims=2)
 end
 
-# Sample from Categorical distributions
+# Helper functions to sync RNGs that ensures coupling works properly and reproducible
 
-randcat_logp(rng::AbstractRNG, unnorm_ℓp::AbstractVector) =
-    randcat(rng, exp.(unnorm_ℓp .- logsumexp(unnorm_ℓp)))
+rand_sync(rng::AbstractRNG, args...) = rand(rng, args...)
```
yebai, line 21:
```suggestion
rand_coupled(rng::AbstractRNG, args...) = rand(rng, args...)
```

src/utilities.jl
```diff
@@ -16,10 +16,22 @@ function Base.randn(rng::AbstractVector{<:AbstractRNG}, T, dim::Int, n_chains::I
...
+
+# This function is needed to use only one RNG while a vector of RNGs are provided
+# in a way that RNGs are still synchronised. This is important if we want to use 
+# same RNGs to couple multiple chains.
+function rand_sync(rngs::AbstractVector{<:AbstractRNG}, args...)
+    # Dummpy calles to sync RNGs
+    foreach(rngs[2:end]) do rng
```
yebai, line 28: This is potentially a waste of computation. Consider to simply "duplicate" the first RNG so that all RNGS are identical. For example:

```
rngs[2:end] = [copy(rngs[1]) for i=1:(length(rngs)-1)]) 
```

src/utilities.jl
```diff
@@ -31,17 +43,43 @@ function randcat(rng::AbstractRNG, p::AbstractVector{T}) where {T}
...
+```
+Then `C .< u'` is
+```
+[
+    0 1
+    0 0
+    0 0
+]
+```
+and thus should have `convert.(Int, vec(sum(C .< u'; dims=1))) .+ 1` equals `[1, 2]`.
```
yebai, line 75:
```suggestion
thus `convert.(Int, vec(sum(C .< u'; dims=1))) .+ 1` equals `[1, 2]`.
```

src/utilities.jl
```diff
@@ -31,17 +43,43 @@ function randcat(rng::AbstractRNG, p::AbstractVector{T}) where {T}
...
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}}, 
     P::AbstractMatrix{T}
 ) where {T}
-    u = rand(rng, T, size(P, 1))
-    C = cumsum(P; dims=2)
-    is = convert.(Int, vec(sum(C .< u; dims=2)))
-    return max.(is, 1)
+    u = rand(rng, T, size(P, 2))
+    C = cumsum(P; dims=1)
+    is = convert.(Int, vec(sum(C .< u'; dims=1))) .+ 1
```
yebai, line 83:
```suggestion
    indices = convert.(Int, vec(sum(C .< u'; dims=1))) .+ 1
```

src/utilities.jl
```diff
@@ -31,17 +43,43 @@ function randcat(rng::AbstractRNG, p::AbstractVector{T}) where {T}
...
+    return max.(is, 1)  # prevent numerical issue for Float32
```
yebai, line 84:
```suggestion
    return max.(indices, 1)  # prevent numerical issue for Float32
```

src/utilities.jl
```diff
@@ -16,10 +16,22 @@ function Base.randn(rng::AbstractVector{<:AbstractRNG}, T, dim::Int, n_chains::I
     return cat(randn.(rng, T, dim)...; dims=2)
 end
 
-# Sample from Categorical distributions
+# Helper functions to sync RNGs that ensures coupling works properly and reproducible
```
yebai, line 19:
```suggestion
""" 
`rand_coupled` produced coupled randomness given a vector of RNGs.  For example, 
when a vector of RNGs is provided, `rand_coupled` returns the same value for all RNGs
and keep all RNGs synchronised.  This is important if we want to couple multiple Markov chains.
"""
```

src/utilities.jl
```diff
@@ -16,10 +16,22 @@ function Base.randn(rng::AbstractVector{<:AbstractRNG}, T, dim::Int, n_chains::I
...
 
-randcat_logp(rng::AbstractRNG, unnorm_ℓp::AbstractVector) =
-    randcat(rng, exp.(unnorm_ℓp .- logsumexp(unnorm_ℓp)))
+rand_sync(rng::AbstractRNG, args...) = rand(rng, args...)
+
+# This function is needed to use only one RNG while a vector of RNGs are provided
```
yebai, line 23: Consider merging these comments into docstring. See above.

src/integrator.jl
xukai92: > Is Ref necessary here, why?

Yes, the signature is `isassigned(::Vector, ::Int)`.

> Also, is the broadcasting for vectorised chains?

What do you mean?

xukai92: Just to make the first reply easier to understand

```julia
julia> x = Vector{Any}(undef, 3)
3-element Array{Any,1}:
 #undef
 #undef
 #undef

julia> x[1] = 1
1

julia> x[2] = 2
2

julia> isassigned(x, 1)
true

julia> isassigned(x, 2)
true

julia> isassigned(x, 3)
false

julia> isassigned.(x, 1:3)
ERROR: MethodError: no method matching isassigned(::Int64, ::Int64)
Closest candidates are:
  isassigned(::Core.SimpleVector, ::Int64) at essentials.jl:651
  isassigned(::Array, ::Int64...) at array.jl:226
  isassigned(::BitArray, ::Int64) at bitarray.jl:107

julia> isassigned.(Ref(x), 1:3)
3-element BitArray{1}:
 1
 1
 0
```

src/utilities.jl
xukai92: I will do this on my local and push because there are other places we need to make this change.

src/utilities.jl
xukai92: What we have is a slightly different issue. Let me explain if from the begining: 

Assume we have two chains with `rng1` and `rng2`. When we are using coupled HMC, what is pass is `rng = [rng1, rng2]` where `rng1 == rng2`. Is this all good for *two* rand call for *two* chains but not to do *one* rand call for *two* chains.

To make this intuitive, consider below are computations for two chains

```
chain 1: ... -> rand(rng1) -> ... -> rand(rng1)
chain 2: ... -> rand(rng2) -> ... -> rand(rng2)
```
is a situation works fine.

But 
```
chain 1: ... -> rand(rng1) -\                  /-> rand(rng1)
                                              |> rand(???) -
chain 2: ... -> rand(rng2) -/                  \-> rand(rng2)
```
is a situation we were not --- we simply using `GLOBAL_RNG` for `???` before, which is correct but doesn't make the computation reproducible. What I'm trying to fix is this reproducibiltiy.

Come back to what you propose, yes, it will fix the issue *if* we do want to use the *identical* rngs for all chains. However, if we did so, our vectorized implementation will enforce all chains to be coupled there, which is not what we want: what we are is, if the passed-in rngs are idential, make sure the rng sequences are same for two chans; if the passed-in rngs are not idential, we don't want to introduce any coupling.

Ideally, we need a coupled chain level rng to do so, i.e. let `??? = rng_coupled` or something. But I don't think this is easy to be done with the current abstraction. Also I think this computation waste is negectble.

yebai: `copy` an RNG will produce an independent copy of that RNG, so copying leads to identical but de-coupled random number generators. That said, I'm happy to merge this as-is for now.

src/integrator.jl
yebai: thanks for the clarification!

PR #200, opened 2020-05-24
Small changes

torfjelde:
Currently going through the code and making some suggestions for small changes as I do.

It's mostly just docstrings, with the exception of an implementation change to `step` where I've replaced what would usually look like `step(...; res = [z for _ = 1:n_steps])` with `step(...; full_trajectory = Val(true))`. 

## Up for discussion
- Maybe just make it `step(...; full_trajectory::Bool = false)` instead? Julia's constant propagation should compile it away in the cases that we use it anyways since it's always hard-coded into the method depending on what it needs.

### DIFF
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -7,6 +7,17 @@
 # We might want to simplify it to `Tuple{Int,Bool}` when we figured out
 # why the it behaves unexpected on Windos 32.
 
+"""
+$(TYPEDEF)
+
+Represents an integrator used to simulate the Hamiltonian system.
+
+# Implementation
+A `AbstractIntegrator` is expected to have the following implementations:
+- `stat`(@ref)
+- `nom_step_size`(@ref)
+- `step_size`(@ref)
+"""
 abstract type AbstractIntegrator end
 
 stat(::AbstractIntegrator) = NamedTuple()
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -59,7 +59,7 @@ function step(
     ϵ = ϵ'
 
     res = if FullTraj
-        Vector{typeof(z)}(undef, n_steps)
+        Vector{P}(undef, n_steps)
     else
         z
     end
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -81,7 +81,17 @@ function step(
     return res
 end
 
+"""
+$(TYPEDEF)
+
+Leapfrog integrator with fixed step size `ϵ`.
+
+# Fields
+
+$(TYPEDFIELDS)
+"""
 struct Leapfrog{T<:AbstractScalarOrVec{<:AbstractFloat}} <: AbstractLeapfrog{T}
+    "Step size."
     ϵ       ::  T
 end
 Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round.(l.ϵ; sigdigits=3)))")
@@ -91,17 +101,29 @@ Base.show(io::IO, l::Leapfrog) = print(io, "Leapfrog(ϵ=$(round.(l.ϵ; sigdigits
 """
 $(TYPEDEF)
 
-Randomly "jitters" the step size at the beginning of each trajectory.
+Leapfrog integrator with randomly "jittered" step size `ϵ` for every trajectory.
 
 # Fields
 
 $(TYPEDFIELDS)
 
 # Description
-
-This might help alleviate issues related to poor interactions with a fixed step size, such as:
-- In regions with high "curvature" the current choice of step size might mean over-shoot leading to almost all steps being rejected. Randomly sampling the step size at the beginning of the trajectories can therefore increase the probability of escaping such high-curvature regions.
-- Exact periodicity of the simulated trajectories might occur, i.e. you might be so unlucky as to simulate the trajectory forwards in time `L ϵ` and ending up at the same point (which results in non-ergodicity; see Section 3.2 in [1]). If momentum is refreshed before each trajectory, then this should not happen *exactly* but it can still be an issue in practice. Randomly choosing the step-size `ϵ` might help alleviate such problems.
+This is the same as `LeapFrog`(@ref) but with a "jittered" step size. This means 
+that at the beginning of each trajectory we sample a step size `ϵ` by adding or 
+subtracting from the nominal/base step size `ϵ0` some random proportion of `ϵ0`, 
+with the proportion specified by `jitter`, i.e. `ϵ = ϵ0 - jitter * ϵ0 * rand()`.
+p
+Jittering might help alleviate issues related to poor interactions with a fixed step size:
+- In regions with high "curvature" the current choice of step size might mean over-shoot 
+  leading to almost all steps being rejected. Randomly sampling the step size at the 
+  beginning of the trajectories can therefore increase the probability of escaping such
+  high-curvature regions.
+- Exact periodicity of the simulated trajectories might occur, i.e. you might be so
+  unlucky as to simulate the trajectory forwards in time `L ϵ` and ending up at the
+  same point (which results in non-ergodicity; see Section 3.2 in [1]). If momentum
+  is refreshed before each trajectory, then this should not happen *exactly* but it
+  can still be an issue in practice. Randomly choosing the step-size `ϵ` might help
+  alleviate such problems.
 
 # References
 1. Neal, R. M. (2011). MCMC using Hamiltonian dynamics. Handbook of Markov chain Monte Carlo, 2(11), 2. ([arXiv](https://arxiv.org/pdf/1206.1901))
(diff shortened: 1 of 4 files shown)


src/integrator.jl
```diff
@@ -40,13 +51,19 @@ function step(
...
-) where {T<:AbstractScalarOrVec{<:AbstractFloat}, P<:PhasePoint}
+    full_trajectory::Val{FullTraj} = Val(false)
+) where {T<:AbstractScalarOrVec{<:AbstractFloat}, P<:PhasePoint, FullTraj}
     n_steps = abs(n_steps)  # to support `n_steps < 0` cases
 
     ϵ = fwd ? step_size(lf) : -step_size(lf)
     ϵ = ϵ'
 
+    res = if FullTraj
+        Vector{typeof(z)}(undef, n_steps)
```
xukai92, line 62:
```suggestion
        Vector{P}(undef, n_steps)
```

src/trajectory.jl
```diff
@@ -246,7 +286,7 @@ function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMat
 end
 
 function samplecand(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
-    zs = step(τ.integrator, h, z, τ.n_steps; res=[z for _ in 1:abs(τ.n_steps)])
+    zs = step(τ.integrator, h, z, τ.n_steps; full_trajectory = Val(true))
```
xukai92, line 289: > Maybe just make it step(...; full_trajectory::Bool = false) instead? Julia's constant propagation should compile it away in the cases that we use it anyways since it's always hard-coded into the method depending on what it needs.

If this true I'd prefer `full_trajectory=true`.

torfjelde: Hmm, yeah it's a bit unclear to me whether or not it works.

For example, if I call `samplecand` I get the following:
```julia
julia> @code_warntype AHMC.samplecand(rng, proposal, hamiltonian, z₀)
Variables
  #self#::Core.Compiler.Const(AdvancedHMC.samplecand, false)
  rng::Core.Compiler.Const(Random._GLOBAL_RNG(), false)
  τ::StaticTrajectory{MultinomialTS,Leapfrog{Float64}}
  h::Hamiltonian{DiagEuclideanMetric{Float64,Array{Float64,1}},typeof(ℓπ),AdvancedHMC.var"#∂ℓπ∂θ#43"{typeof(ℓπ)}}
  z::AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}}
  zs::Union{AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}}, Array{AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}},1}}
  ℓws::Array{Float64,1}
  unnorm_ℓprob::Array{Float64,1}

Body::AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}}
1 ─       Core.NewvarNode(:(unnorm_ℓprob))
│   %2  = Base.getproperty(τ, :integrator)::Leapfrog{Float64}
│   %3  = Base.getproperty(τ, :n_steps)::Int64
│         (zs = AdvancedHMC.step(%2, h, z, %3, true))
│   %5  = Base.broadcasted(AdvancedHMC.energy, zs)::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1},Nothing,typeof(AdvancedHMC.energy),Tuple{Array{AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}},1}}}
│   %6  = Base.materialize(%5)::Array{Float64,1}
│         (ℓws = -%6)
│   %8  = AdvancedHMC.eltype(ℓws)::Core.Compiler.Const(Float64, false)
│   %9  = (%8 <: AdvancedHMC.AbstractVector)::Core.Compiler.Const(false, false)
└──       goto #3 if not %9
2 ─       Core.Compiler.Const(:(ℓws = Core._apply_iterate(Base.iterate, AdvancedHMC.hcat, ℓws)), false)
3 ┄       (unnorm_ℓprob = ℓws)
│   %13 = AdvancedHMC.randcat(rng, zs, unnorm_ℓprob)::AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}}
└──       return %13
```
So there's a `Union` inferred for `zs` which means that the type is not inferred, i.e. constant-propagation isn't working, despite the fact that we can see that `AdvancedHMC.step` is called with `true` as the last argument (I just made it an argument rather than kwarg because I think kwargs aren't "compiled" away as aggressively).

It also doesn't seem to work if I wrap the `step` call:
```julia
julia> a(integrator, hamiltonian, z₀) = AHMC.step(integrator, hamiltonian, z₀, 2, true)
a (generic function with 2 methods)

julia> @code_warntype a(integrator, hamiltonian, z₀)
Variables
  #self#::Core.Compiler.Const(a, false)
  integrator::Leapfrog{Float64}
  hamiltonian::Hamiltonian{DiagEuclideanMetric{Float64,Array{Float64,1}},typeof(ℓπ),AdvancedHMC.var"#∂ℓπ∂θ#43"{typeof(ℓπ)}}
  z₀::AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}}

Body::Union{AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}}, Array{AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}},1}}
1 ─ %1 = AdvancedHMC.step::Core.Compiler.Const(AdvancedHMC.step, false)
│   %2 = (%1)(integrator, hamiltonian, z₀, 2, true)::Union{AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}}, Array{AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}},1}}
└──      return %2
```

However, it _does_ work if I wrap `step` using `const` variables:
```julia
julia> a() = AHMC.step(integrator_const, hamiltonian_const, z_const, 2, true)
a (generic function with 1 method)

julia> @code_warntype a()
Variables
  #self#::Core.Compiler.Const(a, false)

Body::Array{AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}},1}
1 ─ %1 = AdvancedHMC.step::Core.Compiler.Const(AdvancedHMC.step, false)
│   %2 = (%1)(Main.integrator_const, Main.hamiltonian_const, Main.z_const, 2, true)::Array{AdvancedHMC.PhasePoint{Array{Float64,1},AdvancedHMC.DualValue{Float64,Array{Float64,1}}},1}
└──      return %2
```

Therefore I hereby summon the two gods of Julia optimization, oh @mohamed82008 and @devmotion , please provide us peasants with thoust holy guidance!

torfjelde: Alternatively, we just leave it as `Val` for now:)

devmotion: Maybe you observe https://discourse.julialang.org/t/how-does-constant-propagation-work/22735/4?u=devmotion here.

torfjelde: Haha, I just edited the comment to add that link too:) 

Doesn't make much wiser as to what I should do though. Do I leave as is or do you swap to `Bool` and cross my fingers + make it so that I can't inspect the type-stability of calls involving `step`? :confused: 

Feel like leaving it with `Val` is okay for now

yebai: Looks good to me - we can improve this in future PR if necessary.

src/trajectory.jl
```diff
@@ -261,8 +301,22 @@ abstract type DynamicTrajectory{I<:AbstractIntegrator} <: AbstractTrajectory{I}
...
+Standard HMC implementation with fixed total trajectory length.
+
+# Fields
+
+$(TYPEDFIELDS)
+
+# References
+1. Neal, R. M. (2011). MCMC using Hamiltonian dynamics. Handbook of Markov chain Monte Carlo, 2(11), 2. ([arXiv](https://arxiv.org/pdf/1206.1901)) 
+"""
 struct HMCDA{S<:AbstractTrajectorySampler,I<:AbstractIntegrator} <: DynamicTrajectory{I}
```
yebai, line 316: As a side note, HMCDA should be a subtype of static trajectory in fact. Again, we can leave this for a future effort.

src/trajectory.jl
mohamed82008: I don't think constant propagation works with kwargs, not sure though.

PR #201, opened 2020-05-25
Simplify check for finite values

devmotion:


devmotion:
Noticed that it doesn't work for vector-valued parameters, of course :stuck_out_tongue:

PR #204, opened 2020-06-01
Use`include` with the correct order via `@require` in `__init__`

devmotion:
Fixes https://github.com/TuringLang/AdvancedHMC.jl/issues/203.


xukai92:
This PR still uses `include` in `__init__` right?

devmotion:
Yeah but in the correct order. If the `include` statement is inside of the `@require` block, then Requires takes care of everything.

devmotion:
See https://github.com/JuliaPackaging/Requires.jl#requiresjl

xukai92:
Got it. I got confused because of the title. I changed it to something more clear. I will merge this if Travis is happy.

devmotion:
:+1: Yes, indeed, the title was a bit misleading. Tests passed locally on Julia 1.0, so hopefully CI tests pass now :crossed_fingers:

devmotion:
It looks quite good (most test failures are gone) but it seems one test fails since the samples are a bit off:
https://travis-ci.com/github/TuringLang/AdvancedHMC.jl/jobs/342167680#L675

devmotion:
Oh I just noticed that the test failure that was left is also present on master.

devmotion:
Now some other stochastic error showed up, even though the seeds are already fixed there.

src/ad.jl
```diff
@@ -0,0 +1,18 @@
+const ADSUPPORT = (:ForwardDiff, :Zygote)
```
yebai, line 1: Please consider keeping this file under `contrib` folder, i.e. `contrib/ad.jl`. This is mostly to keep the ad and other misc stuff clearly separated from HMC code.  

The folder `contrib` is a bit abused at the moment, we might want to add a `utilities` folder in the future.

devmotion:
https://github.com/devmotion/AdvancedHMC.jl/blob/73be302ffb3fac38900febacc32c34921eed10c0/test/sampler.jl#L88 still errors randomly in one run (Windows x64, Julia nightly) even though a seed is set. I guess that's due to the fact that RNGs are not guaranteed to produce the same numbers in different Julia versions and OS, even if they are initialized with the same seed.

src/ad.jl
devmotion: OK, I changed it :+1:

devmotion:
Tests still fail for StanHMCAdaptor, even with 20000 samples. Is there anything else that could be tuned?

yebai:
I think the default max warmup iteration is 1000, so increasing the total sampling iterations won’t help. You might have to manually increase the number of warmup iterations. 

Also, I’m happy to merge because this is a failure on Julia’s nightly build. We can come back to this if there is a failure on stable releases.

test/sampler.jl
```diff
@@ -9,7 +9,7 @@ include("common.jl")
 θ_init = rand(MersenneTwister(1), D)
 ϵ = 0.2
 n_steps = 10
-n_samples = 12_000
+n_samples = 22_000
 n_adapts = 2_000
```
yebai, line 13:
```suggestion
n_adapts = 4_000
```

test/sampler.jl
```diff
@@ -85,7 +85,7 @@ end
                     # For other adapatation methods that are able to adpat the step size, we use `find_good_stepsize`.
                     τ_used = adaptorsym == :MassMatrixAdaptorOnly ? τ : reconstruct(τ, integrator=reconstruct(lf, ϵ=find_good_stepsize(h, θ_init)))
                     samples, stats = sample(h, τ_used , θ_init, n_samples, adaptor, n_adapts; verbose=false, progress=PROGRESS)
-                    @test mean(samples) ≈ zeros(D) atol=RNDATOL
+                    @test mean(samples[(n_adapts+1):end]) ≈ zeros(D) atol=RNDATOL
```
xukai92, line 88: Maybe use `drop_warmup=true` in the `sample` call above.

test/sampler-vec.jl
```diff
@@ -51,9 +54,12 @@ include("common.jl")
         ]
             τ isa HMCDA && continue
             @test show(adaptor) == nothing
+
+            Random.seed!(100)
             samples, stats = sample(h, τ, θ_init_list[i_test], n_samples, adaptor, n_adapts; verbose=false, progress=false)
```
xukai92, line 59: Maybe `drop_warmup=true` here too.

yebai:
I played with stepsize a bit, but don't have much luck. A quick inspection of the failed assertion suggest the error bound might be too tight: 

```
julia> [-0.142056, -0.624472, 0.0118594, -0.450631, 0.328566] ≈ [0.0, 0.0, 0.0, 0.0, 0.0]
false

julia> a = [-0.142056, -0.624472, 0.0118594, -0.450631, 0.328566]
5-element Array{Float64,1}:
 -0.142056 
 -0.624472 
  0.0118594
 -0.450631 
  0.328566 

julia> using LinearAlgebra

julia> norm(a)
0.8492995618775274
```

test/sampler.jl
yebai: yes, but the effect should be similar to what we currently do.

Issue #94, opened 2019-08-28
Potential improvement for U-Turn detection


yebai:
See https://github.com/stan-dev/stan/pull/2800 and https://discourse.mc-stan.org/t/nuts-misses-u-turns-runs-in-circles-until-max-treedepth/9727/36?u=betanalpha

xukai92:
Related PyMC3 PR: https://github.com/pymc-devs/pymc3/pull/3605

xukai92:
Related issue in DHMC.jl: https://github.com/tpapp/DynamicHMC.jl/issues/115

xukai92:
Closed by https://github.com/TuringLang/AdvancedHMC.jl/pull/207.

PR #207, opened 2020-07-03
Add robust U-turn check

treigerm:
This addresses  #94 and adds the additional U-turn checks. 

The tests still need to be updated but I wanted to get some feedback first whether the general approach here makes sense.
The additional U-turn checks need more information about the tree structure so I had to alter the `BinaryTree` type to keep track of more phase points.

### DIFF
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -602,11 +602,8 @@ end
 
 """
     isterminated(
-        h::Hamiltonian, 
-        t::BinaryTree{<:T},
-        tleft::BinaryTree{<:T}, 
-        tright::BinaryTree{<:T}
-    ) where {T<:StrictGeneralisedNoUTurn}
+        h::Hamiltonian, t::T, tleft::T, tright::T
+    ) where {T<:BinaryTree{<:StrictGeneralisedNoUTurn}}
 
 Detect U turn for two phase points (`zleft` and `zright`) under given Hamiltonian `h`
 using the generalised no-U-turn criterion with additional U-turn checks.
@@ -614,11 +611,8 @@ using the generalised no-U-turn criterion with additional U-turn checks.
 Ref: https://arxiv.org/abs/1701.02434 https://github.com/stan-dev/stan/pull/2800
 """
 function isterminated(
-    h::Hamiltonian, 
-    t::BinaryTree{<:StrictGeneralisedNoUTurn}, 
-    tleft::BinaryTree{<:StrictGeneralisedNoUTurn}, 
-    tright::BinaryTree{<:StrictGeneralisedNoUTurn}
-) 
+    h::Hamiltonian, t::T, tleft::T, tright::T
+) where {T<:BinaryTree{<:StrictGeneralisedNoUTurn}}
     # Classic generalised U-turn check
     t_generalised = BinaryTree(
         t.zleft,
@@ -640,21 +634,15 @@ end
 
 """
     check_left_subtree(
-        h::Hamiltonian, 
-        t::BinaryTree{<:StrictGeneralisedNoUTurn}, 
-        tleft::BinaryTree{<:StrictGeneralisedNoUTurn}, 
-        tright::BinaryTree{<:StrictGeneralisedNoUTurn}
-    )
+        h::Hamiltonian, t::T, tleft::T, tright::T
+    ) where {T<:BinaryTree{<:StrictGeneralisedNoUTurn}}
 
 Do a U-turn check between the leftmost phase point of `t` and the leftmost 
 phase point of `tright`, the right subtree.
 """
 function check_left_subtree(
-    h::Hamiltonian, 
-    t::BinaryTree{<:StrictGeneralisedNoUTurn}, 
-    tleft::BinaryTree{<:StrictGeneralisedNoUTurn}, 
-    tright::BinaryTree{<:StrictGeneralisedNoUTurn}
-)
+    h::Hamiltonian, t::T, tleft::T, tright::T
+) where {T<:BinaryTree{<:StrictGeneralisedNoUTurn}}
     rho = tleft.c.rho + tright.zleft.r
     s = generalised_uturn_criterion(rho, ∂H∂r(h, t.zleft.r), ∂H∂r(h, tright.zleft.r))
     return Termination(s, false)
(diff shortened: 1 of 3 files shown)


xukai92:
Thanks for the PR. 

If I understand correctly, the modified `BinaryTree` struct essentially keeps tracking of two binary trees, one is the left subtree and one is the right subtree, as well as their left- and right-most nodes.
In this case, maybe instead of modifying the struct, we should just keep tracking of two trees within the `build_tree ` and `transition`? For example, currently `build_tree` returns `tree, sampler, termination` and we could modify it to return `tree_left, tree_right, sampler, termination`. Similarly to how trees are passed around in `transition`.

I would prefer this version if it doesn't introduce other complexities because it keeps the `BinaryTree` struct concise.

treigerm:
Thanks for the feedback! I realised that one can directly pass the left and right subtree to `isterminated` which avoids a lot of the changes I introduced. Does this look better now?

The tests also still need to be adapted to account for the new U-turn criterion. Any idea how to best do this? I didn't know how to best adapt the current tests to the new criterion because as far as I can tell the trajectories in the tests don't really have subtrees, right?

xukai92:
> Thanks for the feedback! I realised that one can directly pass the left and right subtree to isterminated which avoids a lot of the changes I introduced. Does this look better now?

Yes! I like the current version more.

> The tests also still need to be adapted to account for the new U-turn criterion. Any idea how to best do this? I didn't know how to best adapt the current tests to the new criterion because as far as I can tell the trajectories in the tests don't really have subtrees, right?

What do you mean by "the trajectories in the tests don't really have subtrees"?

src/trajectory.jl
```diff
@@ -565,19 +575,37 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
+function isterminated(
+    h::Hamiltonian, 
+    t::BinaryTree{<:GeneralisedNoUTurn}, 
+    tleft::BinaryTree{<:GeneralisedNoUTurn}, 
+    tright::BinaryTree{<:GeneralisedNoUTurn}
+)
     rho = t.c.rho
-    s = (dot(rho, ∂H∂r(h, -z0.r)) >= 0) || (dot(-rho, ∂H∂r(h, z1.r)) >= 0)
-    return Termination(s, false)
+    s1 = generalised_uturn_criterion(rho, ∂H∂r(h, t.zleft.r), ∂H∂r(h, t.zright.r))
```
xukai92, line 597: Do we still need this check?

xukai92:
Also, as the new check is a _more strict_ approximation but not _always better_, ideally we should support this via introducing `StrictGeneralisedNoUTurn` and leave the choice of switching between the new and old ones. What we can do here is follow whatever interface required for `StrictGeneralisedNoUTurn` (i.e. passing left and right sub-trees to `isterminated `), then define the `isterminated ` for both `StrictGeneralisedNoUTurn` and `GeneralisedNoUTurn`, just like what it is currently done for `ClassicNoUTurn`.

src/trajectory.jl
```diff
@@ -565,19 +575,37 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
+
+    rho = tleft.c.rho + tright.zleft.r
+    s2 = generalised_uturn_criterion(rho, ∂H∂r(h, t.zleft.r), ∂H∂r(h, tright.zleft.r))
+
+    rho = tleft.zright.r + tright.c.rho
+    s3 = generalised_uturn_criterion(rho, ∂H∂r(h, tleft.zright.r), ∂H∂r(h, t.zright.r))
+    return Termination(s1 || s2 || s3, false)
+end
+
+function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
```
xukai92, line 607: It's good now that we abstract this check out so that doing what I proposed in https://github.com/TuringLang/AdvancedHMC.jl/pull/207#issuecomment-655436469 would be easy as well.

xukai92:
Following https://github.com/TuringLang/AdvancedHMC.jl/pull/207#issuecomment-655436469, we can in fact keep the current `isterminated` and add two functions, one for the new check and one to wrap old ones. Something like below

```julia
function isterminated(h::Hamiltonian, t::BinaryTree{T}, args...) where {T<:Union{ClassicNoUTurn, GeneralisedNoUTurn}}
    return isterminated(h::Hamiltonian, t) # just call the existing functions
end

function isterminated(h::Hamiltonian, t::BinaryTree{T}, tleft, right) where {T<:StrictGeneralisedNoUTurn}
    # Codes for the new check go here
end
```

This makes it clear that `ClassicNoUTurn` and `GeneralisedNoUTurn` only use the combined tree.

Besides, it also makes the existing tests in https://github.com/TuringLang/AdvancedHMC.jl/blob/master/test/trajectory.jl not-broken. Ideally we should also add some test cases for the new check though.

src/trajectory.jl
treigerm: Yes, as I understand the new checks don't catch all U-turns by themselves. Also both Stan (https://github.com/stan-dev/stan/blob/0b92731851947f91d055a61d922b491eb3b14f01/src/stan/mcmc/hmc/nuts/base_nuts.hpp#L178) and PyMC3 (https://github.com/pymc-devs/pymc3/blob/90f48ed22e3c28652903fa27aaaeff59aa29acdf/pymc3/step_methods/hmc/nuts.py#L314) still have this check.

treigerm:
I refactored the code as suggested and I also factored the new checks into their own functions for testing purposes. I added some unit test which make some sanity checks that the new checks properly detect U-turns.

src/trajectory.jl
xukai92: I see. Thanks for clarifying this.

src/trajectory.jl
```diff
@@ -573,13 +595,101 @@ using the generalised no-U-turn criterion.
...
 function isterminated(h::Hamiltonian, t::BinaryTree{<:GeneralisedNoUTurn})
-    # z0 is starting point and z1 is ending point
-    z0, z1 = t.zleft, t.zright
     rho = t.c.rho
-    s = (dot(rho, ∂H∂r(h, -z0.r)) >= 0) || (dot(-rho, ∂H∂r(h, z1.r)) >= 0)
+    s = generalised_uturn_criterion(rho, ∂H∂r(h, t.zleft.r), ∂H∂r(h, t.zright.r))
     return Termination(s, false)
 end
 
+"""
```
xukai92, line 603: The function signature in the comment is different from the actual code. In fact, I think we can simplify all to something like:

```julia
function isterminated(
    h::Hamiltonian, t::T, tleft::T, tright::T
) where {T<:BinaryTree{<:StrictGeneralisedNoUTurn}}
    ...
end
```

which make the code a bit neater.

xukai92: This also applies to `check_left_subtree ` and `check_right_subtree`.

treigerm:
Thanks for the comments! I simplified the function signatures as suggested.

xukai92:
I think that's all we need for the PR. I want to play with this branch locally to reproduce the first 2 plots of the report Michael made here (https://discourse.mc-stan.org/t/nuts-misses-u-turns-runs-in-circles-until-max-treedepth/9727/36) before merging it. If it looks similar I will go merge the PR.

xukai92:
### IID Normal

iid_normal

### Correlated Normal (rho = 0.25)

correlated_normal-rho=0 25

Seems to be consistent to the figures from Michael's report in the sense that the new check only affects the IID normal after dimension reaches around 150. I'm going to merge the PR as the effects seem to be consistent with other implementations.

As a record, [this](https://github.com/TuringLang/AdvancedHMC.jl/files/4907253/report.jl.zip) is my script for the experiments.

Issue #13, opened 2019-03-04
Modular design for refreshing momentum variables


yebai:
We should try to have a modular interface for different momentum refreshment methods (e.g. partial refreshments in [Neal 2011](https://arxiv.org/pdf/1206.1901.pdf))

xukai92:
The current refresh function is here: https://github.com/TuringLang/AdvancedHMC.jl/blob/master/src/hamiltonian.jl#L113.

What's a sensible way to modularise the behaviour? Here are few thoughts / options
1. Parameterize `Hamiltonian`
2. Add another dedicated struct `Refreshment`
    - Then we can implement a parametric `Transition` that consists `Trajectory` and `Refreshment`.

Seems like 2 is btter.

yebai:
I have been thinking about replacing the `metric::M` field of `Hamiltonian` with a more generic design, in a way similar to `ℓπ::Tlogπ`. 

```julia

struct Hamiltonian{M<:AbstractMetric, Tlogκ, T∂logπ∂θ}
    ℓκ::Tlogκ
    ℓπ::Tlogπ
    ∂ℓπ∂θ::T∂logπ∂θ
end
```

Then we can define different `refresh` functions for each `Tlogκ`. I think this design is similar to approach 2 but without an explicit `Refreshment` type.

xukai92:
- What is `Tlogκ` to be more specific? Is it functions or something else?
- I feel the type of kinect energy is orthogonal to how a momentum variable is refreshed.

yebai:
> What is Tlogκ to be more specific? Is it functions or something else?

`Tlogκ` can be a `Distribution` object, e.g. a multivariate `Gaussian` for standard HMC algorithms. 

> I feel the type of kinect energy is orthogonal to how a momentum variable is refreshed.

Yes, but we can make it parametric so that different `refresh/rand` functions can be called.

xukai92:
> Tlogκ can be a Distribution object, e.g. a multivariate Gaussian for standard HMC algorithms.

I guess it's possible. We can even start with replacing the internal of `metric`` via `Distributions` objects as a first step. But I'm not sure how it generalise to say Riemannian  manifold HMC.

> Yes, but we can make it parametric so that different refresh/rand functions can be called.

The question is do we want to make refresh methods as parameter of kinect energy, or we want to make it as parameter of transition, or neither of them but to make the function take another argument and dispatch over structs.

xukai92:
[...]

PR #210, opened 2020-08-01
Struct for kinetic energy and momentum refreshment methods

xukai92:
This PR is to resolve https://github.com/TuringLang/AdvancedHMC.jl/issues/13

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .vscode
+.history
 Manifest.toml
 test/Project.toml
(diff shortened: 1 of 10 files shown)


xukai92:
This new design proposed by Hong seems to work well.
One potential concern is that: 
Since the refreshment method is tied to `Hamiltonian`,
it's not possible to define a mixture kernel of two kernels, one with full refreshment and one with partial refreshment.
This might be a problem in the future.

xukai92:
So maybe we should still consider [my proposal](https://github.com/TuringLang/AdvancedHMC.jl/issues/103#issuecomment-552935345) on putting refreshment into trajectory?

xukai92:
I'm closing this PR due to the issue I pointed in https://github.com/TuringLang/AdvancedHMC.jl/pull/210#issuecomment-667538862. 
We should take the chance to solve #13 when solving #210 instead.

Issue #48, opened 2019-04-04
[RFC]: notations used in this package


yebai:
At the moment, we are using the following notations:

- `logπ` - log target distribution
- `∂logπ∂θ` - gradient
- `prop` - transitioning kernel 
- `θ_init` - initial value for sampling states / parameters

I think this can be improved by

- `∂logπ∂θ` ==> `∇logπ` 
- `prop`  ==> `κ` (kappa) or `τ` (tau) - MH transitioning kernel
- `θ` - parameters, `r` - momentum
- `θ_init` ==> `θ₀` (theta_{0}), `r_init` ==> `r₀` (r_{0})
- `z` := `(θ, r)`   - `(position, momentum)` pair

Any other notations missing?

willtebbutt:
I completely agree with all of @yebai 's proposals. Couple of quick comments:

- I would prefer `τ` to kappa, but don't have a hugely strong feeling either way.
- By `q₀`, do you mean `q₀` or `r₀`?

If the above are resolved, I'm happy :slightly_smiling_face:

yebai:
> By q₀, do you mean q₀ or r₀?

should be ` r₀`.

xukai92:
Nice call! I agree we with these notations! Let's change to them.

xukai92:
@yebai Did we have similar agreement on things like `\epsilon` used in Turing?

yebai:
> @yebai Did we have similar agreement on things like \epsilon used in Turing?

how about using the symbol `ϵ`?

xukai92:
`\epsilon` is `ϵ`

PR #208, opened 2020-07-11
Using MCMCDebugging for Geweke test

xukai92:
This PR aims to resolve https://github.com/TuringLang/AdvancedHMC.jl/issues/87 via MCMCDebugging.jl.

PR #213, opened 2020-08-02
Unifying trajectories

xukai92:
This PR solves #13 and #103 together by explicitly separating `Trajectory` from `HMCKernel`,
without introducing any functionality change.
I also take this chance to unify the notation as discussed in #48,
in which I use `τ` to refer a (numerical) Hamiltonian trajectory and `κ` to refer a MCMC kernel.

As a showcase, below is how to construct NUTS now:

```julia
τ = Trajectory(Leapfrog(1e-3), NoUTurn())
κ = HMCKernel(FullRefreshment(), τ, MultinomialTS)
```

As suggested by https://github.com/TuringLang/AdvancedHMC.jl/issues/103#issue-496507285,
fixed simulation steps or length are unified as termination criteria.
A nice benefit of this is that the parameters regarding trajectory simulation are clear:
e.g. `FixedNSteps(10)`, `NoUTurn(5, 1_000)` or `NoUTurn(max_depth=5)`.

This design mimics the conceptual procedure of doing HMC:

1. Refresh momentum
2. Simulate trajectory
3. Propose a point from the simulated trajectory

although usually the procedure of 2 and 3 are quite coupled.

As a result, the main `transition` function looks like this.
```julia
function transition(rng, h, κ::HMCKernel, z)
    @unpack refreshment, τ, TS = κ
    z = refresh(rng, z, h, refreshment)
    return transition(rng, h, τ, TS, z)
end
```
As a side product of solving #13,
it also unifies how integrators are jittered - all before simulating the trajectory in the same place where momentum variables are refreshed.
With this design, this is the only place where `jitter` and `refresh` is called.


xukai92:
Besides, 
I separated `trajectory.jl` into `kernel.jl` + `trajectory_static.jl` & `trajectory_dynamic.jl`.
In hope to make the review process easier, I reverted this, but it doesn't seem that the changes is well captured by Git.

xukai92:
I hope we can have https://github.com/TuringLang/AdvancedHMC.jl/pull/208 merged first so that this PR can be Geweke tested,
which would give us some confidence that the refactoring didn't break anything.

xukai92:
As a validation of this refactoring doesn't change any functionality,
I run `test/demo.jl` for this branch (62bfed7) and the master, and compare the simulated chains and the statistics,
which are exactly the same.

xukai92:
I added regression tests to compare the exact simulated chains and statistics in daade0a1a01f834d68b5af705d0bf9a4bee4f0ba for the four variants.

- Static HMC with MH
- Static HMC with multinomial
- NUTS with slice
- NUTS with multinomial

These should be a good enough indicator that the functionality doesn't got changed.

xukai92:
I disabled the regression test on CI because the results from master is platform & Julia version dependent.
The file generated is from my local (macOS) with Julia 1.5-rc2;
I will make sure all my commits pass the test locally.
If we were to support it on CI we need to generate chains and statistics for all variants, platforms and Julia version.
I would consider it as a future work.

xukai92:
Just a few notes:

- I believe this option is more modular in terms of constructing HMC algorithms on a single target. It can be seen how it results into a clean construction for users and a clean multiple dispatch in our code base.
- @yebai If you agree with this design, I'm happy to break this PR into a few small ones to make the review eaiser.

yebai:
Yes, it can be helpful to break this PR into two smaller ones: one focusing on unifying trajectory types, another focusing on momentum refreshment.

xukai92:
Yes. Let me close this one first and create one for unifying trajectory types first, which is less controversial.

PR #214, opened 2020-08-05
Unifying trajectories

xukai92:
This PR replaces #213.

This PR solves #103 by explicitly separating `Trajectory` from `HMCKernel`,
without introducing any functionality change.
I left some comments on how to unify the notation as discussed in #48,
in which I use `τ` to refer a (numerical) Hamiltonian trajectory and `κ` to refer a MCMC kernel.
The remaining would be done in a separate PR to make the review easier.

As a showcase, below is how to construct NUTS now:

```julia
τ = Trajectory(Leapfrog(1e-3), NoUTurn())
κ = HMCKernel(τ, MultinomialTS)
```

Old syntax is still supported e.g. as below
```julia
NUTS{TS, TC}(int::AbstractIntegrator) where {TS, TC} = HMCKernel(Trajectory(int, TC()), TS)
```

As suggested by https://github.com/TuringLang/AdvancedHMC.jl/issues/103#issue-496507285,
fixed simulation steps or length are unified as termination criteria.
A nice benefit of this is that the parameters regarding trajectory simulation are clear:
e.g. `FixedNSteps(10)`, `NoUTurn(5, 1_000)` or `NoUTurn(max_depth=5)`.

As a result, the main `transition` function looks like this.
```julia
function transition(rng, h, κ::HMCKernel, z)
    @unpack τ, TS = κ
    z = refresh(rng, z, h)
    return transition(rng, h, τ, TS, z)
end
```
As a side product of this new interface,
it also unifies how integrators are jittered - all before simulating the trajectory in the same place where momentum variables are refreshed.
With this design, this is the only place where `jitter` is called.

A list of changes is

1.  The new interface
2. `GeneralisedNoUTurn` -> `NoUTurn` and `StrictGeneralisedNoUTurn` -> `StrictNoUTurn`
3. Store the termination statistics directly in the binary tree
    - Why? Because termination criteria like `NoUTurn` are in the same role as `FixedNSteps` now. To unify the design, they are only intended to store the algorithm parameters, e.g. number of steps, maximum depth, etc.
4. All tests are updated accordingly
5. A regression test is added to make sure there is no functionality change compared to ef6de39
    - It can only run locally for now using Julia 1.5 on macOS. This is because each Julia version or hardware would generate different chains and I only pre-generated those for my own machine.
    - Making this fully supported on CI is left as a future work.


xukai92:
As mentioned in https://github.com/TuringLang/AdvancedHMC.jl/pull/213#issuecomment-667617776,
I aslo added some regression test by comparing this PR against the master explicitly - see https://github.com/TuringLang/AdvancedHMC.jl/pull/214/commits/31c65179bad16dcd9c431ca259a1d81816a6b26e.
This ensures that this PR doesn't change the functionality at all.

README.md
```diff
@@ -54,7 +54,7 @@ integrator = Leapfrog(initial_ϵ)
 #   - multinomial sampling scheme,
 #   - generalised No-U-Turn criteria, and
 #   - windowed adaption for step-size and diagonal mass matrix
-proposal = NUTS{MultinomialTS, GeneralisedNoUTurn}(integrator)
+proposal = NUTS{MultinomialTS, NoUTurn}(integrator)
```
yebai, line 57: Maybe keep the old no-U-turn type names like `GeneralisedNoUTurn` and `StrictGeneralisedNoUTurn` to minimise unnecessary code changes, which might upset some AHMC users. As an alternative, we can create a type alias `NoUTurn = GeneralisedNoUTurn` for convenience.

src/AdvancedHMC.jl
```diff
@@ -18,11 +18,19 @@ import Parameters: reconstruct
...
 # r: momentum variables
 # z: phase point / a pair of θ and r
+# θ₀: initial position
+# r₀: initial momentum
+# z₀: initial phase point
+# ℓπ: log density of the target distribution
+# ∇ℓπ: gradient of the log density of the target distribution w.r.t θ
+# κ: kernel
+# τ: trajectory
+# ϵ: step size
```
yebai, line 31:
```suggestion
# ϵ: leap-frog integration step size
```

src/AdvancedHMC.jl
```diff
@@ -18,11 +18,19 @@ import Parameters: reconstruct
...
+# L: step number
```
yebai, line 32:
```suggestion
# L: leap-frog integration step number
```

src/AdvancedHMC.jl
```diff
@@ -18,11 +18,19 @@ import Parameters: reconstruct
...
+# λ: integration time
```
yebai, line 33:
```suggestion
# λ: leap-frog integration  time
```

src/sampler.jl
```diff
@@ -8,14 +8,19 @@ function reconstruct(
...
 
-reconstruct(τ::AbstractProposal, ::AbstractAdaptor) = τ
+reconstruct(κ::AbstractProposal, ::AbstractAdaptor) = κ
 function reconstruct(
-    τ::AbstractProposal, adaptor::Union{StepSizeAdaptor, NaiveHMCAdaptor, StanHMCAdaptor}
+    κ::AbstractProposal, adaptor::Union{StepSizeAdaptor, NaiveHMCAdaptor, StanHMCAdaptor}
 )
+    return reconstruct(κ, getϵ(adaptor))
+end
+function reconstruct(κ::AbstractProposal, ϵ::AbstractScalarOrVec{<:Real})
```
yebai, line 17: Is this new `reconstruct` function used elsewhere? If not, maybe we can keep the old implementation which avoids creating one additional function.

src/sampler.jl
```diff
@@ -8,14 +8,19 @@ function reconstruct(
     return reconstruct(h, metric=metric)
 end
 
-reconstruct(τ::AbstractProposal, ::AbstractAdaptor) = τ
+reconstruct(κ::AbstractProposal, ::AbstractAdaptor) = κ
```
yebai, line 11: Maybe: `AbstractProposal` ==> `AbstractKernel`? This can be done in a separate PR. Also, there might be a similar type in `AdvancedMH` which we can inherit here.

src/sampler.jl
```diff
@@ -175,18 +180,18 @@ function sample(
     pm = progress ? ProgressMeter.Progress(n_samples, desc="Sampling", barlen=31) : nothing
     time = @elapsed for i = 1:n_samples
         # Make a step
-        t = step(rng, h, τ, t.z)
-        # Adapt h and τ; what mutable is the adaptor
+        t = transition(rng, h, κ, t.z)
```
yebai, line 183: Any reason for changing function name `step` into `transition`? Maybe consider making it more aligned with `AbstractMCMC` function names.

src/trajectory.jl
```diff
@@ -33,9 +33,45 @@ Abstract Markov chain Monte Carlo proposal.
...
 """
-abstract type AbstractTrajectory{I<:AbstractIntegrator} <: AbstractProposal end
+struct FixedNSteps{T<:Integer} <: StaticTerminationCriterion
+    L::T
+end
+
+"""
+$(TYPEDEF)
+
+Standard HMC implementation with fixed total trajectory length.
```
yebai, line 63:
```suggestion
Standard HMC implementation with a fixed number of leap-frog integration steps.
```

src/trajectory.jl
```diff
@@ -33,9 +33,45 @@ Abstract Markov chain Monte Carlo proposal.
...
+
+# Fields
+
+$(TYPEDFIELDS)
+
+# References
+1. Neal, R. M. (2011). MCMC using Hamiltonian dynamics. Handbook of Markov chain Monte Carlo, 2(11), 2. ([arXiv](https://arxiv.org/pdf/1206.1901)) 
+"""
+struct FixedLength{F<:AbstractFloat} <: StaticTerminationCriterion
```
yebai, line 72: I think `FixedIntegrationTime` is a more accurate term here. Maybe rename it?

src/trajectory.jl
```diff
@@ -51,7 +87,7 @@ $(TYPEDEF)
 
 Samples the end-point of the trajectory.
 """
-struct EndPointTS <: AbstractTrajectorySampler end
+struct MetropolisTS <: AbstractTrajectorySampler end
```
yebai, line 90: This makes sense. We should warn the user though since this is breaking.

src/trajectory.jl
```diff
@@ -110,23 +146,23 @@ Ref: https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/nuts/base_n
 MultinomialTS(rng::AbstractRNG, z0::PhasePoint) = MultinomialTS(z0, zero(energy(z0)))
 
 """
-    SliceTS(s::SliceTS, H0::AbstractFloat, zcand::PhasePoint)
+    init_leafsampler(s::SliceTS, H0::AbstractFloat, zcand::PhasePoint)
```
yebai, line 149: `leafsampler` is a vague term, I think. I prefer keeping the old type name based constructor here.

src/trajectory.jl
```diff
@@ -161,67 +197,58 @@ function mh_accept(rng::AbstractRNG, s::MultinomialTS, s′::MultinomialTS)
 end
 
 """
-    transition(τ::AbstractTrajectory{I}, h::Hamiltonian, z::PhasePoint)
-
-Make a MCMC transition from phase point `z` using the trajectory `τ` under Hamiltonian `h`.
-
-NOTE: This is a RNG-implicit fallback function for `transition(GLOBAL_RNG, τ, h, z)`
+Hamiltonian dynamics numerical simulation trajectories.
```
yebai, line 200:
```suggestion
Hamiltonian dynamics numerically simulated trajectories.
```

README.md
xukai92: Sure!

src/sampler.jl
xukai92: Yes I want to do this change as well but just as you said I want to avoid unnecessary changes for this PR.

src/sampler.jl
xukai92: It's used in the test.

src/sampler.jl
```diff
@@ -47,28 +52,28 @@ end
 function step(
```
xukai92, line 52: @yebal This is the old `step`.

src/sampler.jl
xukai92: The previous `step` function in Line 41 was just a wrapper over `transition` (I pinged you there).
With the new abstraction this wrapper is unnecessary.

src/trajectory.jl
```diff
@@ -517,35 +442,48 @@ Base.:*(d1::Termination, d2::Termination) = Termination(d1.dynamic || d2.dynamic
...
     zright  # right most leaf node
-    c::C    # termination criterion
+    rho::R  # termination statistic
     sum_α   # MH stats, i.e. sum of MH accept prob for all leapfrog steps
     nα      # total # of leap frog steps, i.e. phase points in a trajectory
     ΔH_max  # energy in tree with largest absolute different from initial energy
 end
 
+"Initialize termination statistic"
+rho_init(::ClassicNoUTurn, z) = nothing
```
yebai, line 475: Consider introducing a `TurnStatistic` here. Then `rho_init` can be replaced with a `TurnStatistic` constructor overloaded for various turn termination types, `rho_merge` can be replaced with `combine(::TurnStatistic, ::TurnStatistic)`.

src/trajectory.jl
```diff
@@ -517,35 +442,48 @@ Base.:*(d1::Termination, d2::Termination) = Termination(d1.dynamic || d2.dynamic
 isterminated(d::Termination) = d.dynamic || d.numerical
 
 """
-    Termination(s::SliceTS, nt::NUTS, H0::F, H′::F) where {F<:AbstractFloat}
+    Termination(s::SliceTS, tc, H0::F, H′::F) where {F<:AbstractFloat}
```
yebai, line 445:
```suggestion
     $(SIGNATURES)
```

src/trajectory.jl
```diff
@@ -563,22 +501,22 @@ function combine(treeleft::BinaryTree, treeright::BinaryTree)
...
+        rho_merge(treeleft.rho, treeright.rho),
         treeleft.sum_α + treeright.sum_α,
         treeleft.nα + treeright.nα,
         maxabs(treeleft.ΔH_max, treeright.ΔH_max),
     )
 end
 
 """
-    isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
+    isterminated(h::Hamiltonian, t::BinaryTree)
```
yebai, line 512:
```suggestion
     $(SIGNATURES)
```

src/trajectory.jl
```diff
@@ -563,22 +501,22 @@ function combine(treeleft::BinaryTree, treeright::BinaryTree)
...
 
 Detect U turn for two phase points (`zleft` and `zright`) under given Hamiltonian `h`
 using the (original) no-U-turn cirterion.
 
 Ref: https://arxiv.org/abs/1111.4246, https://arxiv.org/abs/1701.02434
 """
-function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
+function isterminated(::ClassicNoUTurn, h::Hamiltonian, t::BinaryTree)
```
yebai, line 519: After introducing a parametric turn statistic, this function can be simplified into 


```julia

function isterminated(left::TurnStatistic{ClassicNoUTurn}, right::TurnStatistic{ClassicNoUTurn}, h::Hamiltonian)

```

src/trajectory.jl
```diff
@@ -587,79 +525,67 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
 end
 
 """
-    isterminated(h::Hamiltonian, t::BinaryTree{<:GeneralisedNoUTurn})
+    isterminated(h::Hamiltonian, t::BinaryTree)
```
yebai, line 528:
```suggestion
    $(SIGNATURES)
```

Similiar for all other places.

src/trajectory.jl
```diff
@@ -587,79 +525,67 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
 
 Detect U turn for two phase points (`zleft` and `zright`) under given Hamiltonian `h`
 using the generalised no-U-turn criterion.
 
 Ref: https://arxiv.org/abs/1701.02434
 """
-function isterminated(h::Hamiltonian, t::BinaryTree{<:GeneralisedNoUTurn})
-    rho = t.c.rho
+function isterminated(::Union{NoUTurn, StrictNoUTurn}, h::Hamiltonian, t::BinaryTree)
```
yebai, line 535: See comment on https://github.com/TuringLang/AdvancedHMC.jl/pull/214/files#diff-f48defe9803ddea4f759bce857b70de0R519

src/trajectory.jl
```diff
@@ -587,79 +525,67 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
-        t.nα,
-        t.ΔH_max
-    )
-    s1 = isterminated(h, t_generalised)
+## References 
 
-    # U-turn checks for left and right subtree
-    # See https://discourse.mc-stan.org/t/nuts-misses-u-turns-runs-in-circles-until-max-treedepth/9727/33
-    # for a visualisation.
+1. https://arxiv.org/abs/1701.02434
```
yebai, line 551: Consider adding the paper title and author information here.

src/trajectory.jl
```diff
@@ -587,79 +525,67 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
     return Termination(s, false)
 end
 
+isterminated(tc::Union{ClassicNoUTurn, NoUTurn}, h, t, tleft, tright) = isterminated(tc, h, t)
+
 """
-    isterminated(
-        h::Hamiltonian, t::T, tleft::T, tright::T
-    ) where {T<:BinaryTree{<:StrictGeneralisedNoUTurn}}
+    isterminated(tc::StrictNoUTurn, h, t, tleft, tright)
```
yebai, line 544:
```suggestion
    $(SIGNATURES)
```

src/trajectory.jl
```diff
@@ -587,79 +525,67 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
+    #         the rightmost phase point of `tleft`, the left subtree.
     s3 = check_right_subtree(h, t, tleft, tright)
     return s1 * s2 * s3
 end
 
 """
-    check_left_subtree(
-        h::Hamiltonian, t::T, tleft::T, tright::T
-    ) where {T<:BinaryTree{<:StrictGeneralisedNoUTurn}}
+    check_left_subtree(h::Hamiltonian, t::T, tleft::T, tright::T) where {T<:BinaryTree}
```
yebai, line 570:
```suggestion
    $(SIGNATURES)
```

src/trajectory.jl
```diff
@@ -587,79 +525,67 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
+    rho = tleft.rho + tright.zleft.r
     s = generalised_uturn_criterion(rho, ∂H∂r(h, t.zleft.r), ∂H∂r(h, tright.zleft.r))
     return Termination(s, false)
 end
 
 """
-    check_left_subtree(
-        h::Hamiltonian, t::T, tleft::T, tright::T
-    ) where {T<:BinaryTree{<:StrictGeneralisedNoUTurn}}
+    check_right_subtree(h::Hamiltonian, t::T, tleft::T, tright::T) where {T<:BinaryTree}
```
yebai, line 582:
```suggestion
    $(SIGNATURES)
```

src/trajectory.jl
```diff
@@ -668,79 +594,63 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
-    rng::AbstractRNG,
-    nt::NUTS{S,C,I,F},
-    h::Hamiltonian,
-    z::PhasePoint,
-    sampler::AbstractTrajectorySampler,
-    v::Int,
-    j::Int,
-    H0::AbstractFloat,
-) where {I<:AbstractIntegrator,F<:AbstractFloat,S<:AbstractTrajectorySampler,C<:AbstractTerminationCriterion}
+function build_tree(rng, int, tc, h, z, sampler, v, j, H0)
```
yebai, line 600: I think we should keep these type information. It makes the code more readable. and it doesn't hurt to have a few extra lines.

src/trajectory.jl
```diff
@@ -668,79 +594,63 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
     if j == 0
         # Base case - take one leapfrog step in the direction v.
-        z′ = step(nt.integrator, h, z, v)
+        z′ = step(int, h, z, v)
```
yebai, line 603: Consider keeping previous variable name: `integrator` is a lot more readable than `int`.

src/trajectory.jl
```diff
@@ -668,79 +594,63 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
     end
 end
 
 function transition(
-    rng::AbstractRNG,
-    τ::NUTS{S,C,I,F},
-    h::Hamiltonian,
-    z0::PhasePoint,
-) where {I<:AbstractIntegrator,F<:AbstractFloat,S<:AbstractTrajectorySampler,C<:AbstractTerminationCriterion}
+    rng, h, τ::Trajectory{I, C}, ::Type{TS}, z0
```
yebai, line 632: Why changing the code style? I think the old code style is clearer. Perhaps keep that? Similarly for all other places.

src/trajectory.jl
```diff
@@ -668,79 +594,63 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
+) where {I, C<:DynamicTerminationCriterion, TS}
+    @unpack integrator, criterion = τ
```
yebai, line 634: `criterion` is unclear, consider renaming to `termination_ criterion`. It doesn't hurt to use descriptive variable names.

test/demo.jl
```diff
@@ -1,30 +1,30 @@
...
+ϵ₀ = find_good_stepsize(hamiltonian, θ₀)
+integrator = Leapfrog(ϵ₀)
 
 # Define an HMC sampler, with the following components
-#   - multinomial sampling scheme,
-#   - generalised No-U-Turn criteria, and
-#   - windowed adaption for step-size and diagonal mass matrix
-proposal = NUTS{MultinomialTS, GeneralisedNoUTurn}(integrator)
-adaptor = StanHMCAdaptor(MassMatrixAdaptor(metric), StepSizeAdaptor(0.8, integrator))
+#   - (Generalised) no-U-turn criteria simulated by the defined integrator
```
yebai, line 21:
```suggestion
#   - (Generalised) no-U-turn criteria for terminating Hamiltonian simulation
```

test/regression.jl
```diff
@@ -0,0 +1,63 @@
+using Test, Random, BSON, AdvancedHMC, Distributions, ForwardDiff
+
+is_ef6de39 = isdefined(AdvancedHMC, :EndPointTS)
```
yebai, line 3: A better way of performing this regression test is to check out a pinned version of AHMC, run the tests on CI machines with the pinned version and the current version. Then we can compare results without worrying machine-specific details.

test/sampler-vec.jl
```diff
@@ -73,40 +76,42 @@ include("common.jl")
...
             h = Hamiltonian(metricT((D, n_chains)), ℓπ, ∂ℓπ∂θ)
-            t = @elapsed samples, stats = sample(h, τ, θ_init_list[i], n_samples; verbose=false)
-            time_mat[i] = t
+            t = @elapsed samples, stats = sample(h, κ, θ_init_list[i], n_samples; verbose=false)
+            time_vec[i] = t
         end
 
         # Time for multiple runs of single chain
-        time_seperate = Vector{Float64}(undef, n_chains_max)
+        time_sep = Vector{Float64}(undef, n_chains_max)
```
yebai, line 95: `time_seperate` is more readable than `time_sep`.

test/trajectory.jl
```diff
@@ -217,26 +221,23 @@ end
...
 
+    # Step 1 & 2: U-turn checks for left and right subtree; see [3] for a visualisation.
+    # Step 1: Check U-turn between the leftmost phase point of `t` and 
+    #         the leftmost  phase point of `tright`, the right subtree.
     s2 = AdvancedHMC.check_left_subtree(h, t, tleft, tright)
+    # Step 2: Check U-turn between the rightmost phase point of `t` and
+    #         the rightmost phase point of `tleft`, the left subtree.
     s3 = AdvancedHMC.check_right_subtree(h, t, tleft, tright)
+    return s1 * s2 * s3
     @test s1 == s2 == s3
```
yebai, line 241: This line won't run since the previous line has returned.

test/trajectory.jl
```diff
@@ -251,13 +252,13 @@ end
...
             rho = cumsum(traj_r, dims=2)
             
-            ts_hand_isturn_fwd = hand_isturn.(Ref(traj_z[1]), traj_z, [rho[:,i] for i = 1:length(traj_z)], Ref(1))
-            ts_ahmc_isturn_fwd = ahmc_isturn.(Ref(traj_z[1]), traj_z, [rho[:,i] for i = 1:length(traj_z)], Ref(1))
+            ts_hand_isturn_fwd = hand_isturn_classic.(Ref(traj_z[1]), traj_z, [rho[:,i] for i = 1:length(traj_z)], Ref(1))
+            ts_ahmc_isturn_fwd = ahmc_isturn_classic.(Ref(traj_z[1]), traj_z, [rho[:,i] for i = 1:length(traj_z)], Ref(1))
 
-            ts_hand_isturn_generalised_fwd = hand_isturn_generalised.(Ref(traj_z[1]), traj_z, [rho[:,i] for i = 1:length(traj_z)], Ref(1))
-            ts_ahmc_isturn_generalised_fwd = ahmc_isturn_generalised.(Ref(traj_z[1]), traj_z, [rho[:,i] for i = 1:length(traj_z)], Ref(1))
+            ts_hand_isturn_generalised_fwd = hand_isturn.(Ref(traj_z[1]), traj_z, [rho[:,i] for i = 1:length(traj_z)], Ref(1))
```
yebai, line 258: It's ok to keep the current name `hand_isturn_generalised`, which is more informative.

test/trajectory.jl
```diff
@@ -186,29 +186,33 @@ function gettraj(rng, ϵ=0.1, n_steps=50)
...
 
-ahmc_isturn(z0, z1, rho, v=1) =
-    AdvancedHMC.isterminated(h, AdvancedHMC.BinaryTree(z0, z1, ClassicNoUTurn(), 0, 0, 0.0)).dynamic
+function ahmc_isturn_classic(z0, z1, rho, v=1)
+    tree = AdvancedHMC.BinaryTree(z0, z1, nothing, 0, 0, 0.0)
+    return AdvancedHMC.isterminated(ClassicNoUTurn(), h, tree).dynamic
+end
 
-function hand_isturn_generalised(z0, z1, rho, v=1)
+function hand_isturn(z0, z1, rho, v=1)
```
yebai, line 200: It's ok to keep the current name `hand_isturn_generalised`, which is more informative.

yebai:
Thanks, @xukai92. I've done one full pass of this PR; it looks overall good to me. I left many comments above.

src/sampler.jl
yebai: In that case, I suggest we remove this function.

xukai92: Sounds good. I will merge it back to Line 12.

src/trajectory.jl
xukai92: Good suggestion!

src/trajectory.jl
xukai92: It was actually suggested by you :smile: I can make sure to add a deprecation for this.

src/trajectory.jl
```diff
@@ -161,67 +197,58 @@ function mh_accept(rng::AbstractRNG, s::MultinomialTS, s′::MultinomialTS)
...
-    integrator  ::  I
-    "Number of steps to simulate, i.e. length of trajectory will be `n_steps + 1`."
-    n_steps     ::  Int
-end
+    transition(τ::AbstractProposal, h::Hamiltonian, z::PhasePoint)
 
-function Base.show(io::IO, τ::StaticTrajectory{<:EndPointTS})
-    print(io, "StaticTrajectory{EndPointTS}(integrator=$(τ.integrator), λ=$(τ.n_steps)))")
-end
+Make a MCMC transition from phase point `z` using the trajectory `τ` under Hamiltonian `h`.
```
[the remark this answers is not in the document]
xukai92: Why do we need use `an` here?

src/trajectory.jl
```diff
@@ -233,6 +260,14 @@ function transition(
     return Transition(z, tstat)
 end
 
+function transition(rng, h, τ::Trajectory{I, <:FixedLength}, ::Type{TS}, z) where {I, TS}
+    @unpack integrator, criterion = τ
+    # Create the corresponding `FixedNSteps` criterion
+    L = max(1, floor(Int, criterion.λ / nom_step_size(integrator)))
```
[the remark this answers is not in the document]
xukai92: We would need to do this for each transition because `L` is step size dependent, e.g. for HMCDA.
If we were to implement this when reconstructing trajectory, the same computation is needed.
Also I think this computaitn is negligible.

src/trajectory.jl
```diff
@@ -421,80 +410,16 @@ $(TYPEDFIELDS)
...
-
-Create an instance for the No-U-Turn sampling algorithm
-with multinomial sampling and original no U-turn criterion.
-
-Below is the doc for NUTS{S,C}.
-
-$NUTS_DOCSTR
-"""
-NUTS(args...) = NUTS{MultinomialTS, GeneralisedNoUTurn}(args...)
+combine(cleft::T, cright::T) where {T<:NoUTurn} = T(cleft.rho + cright.rho)
```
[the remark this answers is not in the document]
xukai92: Actually should be `tleft` and `tright` because the varaibles are trees.

src/trajectory.jl
xukai92: I quite like this suggestion. Do we want to implement in this PR or a seperate though? It seems need some careful cares.

src/trajectory.jl
xukai92: I don't mind add them back. I'm not removing them because they take a few lines. I'm following the principle of "annotating types only if they are required for multiple dispatch", which may or may not be appreciated by everyone.

src/trajectory.jl
xukai92: Sure.

src/trajectory.jl
xukai92: Sure. It was `termination_ criterion` originally but I thought the context is clear enough so I simplified it. I don't mind rolling them back.

test/regression.jl
xukai92: Yes I agree. I don't know how to do that though. Can you point me some example CI codes on doing so?

test/sampler-vec.jl
xukai92: This is a typo, should be `time_seq` (for sequential) or maybe `time_loop`.

test/trajectory.jl
xukai92: I will remove the previous line.

test/trajectory.jl
xukai92: I just want make it clear why I did this name change of `GeneralisedNoUTurn` -> `NoUTurn`.
Originally, we only have the classic no-U-turn and when we add the generalised no-U-turn we want to make it clear that it's the generalised version.
However, I believe when refer to no-U-turn, people now by default refer to the generalised version.
So in this context, `NoUTurn` by default refers to the generalised version and we use the explicit `Classic` prefix to refer to the original version and the `Strict` prefix refers to a more strict version of `NoUTurn`.

I'm fine to revert these changes though and have some more discussion on this before commiting it.

yebai: Yes, let's keep things descriptive internally. We can create a type alias for `const NoUTurn=GeneralisedNoUTurn`.

test/regression.jl
yebai: [`BenchmarkTools`](https://github.com/JuliaCI/BenchmarkTools.jl) allows users to write tests, and run regression tests automatically. Our use here is a bit special, but the principles are the same I think.

src/trajectory.jl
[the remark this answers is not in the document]
yebai: Thanks for the clarification!

src/trajectory.jl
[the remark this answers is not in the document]
yebai: It's because the first letter in `MCMC` is `M` which begins with vowel sounds.

test/regression.jl
xukai92: Can you point me an example? I only found example where they are looking at regression in time.

src/trajectory.jl
xukai92: Reverted.

src/trajectory.jl
```diff
@@ -161,67 +197,58 @@ function mh_accept(rng::AbstractRNG, s::MultinomialTS, s′::MultinomialTS)
...
-###
+Base.show(io::IO, τ::Trajectory) = print(io, "Trajectory(integrator=$(τ.integrator), criterion=$(τ.criterion))")
 
-"""
-$(TYPEDEF)
-
-Static HMC with a fixed number of leapfrog steps.
+struct HMCKernel{T<:Trajectory, TS<:AbstractTrajectorySampler} <: AbstractProposal
+    τ::T
+    TS::Type{TS}
```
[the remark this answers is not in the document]
xukai92: I changed it to an explict name `trajectory_sampler_type`. I think there is space to improve the interface: Currently the use of type and instance is a bit weird. But let's do it in another dedicated PR.

src/trajectory.jl
```diff
@@ -161,67 +197,58 @@ function mh_accept(rng::AbstractRNG, s::MultinomialTS, s′::MultinomialTS)
...
+###
 
-function transition(
-    rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
-    τ::StaticTrajectory,
-    h::Hamiltonian,
-    z::PhasePoint,
-)
+function transition(rng, h, τ::Trajectory{I, <:FixedNSteps}, ::Type{TS}, z) where {I, TS}
+    @unpack integrator, criterion = τ
```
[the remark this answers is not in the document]
xukai92: Done for all.

src/trajectory.jl
xukai92: Done.

test/trajectory.jl
xukai92: Done.

src/trajectory.jl
xukai92: Reverted both `build_tree` and `transition`.

xukai92:
@yebai This is ready for another look.

Two things I *didn't* address are:
1. Replacing `rho` by `TurnStatistics` 
    - I feel it's better to do this in another PR as it involves changes in many places.
2. Finding a better way to handle `TS`
    - I did rename it to `trajectory_sampler_type` which is more descriptive
    - Ideally I feel we need that we can avoid storing types in `HMCKernel` once we introduce `TurnStatistics`
        - As you suggested, we can parameterize `TurnStatistics` by termination criteria so does trajectory sampling methods
        - If we do this, we don't need to store turn statistics (for trajectory sampling) inside `XXXTS` anymore.
        - Then `MultinomialTS()` (instance) can be used over `MultinomialTS` (type) during construction, which I think is better.
    - Open for discussion though. What's your opinion?

Also, do you think we should introduce a `dev` branch to AHMC?

src/trajectory.jl
```diff
@@ -668,79 +600,75 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
+    τ::Trajectory{I, C},
+    ::Type{TS},
     z0::PhasePoint,
-) where {I<:AbstractIntegrator,F<:AbstractFloat,S<:AbstractTrajectorySampler,C<:AbstractTerminationCriterion}
+) where {I, C<:DynamicTerminationCriterion, TS}
+    @unpack integrator, termination_criterion = τ
     H0 = energy(z0)
-    tree = BinaryTree(z0, z0, C(z0), zero(F), zero(Int), zero(H0))
-    sampler = S(rng, z0)
+    tree = BinaryTree(z0, z0, termination_criterion, zero(H0), zero(Int), zero(H0))
```
sethaxen, line 654: Does switching from `zero(F)` to `zero(H0)` pose any risks?

sethaxen (commented):
This is an excellent refactor! I had a few minor questions/suggestions. Some of them are not necessarily on changes in this PR and don't need to be handled here, but they came up while I was reading.

src/trajectory.jl
```diff
@@ -30,12 +30,48 @@ stat(t::Transition) = t.stat
...
+1. Neal, R. M. (2011). MCMC using Hamiltonian dynamics. Handbook of Markov chain Monte Carlo, 2(11), 2. ([arXiv](https://arxiv.org/pdf/1206.1901))
+"""
+struct FixedNSteps{T<:Integer} <: StaticTerminationCriterion
+    L::T
+end
+
+"""
+$(TYPEDEF)
+
+Standard HMC implementation with a fixed number of leap-frog integration steps.
```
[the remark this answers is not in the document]
xukai92: The other one called `FixedIntegrationTime` is (dedicated) for fixed integration time.

src/trajectory.jl
```diff
@@ -400,11 +389,13 @@ $(TYPEDFIELDS)
...
 """
-struct GeneralisedNoUTurn{T<:AbstractVector{<:Real}} <: AbstractTerminationCriterion
-    "Integral or sum of momenta along the integration path."
-    rho::T
+@with_kw struct GeneralisedNoUTurn{F<:Real} <: DynamicTerminationCriterion
+    max_depth::Int=10
+    Δ_max::F=1000
 end
 
+const NoUTurn = GeneralisedNoUTurn
```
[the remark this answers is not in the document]
xukai92: I want to make the transition of calling the old ones `ClassificNoUTurn` and new ones just `NoUTurn` and `StrictNoUTurn`, i.e. by default we refer to the generalised ones if not specified.

src/trajectory.jl
```diff
@@ -587,79 +531,67 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
-    s1 = isterminated(h, t_generalised)
+## References 
+
+1. Betancourt, M. (2017). A Conceptual Introduction to Hamiltonian Monte Carlo. [arXiv preprint arXiv:1701.02434](https://arxiv.org/abs/1701.02434).
+2. https://github.com/stan-dev/stan/pull/2800
+3. https://discourse.mc-stan.org/t/nuts-misses-u-turns-runs-in-circles-until-max-treedepth/9727/33
+"""
+function isterminated(tc::StrictGeneralisedNoUTurn, h, t, tleft, tright)
+    # Step 0: original generalised U-turn check
+    s1 = isterminated(tc, h, t)
```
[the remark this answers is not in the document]
xukai92: That's a good idea! Can you do that in your NUTS PR?

src/trajectory.jl
```diff
@@ -587,79 +531,67 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
 
 Do a U-turn check between the leftmost phase point of `t` and the leftmost 
 phase point of `tright`, the right subtree.
 """
-function check_left_subtree(
-    h::Hamiltonian, t::T, tleft::T, tright::T
-) where {T<:BinaryTree{<:StrictGeneralisedNoUTurn}}
-    rho = tleft.c.rho + tright.zleft.r
+function check_left_subtree(h::Hamiltonian, t::T, tleft::T, tright::T) where {T<:BinaryTree}
+    rho = tleft.rho + tright.zleft.r
```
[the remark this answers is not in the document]
xukai92: I'm happy to make the proposed change if some benchmark shows it improves the performance.

src/trajectory.jl
xukai92: I think it's fine and actually better. Previous `F` was the type of `Δ_max ` while we are now using the actual type of the target.

src/trajectory.jl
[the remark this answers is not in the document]
sethaxen: Isn't this the docstring for `FixedIntegrationTime`?

src/trajectory.jl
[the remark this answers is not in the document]
sethaxen: Perhaps it would make sense to go the other way around, then? Define `NoUTurn` and then make `GeneralisedNoUTurn` the alias so that users' legacy code works?

src/trajectory.jl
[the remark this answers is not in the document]
sethaxen: sure!

src/trajectory.jl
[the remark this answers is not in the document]
sethaxen: Here's a benchmark of one with the same time complexity:
```julia
julia> using LinearAlgebra, BenchmarkTools, Plots

julia> function foo(a, b, c, d)
           e = a + b
           f = c + d
           dot(e, f)
       end;

julia> function foo2(a, b, c, d)
           dot(a, c) + dot(a, d) + dot(b, c) + dot(b, d)
       end;

julia> ns = 10 .^ (0:6);

julia> times = map(ns) do n
           a, b, c, d = ntuple(_ -> randn(n), 4)
           (@belapsed($foo($a,$b,$c,$d)), @belapsed($foo2($a,$b,$c,$d)))
       end;

julia> plot(ns, [first.(times) last.(times)]; xscale=:log10, yscale=:log10, labels=["allocating" "non-allocating"])
```
tmp

I can include it in a more general allocation-reducing PR though.

test/regression.jl
sethaxen: In my view, no single version of the code should be considered the ground truth. Rather, we can only be confident that features covered by the test suite work correctly. Hence instead of checking out a previous version to test against, it's better to add any missing tests, ensuring they pass on both the old version and the new version.

sethaxen: For performance regressions, you can use PkgBenchmark (see e.g. https://github.com/JuliaFolds/FLoops.jl/blob/master/.github/workflows/benchmark.yml), but I don't think that checks the outputs.

src/trajectory.jl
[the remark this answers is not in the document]
xukai92: Lol yes you're right.

src/trajectory.jl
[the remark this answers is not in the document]
xukai92: There were some debates between me and Hong on if we want to make this transition, and we ended up with this compromise. So with the PR, the recommended way is still the previous version. And we want to use the shorter version privately for a while before making the transition.

src/trajectory.jl
```diff
@@ -517,35 +448,48 @@ Base.:*(d1::Termination, d2::Termination) = Termination(d1.dynamic || d2.dynamic
...
 """
-function Termination(s::SliceTS, nt::NUTS, H0::F, H′::F) where {F<:AbstractFloat}
-    return Termination(false, !(s.ℓu < nt.Δ_max + -H′))
+function Termination(s::SliceTS, tc, H0::F, H′::F) where {F<:AbstractFloat}
+    return Termination(false, !(s.ℓu < tc.Δ_max + -H′))
 end
 
 """
-    Termination(s::MultinomialTS, nt::NUTS, H0::F, H′::F) where {F<:AbstractFloat}
+    Termination(s::MultinomialTS, tc, H0::F, H′::F) where {F<:AbstractFloat}
```
yebai, line 460:
```suggestion
     $(SIGNATURES)
```

src/trajectory.jl
```diff
@@ -517,35 +448,48 @@ Base.:*(d1::Termination, d2::Termination) = Termination(d1.dynamic || d2.dynamic
...
-    return Termination(false, !(-H0 < nt.Δ_max + -H′))
+function Termination(s::MultinomialTS, tc, H0::F, H′::F) where {F<:AbstractFloat}
+    return Termination(false, !(-H0 < tc.Δ_max + -H′))
 end
 
 """
 A full binary tree trajectory with only necessary leaves and information stored.
 """
-struct BinaryTree{C<:AbstractTerminationCriterion}
+struct BinaryTree{R<:Union{Nothing, AbstractVecOrMat}}
```
yebai, line 471: I am not sure a union type is needed here as Julia allows `undef` value for struct fields. For example:
```julia
julia> struct F2
       a :: Array
       F2() = new()
       end

julia> F2()
F2(#undef)
```
Maybe consider redefine the `BinaryTree` struct, and keep `rho` uninitialised when it is not used (i.e. replacing `Nothing` with `undef`)? This might also offer some performance improvements as it improves type stability.

src/trajectory.jl
```diff
@@ -668,79 +600,75 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
 end
 
-function isterminated(h::Hamiltonian, t::BinaryTree{T}, args...) where {T<:Union{ClassicNoUTurn, GeneralisedNoUTurn}}
-    return isterminated(h, t)
-end
-
-"""
-Recursivly build a tree for a given depth `j`.
-"""
+"Recursively build a tree for a given depth `j`."
```
yebai, line 603:
```suggestion

"""
    $(SIGNATURES)

Recursively build a tree for a given depth `j`.
"""
```

src/trajectory.jl
```diff
@@ -668,79 +600,75 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
+    tc::AbstractTerminationCriterion,
     h::Hamiltonian,
     z::PhasePoint,
-    sampler::AbstractTrajectorySampler,
+    sampler::TS,
     v::Int,
     j::Int,
     H0::AbstractFloat,
-) where {I<:AbstractIntegrator,F<:AbstractFloat,S<:AbstractTrajectorySampler,C<:AbstractTerminationCriterion}
+) where {TS}
```
yebai, line 614:
```suggestion
) where {TS <: AbstractTrajectorySampler}
```

src/trajectory.jl
```diff
@@ -668,79 +600,75 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
     if j == 0
         # Base case - take one leapfrog step in the direction v.
-        z′ = step(nt.integrator, h, z, v)
+        z′ = step(integrator, h, z, v)
         H′ = energy(z′)
         ΔH = H′ - H0
         α′ = exp(min(0, -ΔH))
-        sampler′ = S(sampler, H0, z′)
-        return BinaryTree(z′, z′, C(z′), α′, 1, ΔH), sampler′, Termination(sampler′, nt, H0, H′)
+        sampler′ = TS(sampler, H0, z′)
```
yebai, line 621: Consider using `reconstruct` here for consistency and clarity.

src/trajectory.jl
```diff
@@ -668,79 +600,75 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
         return tree′, sampler′, termination′
     end
 end
 
 function transition(
     rng::AbstractRNG,
-    τ::NUTS{S,C,I,F},
     h::Hamiltonian,
+    τ::Trajectory{I, C},
+    ::Type{TS},
```
yebai, line 649: Ideally, we should avoid passing this `::Type{TS}` around.

src/trajectory.jl
```diff
@@ -668,79 +600,75 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
...
     z0::PhasePoint,
-) where {I<:AbstractIntegrator,F<:AbstractFloat,S<:AbstractTrajectorySampler,C<:AbstractTerminationCriterion}
+) where {I, C<:DynamicTerminationCriterion, TS}
```
yebai, line 651: Why removing these type constraints? 
```suggestion
) where {I<:AbstractIntegrator, C<:DynamicTerminationCriterion, TS<:AbstractTrajectorySampler}
```

test/regression.jl
yebai: Ideally, we can keep PRs small, i.e. one PR for a change/feature. For big changes like this PR, we should try to do it in multiple steps. It would make reviewing code much easier, which leads to much quicker merging too.

yebai (changes requested):
Hi @xukai92, I've done another careful pass of this PR and left some comments. There is now some merge conflicts against the `master` branch. Maybe consider fixing that too. 

Overall, I find reviewing this PR a bit painful ; ) I hope we can keep PRs small and incremental in the future given that the stake of introducing bugs is high.

yebai:
Ps. I now feel the introduction of `HMCKernel` may not be necessary and potentially confusing. It is no longer necessary if we remove the `TS` field. It is inaccurate because, in HMC variants based on dynamic trajectories, the chosen trajectory sampler and the trajectory termination criteria (e.g. no-U-turn) is often coupled. That is, the step of simulating a trajectory and the step of picking a candidate phase point is coupled. We often iterate these two steps to get a final proposal. It is not easy to disentangle these two steps without introducing significant performance regression and more memory usage.

xukai92:
> Hi @xukai92, I've done another careful pass of this PR and left some comments. There is now some merge conflicts against the master branch. Maybe consider fixing that too.

Thanks for taking another pass. I will address the comments soon.

> Overall, I find reviewing this PR a bit painful ; ) I hope we can keep PRs small and incremental in the future given that the stake of introducing bugs is high. For a concrete example, the refactoring of Trajectories, TerminationCriteria, and the introduction of HMCKernel each would deserve its own PR...

Sorry about this. I'm actually learning towards separating it. I will work on extracting out the first PR after cooperating your suggestions.

> Ps. I now feel the introduction of HMCKernel may not be necessary and potentially confusing. It is no longer necessary if we remove the TS field. 

I completely agree that if we remove `TS` filed, we can probably merge `HMCKernel` and `Trajectory `. But just as a reminder, it has a few benefits in the future

1. Compatibility: I don't think we will be happy if we were to compose a HMC `Trajectory` with a MH kernel.
2. Modularity of momentum behaviour: partial momentum, coupling can be made modular; and I believe it's better to consider this step as part of the MCMC kernel
3. This last point is questionable: I kind of feel we should try to make the interface working on instances rather than types, it possible. For example, `NUTS{MultinomialTS, GeneralisedNoUTurn}(Leapfrog(1e-3))` -> `HMCKernel(Trajectory(Leapfrog(1e-3), NoUTurn()), MultinomialTS())`. But anyway this is just my very initial thoughts on this. 

But maybe I could just introduce it later when introducing the modularity of momentum, as we are agreeing on making separate PRs now.

> It is inaccurate because, in HMC variants based on dynamic trajectories, the chosen trajectory sampler and the trajectory termination criteria (e.g. no-U-turn) is often coupled. That is, the step of simulating a trajectory and the step of picking a candidate phase point is coupled. We often iterate these two steps to get a final proposal. It is not easy to disentangle these two steps without introducing significant performance regression and more memory usage.

Yes I see that these two steps are coupled for dynamics trajectories; it is exactly the reason we are *dispatching on termination and trajectory sampler together*, right? Not sure what do you mean by "inaccurate" here. I still think they (trajectory and its sampler) are conceptually two concepts; it's just
- For static trajectories, even the implementation *can be* (but not necessarily) separated.
- For dynamic trajectories, it turns out to be the case that we need to dispatch on them together as the implementation is coupled.

PR #220, opened 2020-10-14
Fix adaptation of nominal step size for JitteredLeapfrog

### LINKED ISSUE #218: ϵ0 of JitteredLeapfrog is not adapted
Hi,
Seems like the stepsize is not being adapted when using `JitteredLeapfrog`.
Guess this is because the member name for the stepsize parameter is `ϵ0` not `ϵ` as the others.
I highly doubt this is expected behavior but I just wanna check to be sure.
In theory, adaptation should also be applicable to `JitteredLeapfrog` right?

sethaxen:
This PR fixes #218 using @xukai92's suggestion in https://github.com/TuringLang/AdvancedHMC.jl/issues/218#issuecomment-707167450.

Here's what we get for @Red-Portal's failing example in https://github.com/TuringLang/AdvancedHMC.jl/issues/218#issuecomment-703066979:

```julia
using AdvancedHMC, Distributions, ForwardDiff, Plots

D = 10; initial_θ = rand(D)
ℓπ(θ) = logpdf(MvNormal(zeros(D), ones(D)), θ)
n_samples, n_adapts = 2_000, 1_000

hamiltonian = Hamiltonian(metric, ℓπ, ForwardDiff)

# Leapfrog with 10% Jitter
integrator = JitteredLeapfrog(2.0, 0.1)

proposal = StaticTrajectory(integrator, 4)
adaptor = StanHMCAdaptor(MassMatrixAdaptor(metric), StepSizeAdaptor(0.8, integrator))

samples, stats = sample(hamiltonian, proposal, initial_θ, n_samples, adaptor, n_adapts; progress=true)

# Plot adaptation of stepsize 
Plots.plot([stat[:step_size] for stat in stats])
```
step_size

```julia
Plots.plot([stat[:nom_step_size] for stat in stats])
```
nom_step_size

### DIFF
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -172,7 +172,7 @@ end
 
 nom_step_size(lf::JitteredLeapfrog) = lf.ϵ0
 
-update_nom_step_size(lf::JitteredLeapfrog, ϵ0) = reconstruct(lf, ϵ0=ϵ0, ϵ=ϵ0)
+update_nom_step_size(lf::JitteredLeapfrog, ϵ0) = reconstruct(lf, ϵ0=ϵ0)
 
 # Jitter step size; ref: https://github.com/stan-dev/stan/blob/1bb054027b01326e66ec610e95ef9b2a60aa6bec/src/stan/mcmc/hmc/base_hmc.hpp#L177-L178
 function _jitter(
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -38,12 +38,22 @@ Get the current integration step size.
 """
 function step_size end
 
+"""
+    update_nom_step_size(i::AbstractIntegrator, ϵ) -> AbstractIntegrator
+
+Return a copy of the integrator `i` with the new nominal step size ([`nom_step_size`](@ref))
+`ϵ`.
+"""
+function update_nom_step_size end
+
 abstract type AbstractLeapfrog{T} <: AbstractIntegrator end
 
 step_size(lf::AbstractLeapfrog) = lf.ϵ
 jitter(::Union{AbstractRNG, AbstractVector{<:AbstractRNG}}, lf::AbstractLeapfrog) = lf
 temper(lf::AbstractLeapfrog, r, ::NamedTuple{(:i, :is_half),<:Tuple{Integer,Bool}}, ::Int) = r
 stat(lf::AbstractLeapfrog) = (step_size=step_size(lf), nom_step_size=nom_step_size(lf))
+
+update_nom_step_size(lf::AbstractLeapfrog, ϵ) = reconstruct(lf, ϵ=ϵ)
 
 function step(
     lf::AbstractLeapfrog{T},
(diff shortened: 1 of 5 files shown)


src/integrator.jl
```diff
@@ -162,6 +172,8 @@ end
 
 nom_step_size(lf::JitteredLeapfrog) = lf.ϵ0
 
+update_nom_step_size(lf::JitteredLeapfrog, ϵ0) = reconstruct(lf, ϵ0=ϵ0, ϵ=ϵ0)
```
sethaxen, line 175: Updating `ϵ` seems like the right thing to do, but it seems a little suspicious to me that if we don't update it, then it appears to never be updated by `jitter`,  so adaptation drives `ϵ0` to 0.

sethaxen:
I'd appreciate any suggestions for the right place/way to add an integration test that adaptation is updating the nominal step size sensibly.

sethaxen:
Hm failures seem to be unrelated to this PR.

src/integrator.jl
xukai92: I feel the right thing is to update `ϵ0` as you do here. `ϵ` can probably be anything because it will be override in `jitter`.

xukai92 (dismissed):
Thanks for the PR. It looks good to me. I will wait for the CI to pass.

src/integrator.jl
sethaxen: > `ϵ` can probably be anything because it will be override in `jitter`.

That's what I expected, but not updating `ϵ` results in `ϵ0` being adapted to 0, which indicates that `ϵ` is being used for adaptation somewhere before `jitter` is called.

sethaxen: I'll see if I can run down why this is happening.

sethaxen: Okay, figured it out. `transition` for `StaticTrajectory` was creating a jittered integrator but only used it for reading stats, not for sampling (yikes).

Red-Portal: @sethaxen Wow. I actually worked with that function in my personal fork and never realized it. Thank god you picked that up.

sethaxen:
Blocked by #221

xukai92:
I have #221 merged now. Can you get the commit into this PR and let the CI run? Thanks!

src/integrator.jl
xukai92: I see. This is a tricky bug. Can we have it fixed in this PR as well?

Also FYI, in my PR for unifying interfaces for different trajectories (#214), I unify the jitter (and also momentum refreshment) [here](https://github.com/TuringLang/AdvancedHMC.jl/blob/kx/unify-trajectory-1/src/trajectory.jl#L220) so that this bug is kind of being dodged. Still waiting for Hong to take a final pass of the PR to get it merged though.

sethaxen: > I see. This is a tricky bug. Can we have it fixed in this PR as well?

Yes, I fixed it in 4d09a4b

Ah okay, after this is merged I will review that as well to be sure I understand what it does.

xukai92: Thanks! The fix looks correct to me.

> Ah okay, after this is merged I will review that as well to be sure I understand what it does.

Great. That will be helpful!

xukai92 (approved):

xukai92:
I have it merged given Action CI has passed. 
I don't know what's going on for Travis CI though.

sethaxen:
For Travis, it's still trying to run one CI on Julia 1.0. But one of the test dependencies is restricted to Julia versions greater than at least 1.3 (I didn't track down which, just that it was happening).

sethaxen:
Any objections to me pushing a patch version bump to master and tagging a new release?

xukai92:
We already have a version bump in this PR. Let me tag a new release.

PR #225, opened 2020-10-27
Update IMP.hmc description in readme

### LINKED ISSUE #223: Modify link-out to IMP.hmc
I just notice that #164 updated the README with a link to [IMP.hmc](https://github.com/salilab/hmc), designated "Python interface". IMP.hmc does indeed use Python to interface between AdvancedHMC and something else, but that something else is a large Python/C++ package for protein structure determination. It does not make AdvancedHMC available to generic Python users. It's also a highly experimental package.

Personally, I think a Python interface would be rad (and would make IMP.hmc much simpler), but it's misleading to call IMP.hmc that.

sethaxen:
Closes #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ AdvancedHMC.jl provides a robust, modular and efficient implementation of advanc
 If you are interested in using AdvancedHMC.jl through a probabilistic programming language, please check it out!
 
 **Interfaces**
-- [Python interface](https://github.com/salilab/hmc) for AdvancedHMC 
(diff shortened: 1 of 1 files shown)


yebai:
Thanks, @sethaxen for fixing the description. Any plan to seprate the Python code interfacing AdvancedHMC into a lightweight standalone package?

yebai (approved):

sethaxen:
> Thanks, @sethaxen for fixing the description. Any plan to seprate the Python code interfacing AdvancedHMC into a lightweight standalone package?

Not at this time, though if I find myself doing this again in another Python package, then I'll probably create a simple wrapper. I don't think it'd be hard, I would just need to work around #227 and come up with an API where the user never needs to specify type parameters, but that's straightforward.

Issue #140, opened 2019-12-05
Vectorize NUTS


sethaxen:
It would be nice to have a vectorized/GPU-compatible implementation of NUTS to follow up the excellent work in #117. That PR indicates a main limitation is how to handle different depths across chains.

NumPyro and TensorFlow Probaility (used by PyMC4) have addressed this problem by developing an iterative NUTS approach. You can see their write-ups at https://github.com/pyro-ppl/numpyro/wiki/Iterative-NUTS and https://github.com/tensorflow/probability/blob/master/discussion/technical_note_on_unrolled_nuts.md.

While I haven't benchmarked TFP, there's supposedly not much of a runtime difference between running a single chain and hundreds of chains, at least for simple models. See [this blog post](https://colindcarroll.com/2019/08/18/very-parallel-mcmc-sampling/) for examples.

sethaxen:
@junpenglao shared with me this related Stan discourse: https://discourse.mc-stan.org/t/parallel-dynamic-hmc-merits/10895/

sethaxen:
I plan to implement this in three consecutive PRs, each with their own release.
1. Reimplement NUTS to be iterative. The implementation should be equivalent to the recursive implementation, so that passing the same RNG to both versions produces identical draws and stats. It should also be at least as performant as recursive NUTS and also as readable. If these three criteria are met, then it can replace the current recursive NUTS implementation. Performing these checks will be a lot easier if there's an official benchmark for this package.
2. Vectorize the iterative NUTS implementation. This is "vectorize" in the sense of "use broadcasting everywhere". This will be patterned off the existing Static HMC vectorization. However, I'll also profile using KernelAbstractions. In the former case, this will require updating the `BinaryTree`, etc representations to represent a tree across all chains, and handling control flow across a vector of directions (left/right).
3. Ensure the vectorized NUTS implementation works well on the GPU. This will need to come after @xukai92's fixes for Static HMC sampling on the GPU.

PR #228, opened 2020-10-30
[WIP] Implement an iterative NUTS algorithm

sethaxen:
As discussed in #140, the recursion in the `build_tree` function used for NUTS algorithm is problematic for vectorization. This PR introduces an iterative `build_tree` function that is equivalent to the recursive one.

To facilitate comparison while working on this PR, I have added a type parameter to the signature of `NUTS` to indicate whether to use recursion or iteration. This is used for benchmarking and regression tests. My intention if this is accepted is to revert that commit and delete the recursive `build_tree`, making iterative NUTS the default.

In preliminary benchmarks, the iterative and recursive versions are similar in runtime, allocations, and memory usage. However, for some target distributions, one may be slightly faster than the other. As this gets further along, I'll add more rigorous benchmarks. However, all regression tests currently pass, and reviews are welcome.

### DIFF
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -682,6 +682,28 @@ function isterminated(h::Hamiltonian, t::BinaryTree{T}, args...) where {T<:Union
     return isterminated(h, t)
 end
 
+struct TreeState{T,S,TC}
+    tree::T
+    sampler::S
+    termination::TC
+end
+
+@inline isterminated(state::TreeState) = isterminated(state.termination)
+
+function combine(rng::AbstractRNG, h::Hamiltonian, state::TreeState, state′::TreeState, v::Integer)
+    # TODO: vectorize this branch
+    if v == -1
+        treeleft, treeright = state′.tree, state.tree
+    else
+        treeleft, treeright = state.tree, state′.tree
+    end
+    tree′ = combine(treeleft, treeright)
+    sampler′ = combine(rng, state.sampler, state′.sampler)
+    termination′ = state.termination * state′.termination * isterminated(h, tree′, treeleft, treeright)
+    state′ = TreeState(tree′, sampler′, termination′)
+    return state′
+end
+
 """
 Recursivly build a tree for a given depth `j`.
 """
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -854,6 +854,7 @@ function transition(
     while !isterminated(termination) && j < τ.max_depth
         # Sample a direction; `-1` means left and `1` means right
         v = rand(rng, [-1, 1])
+        # TODO: vectorize this branch
         if v == -1
             # Create a tree with depth `j` on the left
             tree′, sampler′, termination′ = build_tree(rng, τ, h, tree.zleft, sampler, v, j, H0)
@@ -864,6 +865,7 @@ function transition(
             treeleft, treeright = tree, tree′
         end
         # Perform a MH step and increse depth if not terminated
+        # TODO: vectorize this branch
         if !isterminated(termination′)
             j = j + 1   # increment tree depth
             if mh_accept(rng, sampler, sampler′)
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -780,6 +780,61 @@ function build_tree(
     end
 end
 
+"""
+Iteratively build a tree for a given depth `j`.
+"""
+function build_tree(
+    rng::AbstractRNG,
+    nt::IterativeNUTS{S,C,I,F},
+    h::Hamiltonian,
+    z::PhasePoint,
+    sampler::AbstractTrajectorySampler,
+    v::Int,
+    j::Int,
+    H0::AbstractFloat,
+) where {I<:AbstractIntegrator,F<:AbstractFloat,S<:AbstractTrajectorySampler,C<:AbstractTerminationCriterion}
+    ileaf_max = 2^j
+    state_cache_size = j - 1
+
+    if j == 0
+        z′, state′ = build_one_leaf_tree(nt, h, z, sampler, v, H0)
+        return state′.tree, state′.sampler, state′.termination
+    end
+    z′, state′ = build_two_leaf_tree(rng, nt, h, z, sampler, v, H0)
+
+    if j == 1 || isterminated(state′)
+        return state′.tree, state′.sampler, state′.termination
+    end
+
+    # TODO: allocate state_cache in `transition` and reuse for all subtrees
+    state_cache = Vector{typeof(state′)}(undef, state_cache_size)
+    @inbounds state_cache[1] = state′
+    for ileaf in 4:2:ileaf_max
+        # take two steps (build tree with two leaves)
+        z′, state′ = build_two_leaf_tree(rng, nt, h, z′, state′.sampler, v, H0)
+
+        # combine with cached subtrees with same number of leaves
+        combine_range = subtree_cache_combine_range(ileaf)
+        for i in combine_range
+            @inbounds state′ = combine(rng, h, state_cache[i], state′, v)
+        end
+        icache = last(combine_range)
+
+        # if we have terminated, combine even if not same number of leaves
+        # TODO: vectorize this branch
+        if isterminated(state′)
+            for i in (icache - 1):-1:1
+                @inbounds state′ = combine(rng, h, state_cache[i], state′, v)
+            end
+            break
+        end
+
+        # cache state for future checks
+        @inbounds state_cache[icache] = state′
+    end
+    return state′.tree, state′.sampler, state′.termination
+end
+
 function transition(
     rng::AbstractRNG,
     τ::NUTS{S,C,I,F},
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -726,6 +726,18 @@ function build_two_leaf_tree(rng, nt::NUTS, h, z, sampler, v, H0)
 end
 
 """
+    subtree_cache_combine_range(ileaf::Integer) -> UnitRange
+
+Get the indices of the cached trees that need to be combined with the two-leaf tree produced
+at even leaf number `ileaf`.
+"""
+@inline function subtree_cache_combine_range(ileaf)
+    imin = count_ones(ileaf)
+    inum = trailing_zeros(ileaf)
+    return (imin + inum - 2):-1:imin
+end
+
+"""
 Recursivly build a tree for a given depth `j`.
 """
 function build_tree(
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -435,6 +435,9 @@ combine(::ClassicNoUTurn, ::ClassicNoUTurn) = ClassicNoUTurn()
 combine(cleft::T, cright::T) where {T<:GeneralisedNoUTurn} = T(cleft.rho + cright.rho)
 combine(cleft::T, cright::T) where {T<:StrictGeneralisedNoUTurn} = T(cleft.rho + cright.rho)
 
+abstract type TreeBuildingAlgorithm end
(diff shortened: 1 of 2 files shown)


sethaxen:
Alternatively we could keep both the iterative and recursive NUTS implementations and only use the iterative one for vectorized NUTS (dispatching on matrix of positions). The advantage of this is we can always check consistency of iterative and recursive NUTS by passing either a vector or positions or a n by 1 matrix of positions.

src/trajectory.jl
```diff
@@ -718,6 +770,55 @@ function build_tree(
...
+    H0::AbstractFloat,
+) where {I<:AbstractIntegrator,F<:AbstractFloat,S<:AbstractTrajectorySampler,C<:AbstractTerminationCriterion}
+    ileaf_max = 2^j
+    state_cache_size = j
+
+    z′, state′ = build_one_leaf_tree(nt, h, z, sampler, v, H0)
+    j == 0 && return state′.tree, state′.sampler, state′.termination
+
+    # TODO: allocate state_cache in `transition` and reuse for all subtrees
+    state_cache = Vector{typeof(state′)}(undef, state_cache_size)
```
xukai92, line 793: Just to check my understanding. The reason we have the `-1` in L727 and L815 is because we are using the `state_cache` in a binary format, thus `state_cache[end]` is the least significant digit.

xukai92:
I agree we should keep both for a while. 
As you said, doing consistency check on more targets can give us confidence on both implementations.

src/trajectory.jl
sethaxen: The reverse iteration is because we combine a subtree with the most recent subtree of the same size. And that produces a subtree that must be combined with the next most recently merged subtree, which will always be the previous entry in the cache. `subtree_cache_combine_range` isn't very intuitive, but it gives us this range, as checked against a simplified recursion. I'll add a suitable test.

test/trajectory.jl
```diff
@@ -281,4 +283,82 @@ end
             end
         end
     end
-end
\ No newline at end of file
+end
+
+@testset "iterative/recursive NUTS regression" begin
```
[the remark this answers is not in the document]
sethaxen: Sure!

sethaxen:
Okay, NUTS is completely vectorized, and for a single chain, it produces the same output for recursive and iterative NUTS. I've tested with 100 parallel chains with mvnormal and mvcauchy distributions, and the chains all seem to target the right distributions. What's left now is lots of unit and integration tests of individual functions and trying to eliminate redundant function calls. Along the way I'll try to add better tests for recursive NUTS. Since this refactor ended up producing more atomic functions, this should be easier.

I did need to make one functional change to recursive NUTS for consistency with iterative. Namely, if a proposed subtree has terminated, it still draws a random number for the metropolis-hastings step, but it still rejects the tree. This is necessary because vectorizing iterative NUTS requires drawing the random number for all chains even if some have already terminated.

One property of vectorized NUTS is that if the target distribution is heavy-tailed, then there's a high chance one of the chains ends up in the tails, and since all chains are synchronized, that means that all chains must wait for that one to terminate before continuing, which destroys the performance. (this property is known for TFP's vectorized NUTS, see https://github.com/tensorflow/probability/issues/728#issuecomment-573651590. Hence, in practice a user might switch to multithreading or parallel processing when they find that the distribution is really slow to sample.

xukai92:
> I did need to make one functional change to recursive NUTS for consistency with iterative. Namely, if a proposed subtree has terminated, it still draws a random number for the metropolis-hastings step, but it still rejects the tree. This is necessary because vectorizing iterative NUTS requires drawing the random number for all chains even if some have already terminated.

Sounds fine.

I will take a closer look at the PR this week.

src/utilities.jl
```diff
@@ -85,3 +85,46 @@ function randcat(
...
+end
+
+@inline accept!(x, x′, is_accept::Bool) = ifelse(is_accept, x′, x)
+@inline accept!(x, x′, is_accept) = eitheror!(x′, x, is_accept)
+@inline function accept!(x::AbstractMatrix, x′::AbstractMatrix, is_accept::AbstractVector)
+    return eitheror!(x′, x, is_accept')
+end
+
+@inline colwise_dot(x::AbstractVector, y::AbstractVector) = dot(x, y)
+# TODO: this function needs a custom GPU kernel, e.g. using KernelAbstractions
```
xukai92, line 122: What does this comment mean?

src/trajectory.jl
```diff
@@ -110,54 +110,94 @@ Ref: https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/hmc/nuts/base_n
...
+end
+
+function combine(rng::AbstractRNG, s1::MultinomialTS{FT,FT}, s2::MultinomialTS{FT,FT}) where {FT<:AbstractFloat}
     ℓw = logaddexp(s1.ℓw, s2.ℓw)
     zcand = rand(rng) < exp(s1.ℓw - ℓw) ? s1.zcand : s2.zcand
     return MultinomialTS(zcand, ℓw)
 end
+function combine(rng::AbstractRNG, s1::MultinomialTS, s2::MultinomialTS)
+    ℓw = logaddexp.(s1.ℓw, s2.ℓw)
+    is_accept = rand.(rng) .< exp.(s1.ℓw .- ℓw)
```
xukai92, line 176: Why do we have the broadcast `rand` for `rng` here even we have `rng:: AbstractRNG`?

src/trajectory.jl
```diff
@@ -664,115 +721,311 @@ function check_right_subtree(
...
+    is_accept = isone.(v)
+    tree_right = accept!(tree, tree_right, is_accept)
+    tree_left = accept!(tree′, tree_left, is_accept)
+    return tree_left, tree_right
+end
+
 """
 Recursivly build a tree for a given depth `j`.
 """
 function build_tree(
```
xukai92, line 847: I realise after doing this refactoring to the existing NUTS, we cannot check if the new implementation gives the same results as the old one. Any idea on doing this?

xukai92 (commented):
Thanks for the work on vectorised operations. One thing I notice is that it current doesn't support passing a vector of RNGs. Making such interface work can also help testing: we can pass a vector of same RNG and we should expect all the chains being identical.

sethaxen:
Thanks @xukai92 for the review. I got busy and have had to put this on hold. I also think the PR might need to be broken into a few more modest PRs.

One reason I paused this was because in informal benchmarks, batched iterative NUTS for N chains was universally slower than running N chains in serial. I saw this happening for two reasons, which I'll document here before I forget:
1. the abstraction of storing left and right leaves is challenging for batching, because the control flow for picking the next leaf to use as a starting point requires a copy. These copies dominated the runtime. We could avoid this control flow by instead storing the front and back leaves, where a subtree is always expanded in the forward direction. Then it's only when beginning a new tree doubling or checking whether to accept a doubling that these copies are needed. This would require a refactor to the subtrees.
2. When batching NUTS, tree doublings occur until all chains have terminated. As the number of chains N increases, the probability of any one chain hitting some depth also increases. This is especially bad when the posterior has heavy tails. I thought there would be some trade-off, like for a small number of chains, batched iterative NUTS would be faster, but I did not see this in practice, perhaps because of issue (1). Certainly it seems to be the case that parallelizing chains on a GPU would be a bad idea.

I'd like to revisit this in more detail in the future. For example, perhaps a different termination criterion that scales better could be used (e.g. something between terminate all chains when any chain has terminated and terminate only when all chains have terminated), but we must be careful to ensure that such a criterion targets the same distribution and does so efficiently. I'm sure @junpenglao has had similar thoughts RE TFP's batched iterative NUTS implementation.

junpenglao:
Drive by comments on:
> 1. the abstraction of storing left and right leaves is challenging for batching, because the control flow for picking the next leaf to use as a starting point requires a copy. These copies dominated the runtime. We could avoid this control flow by instead storing the front and back leaves, where a subtree is always expanded in the forward direction. Then it's only when beginning a new tree doubling or checking whether to accept a doubling that these copies are needed. This would require a refactor to the subtrees.

I am not familiar with Julia code, but probably you could avoid copying by doing [some](https://github.com/tensorflow/probability/blob/12954f1bae6b15f70d318fb9bed9f4788bf0cfad/tensorflow_probability/python/mcmc/nuts.py#L545-L549) [indexing](https://github.com/tensorflow/probability/blob/12954f1bae6b15f70d318fb9bed9f4788bf0cfad/tensorflow_probability/python/mcmc/nuts.py#L558-L561) and [update of tensor](https://github.com/tensorflow/probability/blob/12954f1bae6b15f70d318fb9bed9f4788bf0cfad/tensorflow_probability/python/mcmc/nuts.py#L662-L673)

sethaxen:
> I am not familiar with Julia code, but probably you could avoid copying by doing [some](https://github.com/tensorflow/probability/blob/12954f1bae6b15f70d318fb9bed9f4788bf0cfad/tensorflow_probability/python/mcmc/nuts.py#L545-L549) [indexing](https://github.com/tensorflow/probability/blob/12954f1bae6b15f70d318fb9bed9f4788bf0cfad/tensorflow_probability/python/mcmc/nuts.py#L558-L561) and [update of tensor](https://github.com/tensorflow/probability/blob/12954f1bae6b15f70d318fb9bed9f4788bf0cfad/tensorflow_probability/python/mcmc/nuts.py#L662-L673).

It's true, there may be other refactors that can avoid the copy. This was just an obvious one. Thanks for the links. I'm not familiar with TensorFlow code but will process when I work on this again.

> Also, I am not sure if it is still valid to only extend subtree in one direction, from Betancount's HMC paper: "In order to ensure a valid correction, this transition from states to trajectories has to exhibit a certain reversibility. Formally we require that the probability of transitions to a trajectory is the same regardless of from which state in that trajectory we might begin" (page 49 of https://arxiv.org/pdf/1701.02434.pdf)

Perhaps we're talking about different things? The NUTS algorithm proceeds by tree-doubling. Each time the tree is doubled, a Bernoulli random variable determines whether we double on the left or the right. Let's call our original tree A and the new tree B. Tree B is constructed in the same way as A (i.e. via tree doubling with all of the same u-turn checks before accepting new subtrees), except that it is only extended in one direction (else it could overlap A). Otherwise tree A and B could overlap. Informally, as long as tree B does not encounter a termination condition internally or when merged with A, then given the merged tree A+B, it is not possible to determine whether tree A or B was the original tree. Due to the recursive nature, of the tree doubling, we cannot determine which point on the trajectory was the starting point, so we have reversibility. The extension of B in one direction to balance the tree is what I am referring to.

junpenglao:
> The extension of B in one direction to balance the tree is what I am referring to.

Maybe it is more implementation detail in your PR then - In isolation, B is always in one direction, as you select the starting point (left or right of A) and direction of the momentum (-1 if left, 1 if right). After B is complete, you added that to A (`reverse(B)+A` or `A+B`, `+` is concatenation conceptually). So IIUC you are saying you will only do `A+B`, which I worry it might invalid the termination criteria.

sethaxen:
> After B is complete, you added that to A (`reverse(B)+A` or `A+B`, `+` is concatenation conceptually). So IIUC you are saying you will only do `A+B`, which I worry it might invalid the termination criteria.

Ah, no, that is not what I am saying. `reverse(B)+A` or `A+B` would be the operations, implicitly.

xukai92:
> Thanks @xukai92 for the review. I got busy and have had to put this on hold. I also think the PR might need to be broken into a few more modest PRs.

No worries. Maybe we can start with a PR that only updating the interfaces/abstractions for the recursive NUTS without modifying any functionality? We can make sure that everything stays the same with the new interfaces. Then pushing another PR to add the vectorised version would be more smooth.

PR #240, opened 2021-01-08
Refactoring termination criterion

xukai92:
This PR extracts the codes on refactoring termination criteria abstraction from #214. 
There changes are needed for unifying static and dynamic trajectories.

Changes are
1. Introduced `AbstractTerminationCriterion`, `StaticTerminationCriterion <: AbstractTerminationCriterion` and `DynamicTerminationCriterion <: AbstractTerminationCriterion `
2. Introduced `FixedNSteps` and `FixedIntegrationTime`
3. Refactored `GeneralisedNoUTurn` and `StrictGeneralisedNoUTurn` so that they only contains algorithm parameters but not turn statistics
4. Introduced `TurnStatistics` as suggested in https://github.com/TuringLang/AdvancedHMC.jl/pull/214/files#r467505568

Along with necessary updates in test cases and docs (+some clean-ups).


xukai92:
@sethaxen Can you please take a look at this PR? 
I guess the `TurnStatistic ` is also what you wanted to introduce while doing the vec-NUTS PR IRCC.
Does what I doing here suit your usage?

yebai:
I'll take a look at this PR soon.

src/trajectory.jl
```diff
@@ -27,30 +25,55 @@ end
...
+Standard HMC implementation with a fixed integration time.
+
+# Fields
+$(TYPEDFIELDS)
+
+# References
+1. Neal, R. M. (2011). MCMC using Hamiltonian dynamics. Handbook of Markov chain Monte Carlo, 2(11), 2. ([arXiv](https://arxiv.org/pdf/1206.1901)) 
 """
+struct FixedIntegrationTime{F<:AbstractFloat} <: StaticTerminationCriterion
+    "Total length of the trajectory, i.e. take `floor(λ / integrator_step)` number of leapfrog steps."
```
yebai, line 62:
```suggestion
    "Total length of the trajectory, i.e. take `floor(λ / integrator_step_size)` number of leapfrog steps."
```

src/trajectory.jl
```diff
@@ -177,32 +200,31 @@ end
...
 
 # References
 1. Neal, R. M. (2011). MCMC using Hamiltonian dynamics. Handbook of Markov chain Monte Carlo, 2(11), 2. ([arXiv](https://arxiv.org/pdf/1206.1901))
 """
-struct StaticTrajectory{S<:AbstractTrajectorySampler, I<:AbstractIntegrator} <: AbstractTrajectory{I}
-    "Integrator used to simulate trajectory."
-    integrator  ::  I
-    "Number of steps to simulate, i.e. length of trajectory will be `n_steps + 1`."
-    n_steps     ::  Int
+struct StaticTrajectory{S<:AbstractTrajectorySampler, I<:AbstractIntegrator, TC<:StaticTerminationCriterion} <: AbstractTrajectory{I}
```
yebai, line 211: Maybe 
```suggestion
struct StaticTrajectory{
    S<:AbstractTrajectorySampler, 
    I<:AbstractIntegrator, 
    TC<:StaticTerminationCriterion
    } <: AbstractTrajectory{I}
```

src/trajectory.jl
```diff
@@ -326,31 +348,30 @@ abstract type DynamicTrajectory{I<:AbstractIntegrator} <: AbstractTrajectory{I}
...
 
 # References
 1. Neal, R. M. (2011). MCMC using Hamiltonian dynamics. Handbook of Markov chain Monte Carlo, 2(11), 2. ([arXiv](https://arxiv.org/pdf/1206.1901)) 
 """
-struct HMCDA{S<:AbstractTrajectorySampler,I<:AbstractIntegrator} <: DynamicTrajectory{I}
-    "Integrator used to simulate trajectory."
-    integrator  ::  I
-    "Total length of the trajectory, i.e. take `floor(λ / integrator_step)` number of leapfrog steps."
-    λ           ::  AbstractFloat
+struct HMCDA{S<:AbstractTrajectorySampler, I<:AbstractIntegrator, TC<:StaticTerminationCriterion} <: DynamicTrajectory{I}
```
yebai, line 359: Maybe `HMCDA` should be a subtype of `StaticTrajectory{I}`?

src/trajectory.jl
```diff
@@ -373,98 +394,102 @@ end
...
+# Fields
+$(TYPEDFIELDS)
+
 # References
 1. Hoffman, M. D., & Gelman, A. (2014). The No-U-Turn Sampler: adaptively setting path lengths in Hamiltonian Monte Carlo. Journal of Machine Learning Research, 15(1), 1593-1623. ([arXiv](http://arxiv.org/abs/1111.4246))
 """
-struct ClassicNoUTurn <: AbstractTerminationCriterion end
-
-ClassicNoUTurn(::PhasePoint) = ClassicNoUTurn()
+Base.@kwdef struct ClassicNoUTurn{F<:AbstractFloat} <: DynamicTerminationCriterion
```
yebai, line 411: nice!

src/trajectory.jl
```diff
@@ -373,98 +394,102 @@ end
...
+combine(tsl::T, tsr::T) where {T<:TurnStatistic} = TurnStatistic(tsl.rho + tsr.rho)
 
 ##
 ## NUTS
 ##
 
 """
-Dynamic trajectory HMC using the no-U-turn termination criteria algorithm.
+$(TYPEDEF)
+Dynamic HMC using the no-U-turn termination criteria algorithm.
```
yebai, line 468:
```suggestion
Dynamic HMC algorithm using the no-U-turn termination criteria.
```

src/trajectory.jl
```diff
@@ -557,31 +573,29 @@ Return the value with the largest absolute value.
...
-    isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
-
+$(SIGNATURES)
 Detect U turn for two phase points (`zleft` and `zright`) under given Hamiltonian `h`
 using the (original) no-U-turn cirterion.
 
 Ref: https://arxiv.org/abs/1111.4246, https://arxiv.org/abs/1701.02434
 """
-function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
+function isterminated(::ClassicNoUTurn, h::Hamiltonian, t::BinaryTree)
```
yebai, line 598: Any reason for replace parametric `BinaryTree{AbstractTerminationCriterion}` with `isterminated(::ClassicNoUTurn, h::Hamiltonian, t::BinaryTree)`? Dispatching on parametric typed implementation is probably more natural than passing an additional type parameter..

src/trajectory.jl
```diff
@@ -590,42 +604,30 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
-
+$(SIGNATURES)
 Detect U turn for two phase points (`zleft` and `zright`) under given Hamiltonian `h`
 using the generalised no-U-turn criterion.
 
 Ref: https://arxiv.org/abs/1701.02434
 """
-function isterminated(h::Hamiltonian, t::BinaryTree{<:GeneralisedNoUTurn})
-    rho = t.c.rho
+function isterminated(::GeneralisedNoUTurn, h::Hamiltonian, t::BinaryTree)
```
yebai, line 613: Similar here - parametric type-based dispatching is more Julian..

src/trajectory.jl
```diff
@@ -590,42 +604,30 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
-    t_generalised = BinaryTree(
-        t.zleft,
-        t.zright,
-        GeneralisedNoUTurn(t.c.rho),
-        t.sum_α,
-        t.nα,
-        t.ΔH_max
-    )
-    s1 = isterminated(h, t_generalised)
+    tc::StrictGeneralisedNoUTurn, h::Hamiltonian, t, tleft, tright
```
yebai, line 627: Similar here - parametric type-based dispatching is more Julian..

src/trajectory.jl
```diff
@@ -373,98 +394,102 @@ end
...
 2. [https://github.com/stan-dev/stan/pull/2800](https://github.com/stan-dev/stan/pull/2800)
 """
-struct StrictGeneralisedNoUTurn{T<:AbstractVector{<:Real}} <: AbstractTerminationCriterion
+Base.@kwdef struct StrictGeneralisedNoUTurn{F<:AbstractFloat} <: DynamicTerminationCriterion
+    max_depth::Int=10
+    Δ_max::F=1000.0
+end
+GeneralisedNoUTurn(tc::StrictGeneralisedNoUTurn) = GeneralisedNoUTurn(tc.max_depth, tc.Δ_max)
+
+struct TurnStatistic{T}
```
yebai, line 449: Maybe considering
```suggestion
struct TurnStatistic{T, C<:AbstractTerminationCriterion}
```
which also removes the need to pass explicit termination criterion types in functions `isterminated(::ClassicNoUTurn, h::Hamiltonian, t::BinaryTree)` below.

src/trajectory.jl
```diff
@@ -671,31 +667,29 @@ function generalised_uturn_criterion(rho, p_sharp_minus, p_sharp_plus)
     return (dot(rho, p_sharp_minus) <= 0) || (dot(rho, p_sharp_plus) <= 0)
 end
 
-function isterminated(h::Hamiltonian, t::BinaryTree{T}, args...) where {T<:Union{ClassicNoUTurn, GeneralisedNoUTurn}}
-    return isterminated(h, t)
+function isterminated(tc::TC, h::Hamiltonian, t::BinaryTree, _tleft, _tright) where {TC<:Union{ClassicNoUTurn, GeneralisedNoUTurn}}
```
yebai, line 670: Similar here - `tc` is a filed of `t::BinaryTree` plus parametric type-based dispatching is more Julian..

yebai (changes requested):
thanks, @xukai92 - it looks very good overall. I left a few minor comments below.

src/trajectory.jl
xukai92: `StaticTrajectory` is not yet an abstract type in this PR. This will be updated when unifying the types, which is planned after this PR.

src/trajectory.jl
xukai92: As I was not putting any type infomation about U-turn into turn stats.
By doing what you suggest on `TurnStatistic`, this problem can be resolve.

src/trajectory.jl
```diff
@@ -590,42 +608,30 @@ function isterminated(h::Hamiltonian, t::BinaryTree{<:ClassicNoUTurn})
...
-        GeneralisedNoUTurn(t.c.rho),
-        t.sum_α,
-        t.nα,
-        t.ΔH_max
-    )
-    s1 = isterminated(h, t_generalised)
+    tc::StrictGeneralisedNoUTurn, h::Hamiltonian, t, tleft, tright
+)
+    # (Non-strict) generalised U-turn check
+    s1 = isterminated(GeneralisedNoUTurn(tc), h, t)
```
xukai92, line 634: @yebai Something less elegent would appear here, if we were to put `tc` as a field of `BinaryTree`.
That is, we have to manually reconstruct the tree, as what we are doing master ATM: https://github.com/TuringLang/AdvancedHMC.jl/blob/master/src/trajectory.jl#L620-L627

PS, you cannot do this with reconstruct, because the type changes, by which I mean
```julia
julia> using Parameters

julia> struct A{T}
         v::T
       end

julia> a = A(1.0)
A{Float64}(1.0)

julia> reconstruct(a, v="1")
ERROR: MethodError: Cannot `convert` an object of type String to an object of type Float64
```

I guess the root reason here is that, `tc` is indeed _not_ a property of a tree but a surely a property of `NUTS`: `isterminated` should take both the tree and the criterion as input, to decide if the termination happens. For the same tree, based on different termianiton criteria, it could terminate or not.

xukai92: I hope to keep the current design if possible. I think pure multiple distapching without parametric types is also Julian :)

yebai: thanks for the clarification, I agree that the new design is more modular. However, I feel it introduces a bit more clutter as well. It would be nice to simplify a bit here. For example, we can keep the `turn_statistic` field's type generic. Using your illustration, this would look like 

```julia
julia> struct A
         v
       end

julia> a = A(1.0)
A{Float64}(1.0)

julia> reconstruct(a, v="1")

```

Then, we can modify the function `isterminated` to dispatch based on the type of `BinaryTree.turn_statistic` rather than the type of `BinaryTree`.

xukai92: Making it generic will cause type instability issues whenever `a.v` is used for dispatching functions.

For example,
```julia
julia> struct A
       v
       end

julia> f(::AbstractFloat) = 1.0

julia> f(::AbstractString) = "1.0"
f (generic function with 2 methods)

julia> g(a::A) = f(a.v)
g (generic function with 1 method)

julia> @code_warntype g(A(1.0))
Variables
  #self#::Core.Compiler.Const(g, false)
  a::A

Body::Union{Float64, String}
1 ─ %1 = Base.getproperty(a, :v)::Any
│   %2 = Main.f(%1)::Union{Float64, String}
└──      return %2
```

xukai92:
@torfjelde Any chance you have reviewed this PR?

torfjelde (approved):
LGTM! Have a couple of comments/questions but nothing breaking.

src/trajectory.jl
```diff
@@ -177,32 +200,35 @@ end
...
-    integrator  ::  I
-    "Number of steps to simulate, i.e. length of trajectory will be `n_steps + 1`."
-    n_steps     ::  Int
+struct StaticTrajectory{
+    S<:AbstractTrajectorySampler, 
+    I<:AbstractIntegrator, 
+    TC<:StaticTerminationCriterion
+    } <: AbstractTrajectory{I}
+    integrator              ::  I
+    termination_criterion   ::  TC
```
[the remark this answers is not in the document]
xukai92: I've added the one for integrator back! Thanks for pointing out.

src/trajectory.jl
```diff
@@ -296,7 +322,7 @@ function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMat
 end
 
 function sample_phasepoint(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
-    n_steps = abs(τ.n_steps)
+    n_steps = abs(τ.termination_criterion.L)
```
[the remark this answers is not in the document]
xukai92: This is supposed to be introduced in the next PR when unifying trajectories types.

xukai92 (commented):
Thanks @torfjelde!

xukai92:
@yebai I guess this PR is good to go once the CI passes.

yebai:
It seems that CI is broken due to mismatched function signatures. 

https://github.com/TuringLang/AdvancedHMC.jl/pull/240/checks?check_run_id=1866519262#step:5:385

xukai92:
Opps. I by mistake committed some temporary changes and pushed them along with the commit to adapt Tor's suggestion.
I've reverted these changes and hopefully the CI shall pass now.

sethaxen:
@xukai92 so sorry for the late reply! I'll review this in a few hours.

sethaxen:
> @sethaxen Can you please take a look at this PR?
> I guess the `TurnStatistic ` is also what you wanted to introduce while doing the vec-NUTS PR IRCC.
> Does what I doing here suit your usage?

I don't recall adding a `TurnStatistic`, and I don't see anything to that effect in #228. Haven't had a chance to go through this PR in detail, but nothing looks off.

yebai (approved):

PR #242, opened 2021-01-18
Fixed default type for NesterovDualAveraging

torfjelde:
`NesterovDualAveraging` has a default constructor which does not respect the types of the given parameters. This PR fixes this. 

Note that `δ` and the rest of the parameters now wrapped with `T` are required to have the same type, and so the current constructor simply fails if `δ` is a `Float32`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "AdvancedHMC"
 uuid = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
-version = "0.2.27"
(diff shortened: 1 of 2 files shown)


cpfiffer (approved):
Seems like a simple fix -- I'll approve but you may wish to wait for a real adult on this one. 

Could you increment the version number as well?

xukai92 (approved):

yebai:
Thanks, @torfjelde!

Issue #184, opened 2020-02-27
Sensible Defaults


willtebbutt:
While the example in the README is important in that it exposes all of the useful bits of AdvancedHMC, it would be nice if it were possible to define a couple of sensible defaults. e.g.
```julia
vanilla_hmc_sampler = HMC(step_size, n_steps)
reasonable_nuts_sampler = NUTS(step_size)
```
I think that this kind of thing would be a helpful addition to the package, it makes it more straightforward to get up and running.

In particular, it would make writing examples for `Stheno.jl`s documentation quite a lot less verbose.

scheidan:
I'd like to bump this issue. I'm planning to use `AdvancedHMC` for a teaching exercise (outside of Turing) so really need a a much simpler, user-friendly  interface.

For my own use I wrote this wrapper function below (more or less directly from the Readme). Is something like this of scope for this package? It would make it much easier for users  not using Turing to profit from your work.

`````
# NUTS sampler as used by STAN
Based on documentation of AdvancedHMC.

```Julia
stanHMC(lp::Function, ∇lp::Function,  θ_init,;
        n_samples::Int=1000, n_adapts::Int=n_samples÷2)
```

## Arguments
- `lp`:  log density to sample from (up to a constant)
- `∇lp`: function that computes gradient
- `n_samples::Int=1000`: number of samples
- `n_adapts::Int=n_samples÷2`: length of adaptation

### Return Value
A named tuple with fields:
- `samples`: array containing the samples
- `stats`: contains various statistics form the sampler. See AdvancedHMC documentation.
"""
function advancedHMC_NUTS(lp::Function, ∇lp::Function,  θ_init,;
                          n_samples::Int=1000, n_adapts::Int=n_samples÷2)

    # Define a Hamiltonian system
    D = length(θ_init)   # number of parameters
    metric = DiagEuclideanMetric(D)

    # choose AD framework or provide a function manually
    hamiltonian = Hamiltonian(metric, lp, θr -> (lp(θr), ∇lp(θr)))

    # Define a leapfrog solver, with initial step size chosen heuristically
    initial_ϵ = find_good_stepsize(hamiltonian, θ_init)
    integrator = Leapfrog(initial_ϵ)

    # Define an HMC sampler, with the following components
    #   - multinomial sampling scheme,
    #   - generalised No-U-Turn criteria, and
    #   - windowed adaption for step-size and diagonal mass matrix
[...]

PR #245, opened 2021-02-10
Unifying trajectories

xukai92:
This PR unifies all trajectories under the same interface.
For example, NUTS shall be constructed by:

```julia
τ = Trajectory{MultinomialTS}(Leapfrog(1e-3), GeneralisedNoUTurn())
```

Old constructors are supported but deprecated.

Besides, the way to unify `FixedNSteps` and `FixedIntegrationTime` is via a new-introduced `nsteps(::Trajectory)`, suggested by @torfjelde.
- Previously, `FixedIntegrationTime` was converted to `FixedNSteps` before calling `transition`, and `termination_criterion.L` appears here and there, which is not ideal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ integrator = Leapfrog(initial_ϵ)
 #   - multinomial sampling scheme,
 #   - generalised No-U-Turn criteria, and
 #   - windowed adaption for step-size and diagonal mass matrix
-proposal = Trajectory{MultinomialTS}(integrator, GeneralisedNoUTurn())
+proposal = NUTS{MultinomialTS, GeneralisedNoUTurn}(integrator)
 adaptor = StanHMCAdaptor(MassMatrixAdaptor(metric), StepSizeAdaptor(0.8, integrator))
 
 # Run the sampler to draw samples from the specified Gaussian, where
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ integrator = Leapfrog(initial_ϵ)
 #   - multinomial sampling scheme,
 #   - generalised No-U-Turn criteria, and
 #   - windowed adaption for step-size and diagonal mass matrix
-proposal = NUTS{MultinomialTS, GeneralisedNoUTurn}(integrator)
+proposal = Trajectory{MultinomialTS}(integrator, GeneralisedNoUTurn())
 adaptor = StanHMCAdaptor(MassMatrixAdaptor(metric), StepSizeAdaptor(0.8, integrator))
 
 # Run the sampler to draw samples from the specified Gaussian, where
--- a/src/AdvancedHMC.jl
+++ b/src/AdvancedHMC.jl
@@ -34,12 +34,34 @@ include("integrator.jl")
 export Leapfrog, JitteredLeapfrog, TemperedLeapfrog
 
 include("trajectory.jl")
-@deprecate find_good_eps find_good_stepsize
-export EndPointTS, SliceTS, MultinomialTS, 
-       StaticTrajectory, HMCDA, NUTS, 
-       ClassicNoUTurn, GeneralisedNoUTurn, 
-       StrictGeneralisedNoUTurn,
+export Trajectory,
+       FixedNSteps, FixedIntegrationTime,
+       ClassicNoUTurn, GeneralisedNoUTurn, StrictGeneralisedNoUTurn,
+       EndPointTS, SliceTS, MultinomialTS, 
        find_good_stepsize
+
+# Deprecations for trajectory.jl
+
+struct AbstractTrajectory <: AbstractProposal end
+
+struct StaticTrajectory{TS} end
+@deprecate StaticTrajectory{TS}(int::AbstractIntegrator, L) where {TS} Trajectory{TS}(int, FixedNSteps(L))
+@deprecate StaticTrajectory(int::AbstractIntegrator, L) Trajectory{EndPointTS}(int, FixedNSteps(L))
+@deprecate StaticTrajectory(ϵ::AbstractScalarOrVec{<:Real}, L) Trajectory{EndPointTS}(Leapfrog(ϵ), FixedNSteps(L))
+
+struct HMCDA{TS} end
+@deprecate HMCDA{TS}(int::AbstractIntegrator, λ) where {TS} Trajectory{TS}(int, FixedIntegrationTime(λ))
+@deprecate HMCDA(int::AbstractIntegrator, λ) Trajectory{MetropolisTS}(int, FixedIntegrationTime(λ))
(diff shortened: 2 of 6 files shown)


src/trajectory.jl
```diff
@@ -322,8 +308,8 @@ function randcat(rng, zs::AbstractVector{<:PhasePoint}, unnorm_ℓP::AbstractMat
     return z
 end
 
-function sample_phasepoint(rng, τ::StaticTrajectory{MultinomialTS}, h, z)
-    n_steps = abs(τ.termination_criterion.L)
+function sample_phasepoint(rng, τ::Trajectory{MultinomialTS}, h, z)
+    n_steps = abs(nsteps(τ))
```
torfjelde, line 312:
```suggestion
    n_steps = nsteps(τ)
```

torfjelde (approved):
LGTM! Just address my one comment. And, though it's not directly related to this PR, after we completely remove the NUTS struct it would still be useful to have a NUTS method which simply constructs the corresponding `Trajectory` so that people who aren't too familiar with the topic find what they expect.

Also, I really like this work:)

src/AdvancedHMC.jl
```diff
@@ -34,13 +34,35 @@ include("integrator.jl")
...
+@deprecate StaticTrajectory{TS}(int::AbstractIntegrator, L) where {TS} Trajectory{TS}(int, FixedNSteps(L))
+@deprecate StaticTrajectory(int::AbstractIntegrator, L) Trajectory{EndPointTS}(int, FixedNSteps(L))
+@deprecate StaticTrajectory(ϵ::AbstractScalarOrVec{<:Real}, L) Trajectory{EndPointTS}(Leapfrog(ϵ), FixedNSteps(L))
+
+struct HMCDA{TS} end
+@deprecate HMCDA{TS}(int::AbstractIntegrator, λ) where {TS} Trajectory{TS}(int, FixedIntegrationTime(λ))
+@deprecate HMCDA(int::AbstractIntegrator, λ) Trajectory{MetropolisTS}(int, FixedIntegrationTime(λ))
+@deprecate HMCDA(ϵ::AbstractScalarOrVec{<:Real}, λ) Trajectory{MetropolisTS}(Leapfrog(ϵ), FixedIntegrationTime(λ))
+
+struct NUTS{TS, TC} end
```
xukai92, line 57: @torfjelde Is this what you are asking for?

src/trajectory.jl
xukai92: This will not support negative step sizes any more.

torfjelde: Ah of course, sorry!

src/AdvancedHMC.jl
torfjelde: No, I was thinking more like the
```julia
NUTS(int::AbstractIntegrator, args...; kwargs...) = Trajectory{MultinomialTS}(int, GeneralisedNoUTurn(args...; kwargs...))
```
below but _not_ deprecating it. You will inevitably have people coming in that expect a `NUTS` method to exist and will be confused when either
1. It doesn't exists (which should be the case later if we deprecate it, i.e. indicate that in the future versions this will no longer be implemented).
2. There's a deprecation warning.

xukai92: Cool. I will remove the deprecation then.

xukai92: I updated this. I guess this is enough for this PR but we should also add a default for `HMC` in another PR, as requested by Will here: https://github.com/TuringLang/AdvancedHMC.jl/issues/184. Are there other reasonable defaults?

xukai92:
CI passed now and it's ready for a final look @yebai

xukai92:
The only CI which "fails" due to running out of time. 
For some reaon the Geweke tests are pretty slow on Julia 1.0 (>4h) vs nightly (~1.5h).
I check previous CI logs and this was true before (this PR) as well.

xukai92:
To be safe I just triggered a re-run of all CI. Let's see how it goes this time.

src/AdvancedHMC.jl
```diff
@@ -34,13 +34,54 @@ include("integrator.jl")
...
+@deprecate StaticTrajectory(ϵ::AbstractScalarOrVec{<:Real}, L) Trajectory{EndPointTS}(Leapfrog(ϵ), FixedNSteps(L))
+
+struct HMCDA{TS} end
+@deprecate HMCDA{TS}(int::AbstractIntegrator, λ) where {TS} Trajectory{TS}(int, FixedIntegrationTime(λ))
+@deprecate HMCDA(int::AbstractIntegrator, λ) Trajectory{MetropolisTS}(int, FixedIntegrationTime(λ))
+@deprecate HMCDA(ϵ::AbstractScalarOrVec{<:Real}, λ) Trajectory{MetropolisTS}(Leapfrog(ϵ), FixedIntegrationTime(λ))
+
+@deprecate find_good_eps find_good_stepsize
+
+export AbstractTrajectory, StaticTrajectory, HMCDA, find_good_eps
```
yebai, line 83: Do we need to export `AbstractTrajectory`?

yebai (commented):
Excellent work - many thanks to Kai! I left some minor comments below.

README.md
```diff
@@ -54,7 +54,7 @@ integrator = Leapfrog(initial_ϵ)
 #   - multinomial sampling scheme,
 #   - generalised No-U-Turn criteria, and
 #   - windowed adaption for step-size and diagonal mass matrix
-proposal = NUTS{MultinomialTS, GeneralisedNoUTurn}(integrator)
+proposal = Trajectory{MultinomialTS}(integrator, GeneralisedNoUTurn())
```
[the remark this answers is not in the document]
xukai92: We still support the old interface. I guess it's actually better for us to keep the old one in the README.

src/AdvancedHMC.jl
xukai92: We don't. Let me remove this.

src/AdvancedHMC.jl
```diff
@@ -34,13 +34,54 @@ include("integrator.jl")
 export Leapfrog, JitteredLeapfrog, TemperedLeapfrog
 
 include("trajectory.jl")
-@deprecate find_good_eps find_good_stepsize
-export EndPointTS, SliceTS, MultinomialTS, 
-       StaticTrajectory, HMCDA, NUTS, 
-       ClassicNoUTurn, GeneralisedNoUTurn, 
-       StrictGeneralisedNoUTurn,
+export Trajectory,
+       FixedNSteps, FixedIntegrationTime,
```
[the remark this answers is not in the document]
xukai92: Created an issue for this and some other potential naming improvements https://github.com/TuringLang/AdvancedHMC.jl/issues/246.

xukai92:
Is this good to go?

PR #247, opened 2021-02-16
Introduce HMCKernel and momentum refreshment structs

xukai92:
We are finally here with this interface

```julia
τ = Trajectory(Leapfrog(1e-3), GeneralisedNoUTurn())
κ = HMCKernel(FullRefreshment(), τ)
```

To sum up, this PR introduces a very thin kernel wrapper over refreshment behaviour and the trajectory structs.
Internally, it as simple as
```julia
function transition(rng, h, κ, z)
    @unpack refreshment, τ = κ
    z = refresh(rng, refreshment, h, z)
    return transition(rng, τ, h, z)
end
```
As a benefit of such modularity, this is the only place now `jitter` and `refresh` is called.
This is also inline with how we unify Metropolis and multinomial HMC in our paper.

Some notes one things you may ask during review this PR:

1. Why do we need `AbstractMCMCKernel`?
  - This allows a user to specifc other kernels (RWMH, mixture) when needed.
2. Why do we still keep `AbstractTrajectory`?
  - To allow Turing compile so that CI passes
  - This will be removed after we update stuff on the Turing side

Also note that codes are not necessarily in the best location they should be.
I didn't take care of this aspect as I want to minimise code change for review purposes.

### DIFF
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -138,14 +138,14 @@ phasepoint(
     h::Hamiltonian
 ) where {T<:Real} = phasepoint(h, θ, rand(rng, h.metric))
 
-abstract type AbstractRefreshment end
+abstract type AbstractMomentumRefreshment end
 
 "Completly resample new momentum."
-struct FullRefreshment <: AbstractRefreshment end
+struct FullMomentumRefreshment <: AbstractMomentumRefreshment end
 
 refresh(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
-    ::FullRefreshment,
+    ::FullMomentumRefreshment,
     h::Hamiltonian,
     z::PhasePoint,
 ) = phasepoint(h, z.θ, rand(rng, h.metric))
@@ -157,13 +157,13 @@ Partial momentum refreshment with refresh rate `α`.
 
 1. Neal, Radford M. "MCMC using Hamiltonian dynamics." Handbook of markov chain monte carlo 2.11 (2011): 2.
 """
-struct PartialRefreshment{F<:AbstractFloat} <: AbstractRefreshment
+struct PartialMomentumRefreshment{F<:AbstractFloat} <: AbstractMomentumRefreshment
     α::F
 end
 
 refresh(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
-    ref::PartialRefreshment,
+    ref::PartialMomentumRefreshment,
     h::Hamiltonian,
     z::PhasePoint,
 ) = phasepoint(h, z.θ, ref.α * z.r + (1 - ref.α^2) * rand(rng, h.metric))
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -138,15 +138,32 @@ phasepoint(
     h::Hamiltonian
 ) where {T<:Real} = phasepoint(h, θ, rand(rng, h.metric))
 
+abstract type AbstractRefreshment end
+
+"Completly resample new momentum."
+struct FullRefreshment <: AbstractRefreshment end
+
 refresh(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
+    ::FullRefreshment,
+    h::Hamiltonian,
     z::PhasePoint,
-    h::Hamiltonian
 ) = phasepoint(h, z.θ, rand(rng, h.metric))
 
-# refresh(
-#     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
-#     z::PhasePoint,
-#     h::Hamiltonian,
-#     α::AbstractFloat
-# ) = phasepoint(h, z.θ, α * z.r + (1 - α^2) * rand(rng, h.metric))
+"""
+Partial momentum refreshment with refresh rate `α`.
+
+## References
+
+1. Neal, Radford M. "MCMC using Hamiltonian dynamics." Handbook of markov chain monte carlo 2.11 (2011): 2.
+"""
+struct PartialRefreshment{F<:AbstractFloat} <: AbstractRefreshment
+    α::F
+end
+
+refresh(
+    rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
+    ref::PartialRefreshment,
+    h::Hamiltonian,
+    z::PhasePoint,
+) = phasepoint(h, z.θ, ref.α * z.r + (1 - ref.α^2) * rand(rng, h.metric))
--- a/src/sampler.jl
+++ b/src/sampler.jl
@@ -156,7 +156,7 @@ Sample `n_samples` samples using the proposal `κ` under Hamiltonian `h`.
 function sample(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
     h::Hamiltonian,
-    κ::AbstractMCMCKernel,
+    κ::HMCKernel,
     θ::T,
     n_samples::Int,
     adaptor::AbstractAdaptor=NoAdaptation(),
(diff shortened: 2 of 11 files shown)


src/trajectory.jl
```diff
@@ -25,7 +25,7 @@ end
 "Returns the statistics for transition `t`."
 stat(t::Transition) = t.stat
 
-abstract type AbstractProposal end
+abstract type AbstractMCMCKernel end
```
torfjelde, line 28: The one annoying bit with this is that we're really moving towards reimplementing parts of AbstractMCMC.jl. Ideally we'd just work off that. 

But I guess that's work for another PR? :confused:

torfjelde (approved):
Generally looks good to me! Only have a couple of minor comments/questions.

src/trajectory.jl
xukai92: I guess it's fine for now. 
AbstractMCMC.jl ATM is mostly defining _sampler_ interface (i.e. how `sample` works by abstracting out how parallel stuff should be done, how samples should be collected, etc.) while this _kernel_ interface is for controlling how to manipulate kernels (for now the only thing we will do is making a mixture of a few).
But I agree this kernel interface should also be moved to an upsteam package ideally.

src/sampler.jl
```diff
@@ -154,7 +156,7 @@ Sample `n_samples` samples using the proposal `τ` under Hamiltonian `h`.
 function sample(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
     h::Hamiltonian,
-    τ::AbstractProposal,
+    κ::AbstractMCMCKernel,
```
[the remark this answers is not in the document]
xukai92: I think you are right here.
Even in the case we define a mixture kernel, this `sample` function will break. E.g. we would expect different stats (HMC's vs RWMH's) from different iterations, thus breaking some code here.

src/trajectory.jl
torfjelde: I agree, probably something for a later point:)

torfjelde (approved):
LGTM!

src/hamiltonian.jl
```diff
@@ -138,15 +138,32 @@ phasepoint(
...
+
 refresh(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
+    ::FullRefreshment,
+    h::Hamiltonian,
     z::PhasePoint,
-    h::Hamiltonian
 ) = phasepoint(h, z.θ, rand(rng, h.metric))
 
-# refresh(
```
xukai92, line 147: @yebai This was not enabled before.

src/sampler.jl
```diff
@@ -173,19 +175,19 @@ function sample(
...
     time = @elapsed for i = 1:n_samples
-        # Make a step
-        t = step(rng, h, τ, t.z)
-        # Adapt h and τ; what mutable is the adaptor
+        # Make a transition
+        t = transition(rng, h, κ, t.z)
+        # Adapt h and κ; what mutable is the adaptor
         tstat = stat(t)
-        h, τ, isadapted = adapt!(h, τ, adaptor, i, n_adapts, t.z.θ, tstat.acceptance_rate)
+        h, κ, isadapted = adapt!(h, κ, adaptor, i, n_adapts, t.z.θ, tstat.acceptance_rate)
```
torfjelde, line 182: Actually, I have one final suggested change.

Atm we have the following defined:
- `transition(rng, hamiltonian, kernel, z)`
- `transition(rng, trajectory, hamiltonian, z)`
Should we maybe have the same ordering on these, e.g. `transition(rng, kernel, hamiltonian, z)` for the first one instead?

torfjelde (changes requested):

src/hamiltonian.jl
```diff
@@ -138,15 +138,32 @@ phasepoint(
     h::Hamiltonian
 ) where {T<:Real} = phasepoint(h, θ, rand(rng, h.metric))
 
+abstract type AbstractRefreshment end
```
yebai, line 141: `AbstractRegreshment` is a bit vague - it's a bit hard to understand what it means. Maybe rename this to `AbstractMomentum`?

src/hamiltonian.jl
```diff
@@ -138,15 +138,32 @@ phasepoint(
...
+
+"Completly resample new momentum."
+struct FullRefreshment <: AbstractRefreshment end
+
 refresh(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
+    ::FullRefreshment,
```
yebai, line 148: Hi @xukai92, as discussed earlier today, I think we can 

1. Define `Trajectory` as a subtype of `AbstractMCMCKernel`
2. Add a field `MomentumRefreshment` to `Trajectory`, or `Hamiltonian`. 

This is probably more consistent with our `Trajectory` type, which is effectively an MCMC kernel -- `Trajectory` contains how we numerically simulate a trajectory, and how to pick a phase point from a simulated trajectory. Therefore wrapping the `Trajectory` type inside another kernel type is a bit redundant in my view.

src/hamiltonian.jl
```diff
@@ -138,15 +138,32 @@ phasepoint(
...
-#     α::AbstractFloat
-# ) = phasepoint(h, z.θ, α * z.r + (1 - α^2) * rand(rng, h.metric))
+"""
+Partial momentum refreshment with refresh rate `α`.
+
+## References
+
+1. Neal, Radford M. "MCMC using Hamiltonian dynamics." Handbook of markov chain monte carlo 2.11 (2011): 2.
+"""
+struct PartialRefreshment{F<:AbstractFloat} <: AbstractRefreshment
```
yebai, line 160:
```suggestion
struct PartialMomentumRefreshment{F<:AbstractFloat} <: AbstractMomentumRefreshment
```

src/hamiltonian.jl
yebai: thanks, @xukai92!

src/hamiltonian.jl
xukai92: I will do a global replacement on my side.

src/sampler.jl
xukai92: I'm aware of this and agree with you. 
Let's do it in an individual PR to avoid change confusion.
I created an issue (https://github.com/TuringLang/AdvancedHMC.jl/issues/248) to track this.

src/sampler.jl
```diff
@@ -42,32 +45,31 @@ function sample_init(
     return h, t
 end
 
-# A step is a momentum refreshment plus a transition
-function step(
+function transition(
```
xukai92, line 48: Pointer.

src/hamiltonian.jl
xukai92: Yes I can see this alternative has its own pros by being less nested.

Also to record the advatange of the current design: it makes introducing a new type of HMC kernel easier as one doesn't need to deal with refreshment and integrator stuff *within the trajectory*.
This is a good result of having this nested design because all pre-simulation code appears only in [this function](https://github.com/TuringLang/AdvancedHMC.jl/pull/247/files#r579826616) now, which are previously located in  the transition function for each trajecotry, seperately. I personally found this is useful from a development potin of view because I did encounter this problem when refactoring the integrator code -- I had to find all the places that the integrator was jittered.

As both sides have its own pros and cons, I guess it's better for us to discuss it a bit before making changes.

xukai92: I created a discussion here https://github.com/TuringLang/AdvancedHMC.jl/discussions/250.

yebai (approved):

src/sampler.jl
torfjelde: Sounds good :+1:

torfjelde (approved):
LGTM!

PR #255, opened 2021-03-31
Basic CUDA support

treigerm:
This PR is mostly based on code @xukai92  wrote and this is an initial draft to merge those changes into AHMC. I have done some local sanity checks that it runs without problems. In the future we might want to look into making this implementation more generic so it works for not just CUDA but also other hardware.

## TODOs

- [x] Testing. What is the best way to test GPU code? Does GitHub support running automated tests on GPUs? Is there a way to check in the test suite whether the current machine is CUDA enabled?
- [x] Documentation.

### DIFF
--- a/src/contrib/cuda.jl
+++ b/src/contrib/cuda.jl
@@ -13,6 +13,8 @@ function refresh(
     return phasepoint(h, z.θ, r)
 end
 
+# TODO: Ideally this should be merged with the CPU version. The function is 
+#       essentially the same but sampling requires a custom call to CUDA.
 function mh_accept_ratio(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}}, Horiginal::TA, Hproposal::TA,
 ) where {T<:AbstractFloat, TA<:CUDA.CuArray{<:T}}
@@ -22,6 +24,8 @@ function mh_accept_ratio(
     #       the chains. We need to revisit this more rigirously 
     #       in the future. See discussions at 
     #       https://github.com/TuringLang/AdvancedHMC.jl/pull/166#pullrequestreview-367216534
-    accept = TA(rand(rng, T, length(Horiginal))) .< α
+    r = CUDA.CuArray{Float32, 1}(undef, length(Horiginal))
+    CUDA.CURAND.rand!(r)
+    accept = r .< α
     return accept, α
 end
--- a/src/contrib/cuda.jl
+++ b/src/contrib/cuda.jl
@@ -1,0 +1,25 @@
+import .CUDA
+
+function refresh(
+    rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
+    ::FullMomentumRefreshment,
+    h::Hamiltonian,
+    z::PhasePoint{T}
+) where {T<:CUDA.CuArray}
+    r = CUDA.CuArray{Float32, 2}(undef, size(h.metric)...)
+    CUDA.CURAND.randn!(r)
+    return phasepoint(h, z.θ, r)
+end
+
+function mh_accept_ratio(
+    rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}}, Horiginal::TA, Hproposal::TA,
+) where {T<:AbstractFloat, TA<:CUDA.CuArray{<:T}}
+    α = min.(one(T), exp.(Horiginal .- Hproposal))
+    # NOTE: There is a chance that sharing the RNG over multiple
+    #       chains for accepting / rejecting might couple
+    #       the chains. We need to revisit this more rigirously 
+    #       in the future. See discussions at 
+    #       https://github.com/TuringLang/AdvancedHMC.jl/pull/166#pullrequestreview-367216534
+    accept = TA(rand(rng, T, length(Horiginal))) .< α
+    return accept, α
+end
--- a/src/contrib/cuda.jl
+++ b/src/contrib/cuda.jl
@@ -6,9 +6,9 @@ function refresh(
     rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
     ::FullMomentumRefreshment,
     h::Hamiltonian,
-    z::PhasePoint{T}
-) where {T<:CUDA.CuArray}
-    r = CUDA.CuArray{Float32, 2}(undef, size(h.metric)...)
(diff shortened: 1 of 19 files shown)


xukai92:
Thanks for the PR Tim!

> Testing. What is the best way to test GPU code? Does GitHub support running automated tests on GPUs?

There is https://github.com/JuliaGPU/buildkite from JuliaGPU but I've never tried it.
I don't either know if this approach is free as it requires GPU machines?
It would be great if you can take a look and let me know what's required to have this set up.

> Is there a way to check in the test suite whether the current machine is CUDA enabled?

`CUDA.functional()` (https://github.com/JuliaGPU/CUDA.jl/blob/master/src/initialization.jl#L9-L23)?

src/contrib/cuda.jl
```diff
@@ -0,0 +1,27 @@
...
+function mh_accept_ratio(
+    rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}}, Horiginal::TA, Hproposal::TA,
+) where {T<:AbstractFloat, TA<:CUDA.CuArray{<:T}}
+    α = min.(one(T), exp.(Horiginal .- Hproposal))
+    # NOTE: There is a chance that sharing the RNG over multiple
+    #       chains for accepting / rejecting might couple
+    #       the chains. We need to revisit this more rigirously 
+    #       in the future. See discussions at 
+    #       https://github.com/TuringLang/AdvancedHMC.jl/pull/166#pullrequestreview-367216534
+    accept = TA(rand(rng, T, length(Horiginal))) .< α
```
xukai92, line 25: Two comments here.
1. It seems that we can also make `TA(rand())` to directly sample on GPU instead of sample on CPU and move to GPU. Should give us some extra perfermance gain but I did miss it in my gist version.
2. It would be nice to merge this whole function with the CPU version somehow. This is not urgent if there is not clear way but worth put a note in the comment as TODO.

src/sampler.jl
```diff
@@ -203,6 +203,10 @@ function sample(
             n_chains = 1
         else
             n_chains = size(θ, 2)
+            # Make sure that arrays are on CPU before printing.
```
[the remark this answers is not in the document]
treigerm: The `join` function which is used for printing relies on scalar indexing which produces a warning when running with a CUDA array i.e. something like:
```
Warning: Performing scalar operations on GPU arrays: This is very slow, consider disallowing these operations with `allowscalar(false)`
```

By doing `CUDA.allowscalar(false)` I detected that `join` seems to be the only place of scalar indexing which is why I did the conversion to CPU arrays.

treigerm:
> There is https://github.com/JuliaGPU/buildkite from JuliaGPU but I've never tried it.
> I don't either know if this approach is free as it requires GPU machines?
> It would be great if you can take a look and let me know what's required to have this set up.

I will look into it!

src/sampler.jl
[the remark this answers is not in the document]
xukai92: I see. Make sense!

src/contrib/cuda.jl
treigerm: Looking at the [CUDA documentation](https://juliagpu.gitlab.io/CUDA.jl/usage/array/#Random-numbers) it looks like it is possible to pass in a CUDA random number generator which directly samples on the GPU. I will look into whether that way we can avoid having special GPU functions for `mh_accept_ratio` and `refresh`.

treigerm: I used the `randn` trick as an initial solution. My comment was in response to your second point above but that discussion can be left for a separate PR!

treigerm:
I added a very basic test which just checks whether the code runs without errors. Also I made a check so that the GPU tests are only run if a GPU is available.

The README at https://github.com/JuliaGPU/buildkite mentions that the JuliaGPU organization is able to provide some infrastructure to run GPU tests. I will write a message on the GPU Slack channel to ask whether we could access to it.

test/common.jl
```diff
@@ -9,43 +9,49 @@ const DETATOL = 1e-3 * D * TRATIO
 # Random tolerance
 const RNDATOL = 5e-2 * D * TRATIO * 2
 
-# Hand-coded multivaite Gaussain
+# Hand-coded multivariate Gaussian
```
xukai92, line 12: Let's make a TODO note on improving this implementation here.
We either need to implement a more Distributions.jl like interface as it's currently done in https://github.com/xukai92/VecTargets.jl or directly use this package if we decide to register it.

test/runtests.jl
```diff
@@ -16,6 +16,14 @@ println(ENV)
         "models",
     ]
 
+    if CUDA.functional()
+        @eval module Test_cuda
```
xukai92, line 20:
```suggestion
        @eval module TestCUDA
```

test/common.jl
treigerm: I guess it would make the most sense to just use https://github.com/xukai92/VecTargets.jl directly because otherwise it seems as though we would just end up implementing the same functionality that's in that package already. Have you tested https://github.com/xukai92/VecTargets.jl on CUDA yet?

treigerm:
Is there anything that would speak against merging this already, i.e. are there things that need to still be done? I can add a section to the README explaining how to use AHMC with CUDA if that's helpful.

xukai92:
> Is there anything that would speak against merging this already, i.e. are there things that need to still be done? 

No. It's not proceeded only because this PR is current as a "Draft" so I thought you are still working on it :D

> I can add a section to the README explaining how to use AHMC with CUDA if that's helpful.

Maybe a very simple one is enough for now.

treigerm:
> No. It's not proceeded only because this PR is current as a "Draft" so I thought you are still working on it :D
Ah I see that makes sense :D. 

I will add a very short paragraph to the README and then make it a proper PR.

xukai92:
Can you change to to Ready for reivew once you're done?
There is also some merge conflicts for Project.toml need to solve.

treigerm:
Will do so!

treigerm:
I'm not sure whether the test failure is related to this PR. The fact that it only happens for one of the CI runs might be an indication that it's a stochastic thing.

xukai92:
After retrigerring, all tests pass. I'm merging this PR now. Thanks for the contribution.

xukai92 (approved):

Issue #253, opened 2021-03-23
Compatibility with ComponentArrays.jl


scheidan:
It would be great if we could use `AdvancedHMC.jl` with models using `ComponenArrays.jl`, especially because the they work nicely together with `ForwardDiff` and `Zygote`.

I think the problem is the function `phasepoint` defined [here](https://github.com/TuringLang/AdvancedHMC.jl/blob/15cb65a4e8c804e69f122ee952ca872fea54ecc1/src/hamiltonian.jl#L77):
```Julia
# If position variable and momentum variable are in different containers,
# move the momentum variable to that of the position variable.
# This is needed for AHMC to work with CuArrays (without depending on it).
phasepoint(
    h::Hamiltonian,
    θ::T1,
    _r::T2;
    r=T1(_r),  # <- this does not work for ComponentArrays!
    ℓπ=∂H∂θ(h, θ),
    ℓκ=DualValue(neg_energy(h, r, θ), ∂H∂r(h, r))
) where {T1<:AbstractVecOrMat,T2<:AbstractVecOrMat} = PhasePoint(θ, r, ℓπ, ℓκ)
```
Not sure how much this is related to #159.

A simple test case:
```Julia
using ComponentArrays

x = randn(100) .+ 3
d1(x, p) = -sum(abs2, x .- p.μ)/(p.σ)^2
d1(p) = d1(x, p)

p1 = ComponentVector(μ = 2.0, σ=1)
d1(p1)

p_grad1(p1) = ForwardDiff.gradient(d1, p1)
p_grad1(p1)


# Define a Hamiltonian system
D = length(p1)          # number of parameters
metric = DiagEuclideanMetric(D)

# choose AD framework or provide a function manually
hamiltonian = Hamiltonian(metric, d1, ForwardDiff)

# Define a leapfrog solver, with initial step size chosen heuristically
initial_ϵ = find_good_stepsize(hamiltonian, p1)  # <- fails here
integrator = Leapfrog(initial_ϵ)

# Define an HMC sampler, with the following components
proposal = NUTS{MultinomialTS, GeneralisedNoUTurn}(integrator)
adaptor = StanHMCAdaptor(MassMatrixAdaptor(metric), StepSizeAdaptor(0.8, integrator))

# -- run sampler
n_samples = 100
n_adapts = 50
raw_samples, stats = sample(hamiltonian, proposal, p1, n_samples,
                            adaptor, n_adapts; progress=true)    # <- fails here too

```

[...]

PR #257, opened 2021-04-01
support ComponentArrays

scheidan:
Implements #253.
Please let me know if I shoudl change something.

### DIFF
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -70,15 +70,18 @@ phasepoint(
 
 # If position variable and momentum variable are in different containers,
 # move the momentum variable to that of the position variable.
-# This is needed for AHMC to work with CuArrays (without depending on it).
-phasepoint(
+# This is needed for AHMC to work with CuArrays and other Arrays (without depending on it).
+function phasepoint(
     h::Hamiltonian,
     θ::T1,
-    _r::T2;
-    r=T1(_r),
-    ℓπ=∂H∂θ(h, θ),
-    ℓκ=DualValue(neg_energy(h, r, θ), ∂H∂r(h, r))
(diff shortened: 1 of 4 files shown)


scheidan:
Some CI tests failed but I think it has nothing to do with this commit. Unfortuantlay it  says only `This check failed` when clicking on details... 

Locally the tests pass.

src/hamiltonian.jl
```diff
@@ -70,15 +70,18 @@ phasepoint(
 
 # If position variable and momentum variable are in different containers,
 # move the momentum variable to that of the position variable.
-# This is needed for AHMC to work with CuArrays (without depending on it).
-phasepoint(
+# This is needed for AHMC to work with CuArrays and other Arrays (without depending on it).
+function phasepoint(
```
xukai92, line 74: Can you implement your change while keeping the current interface of optionally provide `ℓπ` and `ℓκ`?
You can create a function for your version of initializing `r`, e.g. `safe_rsimilar(θ, _r) = (r = similar(θ); copyto!(r, _r); r)`, and update `r=safe_rsimilar((θ, _r))`

xukai92 (changes requested):

xukai92:
Thanks for the PR. 
The proposed way to support ComponentArrays looks great to me.
But the current implementation changes the interface, which is not desirable.
Please check my comments on addressing this.

Also, can you add a test with the use case of ComponentArrays so that we know this PR works as expected?
The example you have in https://github.com/TuringLang/AdvancedHMC.jl/issues/253 should be good enough.
You should be able to add packages only for testing purposes in the `[extras]` section in the Project.toml file.

scheidan:
Thanks for your review @xukai92! Your suggestion made a lot of sense.

xukai92 (approved):

xukai92:
Great! The changes look good to me. 
It's quite neat to have ComponentArray offically supported in AHMC.
Thanks for the PR again!

PR #259, opened 2021-04-06
Adopt AbstractMCMC.jl interface

torfjelde:
## Things to discuss
- [x] AFAIK, the way to customize the logging in AbstractMCMC.jl is to pass `progress=false` to the underlying `AbstractMCMC.mcmcsample` and then use the `callback` keyword argument to log the progress. So the question is: should we do this so as to preserve the current logging functionality?
- [x] To replicate the current summarization functionality (e.g. inform the user of average acceptance rates and EBFMI) as a post-sample step, we can overload `StatsBase.sample` and then perform this step after the call to `AbstractMCMC.mcmcsample`. Should we do this?


src/trajectory.jl
```diff
@@ -211,7 +213,7 @@ nsteps(τ::Trajectory{TS, I, TC}) where {TS, I, TC<:FixedIntegrationTime} =
 ## Kernel interface
 ##
 
-struct HMCKernel{R, T<:Trajectory} <: AbstractMCMCKernel 
+struct HMCKernel{R, T<:Trajectory} <: AbstractMCMCKernel
```
xukai92, line 216: Just to double check this is just the removal of a space?

src/sampler.jl
```diff
@@ -88,125 +87,68 @@ function Adaptation.adapt!(
...
-    end
-    return θs, stats
+    # Compute next transition and state.
+    state = HMCState(0, t, adaptor)
+
+    # Take actual first step
+    return AbstractMCMC.step(rng, model, spl, state; kwargs...)
+end
+
+function AbstractMCMC.step(
```
xukai92, line 124: I think we should keep the old signature, which would construct `HamiltonianModel` and `HMCState` then call this function.

src/sampler.jl
```diff
@@ -88,125 +87,68 @@ function Adaptation.adapt!(
...
-sample(
-    h::Hamiltonian,
-    κ::AbstractMCMCKernel,
-    θ::AbstractVecOrMat{<:AbstractFloat},
-    n_samples::Int,
+function AbstractMCMC.step(
+    rng::AbstractRNG,
+    model::HamiltonianModel,
+    spl::HMCKernel;
+    init_params, # TODO: implement this. Just do `rand`? Need dimensionality though.
```
xukai92, line 110: I can see multiple ways to do this but it's unclear what's the best.
Can you file a PR/discussion to chat on this?

xukai92:
> So the question is: should we do this so as to preserve the current logging functionality?

I think we should do this as to keep the functionality unchanged.
We can decide later if that's hard to maintain or not being as useful.

> To replicate the current summarization functionality ...

Does it make sense to add this functionality in AbtractMCMC first? Seems to be a useful thing to have in general.

src/trajectory.jl
torfjelde: Yep. But also, now `AbstractMCMCKernel` is `AbstractMCMC.AbstractSampler` (I'll remove this alias, and make it explicit).

cpfiffer:
>     * [ ]  To replicate the current summarization functionality (e.g. inform the user of average acceptance rates and EBFMI) as a post-sample step, we can overload `StatsBase.sample` and then perform this step after the call to `AbstractMCMC.mcmcsample`. Should we do this?

Yes, that's the intended use case. :+1: 



>     * AFAIK, the way to customize the logging in AbstractMCMC.jl is to pass `progress=false` to the underlying `AbstractMCMC.mcmcsample` and then use the `callback` keyword argument to log the progress. So the question is: should we do this so as to preserve the current logging functionality?

Yeah, I think the AHMC logging stuff should be moved to a callback.

src/sampler.jl
torfjelde: Hmm, maybe. We have already introduced breaking changes on master with the introduction of `HMCKernel`, etc., so it seems like it's as good a time as any to also transition completely over to AbstractMCMC.jl interface?

torfjelde:
> Does it make sense to add this functionality in AbtractMCMC first? Seems to be a useful thing to have in general.

Not needed; with https://github.com/TuringLang/AbstractMCMC.jl/pull/56 merged we can implement this as a callback very easily, and just make this callback the default.

src/sampler.jl
torfjelde: And it's not like there's anything in the current impl that isn't supported by the new implementation:)

src/trajectory.jl
xukai92: Is there any side effect it might cause?

src/sampler.jl
xukai92: There wasn't actually breaking changes to the outmost intefaces (e.g. those need to call sample).

torfjelde: Hmm, what do you mean? E.g. the example in the README doesn't work now since `proposal` isn't constructed in the same way (might also be other issues, but that's the immediate one that came to mind).

src/trajectory.jl
torfjelde: Don't think so. Anything particular in mind or just general question? 

The only thing is that if we remove `AbstractMCMCKernel` completely, not even aliasing, e.g. CoupledHMC.jl won't work.

Maybe best approach is to just make `AbstractMCMCKernel <: AbstractMCMC.AbstractSampler`?

src/sampler.jl
xukai92: > Hmm, what do you mean? E.g. the example in the README doesn't work now since `proposal` isn't constructed in the same way (might also be other issues, but that's the immediate one that came to mind).

Old constructors should be supported through https://github.com/TuringLang/AdvancedHMC.jl/blob/master/src/AdvancedHMC.jl#L67. 
If there is anything missing we should add them back there.
We should aim to keep outmost interface non-breaking for a few releases before completely removing them.

Also it's a bit cumbersome to write `sample(..., HamiltonianModel(hamiltonian), ..)` instead of sample(..., hamiltonian, ..), no?

src/trajectory.jl
xukai92: I think it's fine as long as CoupledHMC is fine (with sutff like mixture kernels still working).

src/sampler.jl
torfjelde: Sure, I'm cool with adding the old `sample` back and deprecating:) That seems like a clean approach :+1:  

> Also it's a bit cumbersome to write sample(..., HamiltonianModel(hamiltonian), ..) instead of sample(..., hamiltonian, ..), no?

I'm also cool with either of the following:
1. Making `AbstractHamiltonian <: AbstractMCMC.AbstractModel`, which IMO seems sensible.
2. Additional overload that will automagically wrap `hamiltonian` in `HamiltonianModel`.

src/sampler.jl
torfjelde: Will do! But for now I'll just extract the dimensionality from the metric and use `rand`. But yeah, I'll open an issue after this PR so we can figure out potential better defaults.

src/sampler.jl
xukai92: It seems that option 1 is more sensible, as we already making kernel sub-typed of samples from AbstractMCMC.

torfjelde: Yeaah I've realized I need to make quite a few changes. Because everything is mutated somehow, we essentially need to put everything in the sampler `state`.

Sooo the question is what we actually use for the `model`, etc. As of right now they would just be "fake" in the sense that they only contain the initial `Hamiltonian` and the initial `AbstractHMCKernel` :confused: We *could* mutate them, but that seems very much against the AbstractMCMC.jl interface...

xukai92: Hmm. Do you mean it actually only make sense to make the log density and gradient as model but treating metric differently?

torfjelde: "Only" is maybe strong, but yeah, I think it makes more sense to keep the `hamiltonian` and the `metric` separate. Potentially do away completely with the `hamiltonian` and only keep metric or, alternatively, make `hamiltonian` nothing but a wrapper around `metric` (in case there are future cases we might want to cover) and just make a `DifferentiableDensityModel <: AbstractModel` (could even move this + `DensityModel` from AbstractMH.jl into AbstractMCMC.jl at some point, so these can be re-used e.g. for MALA which is present in AdvancedHMC.jl)

Project.toml
```diff
@@ -3,6 +3,7 @@ uuid = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
 version = "0.2.28"
```
xukai92, line 3: Please also increment the version here.

torfjelde: Should we make a minor version bump? Seems like quite a significant change given the introduction of `HMCKernel` too

xukai92: Yeah a minor version bump sounds suitable.

torfjelde:
Btw, waiting for https://github.com/TuringLang/Turing.jl/pull/1579 since tests depend on Turing.

yebai:
Is there anything missing from this PR? If there is nothing major, let's try to merge this PR since it will benefit several downstream tasks.

HarrisonWilde:
I've been working off the branch for this PR to support tempering on AHMC via the new MCMCTempering package and noticed that upon finishing sampling, even if I have `using MCMCChains`, I still just get back a vector of transitions from sampling, I think it would be nice to add a function similar to the one in Turing/Inference that implements `AbstractMCMC.bundle_samples` to actually make use of `Chains`, what do others think @torfjelde @yebai ?

HarrisonWilde:
I see yes, that is a fair assessment, I am using just AHMC, haven't tried with Turing but I agree it perhaps makes less sense to have something specific here for working with AHMC. On another note, I agree regarding Tor's comments on the state of the `model` currently, it would be better perhaps to rely on some standard definition in AMCMC.

torfjelde:
> What is the minimal expectation for the return from sample though? As currently it appears to be this vector of transitions rather than the more standard vector of thetas you see on master?

Glad you brought this up! I'm not certain tbh :confused: @xukai92 @yebai What do you think? Should it return `PhasePoint` or just thetas? I'm with whatever, but made it return `PhasePoint` initially.

xukai92:
> What do you think? Should it return PhasePoint or just thetas?

I think we should not change the return at least for this PR.
Plus now stuff like ComponentArrays is supported, you can even have simple "variable bindings".

> Not happy with how we have a DifferentiableModel in this repo. This should preferably go somewhere else, e.g. AMCMC. Alternatively we can make use of LogDensityProblems.jl, which might be a good approach (me and @devmotion discussed this on slack a couple of weeks ago).

Both are OK but I don't think we have to make it perfect **within this PR** as long as it's used behind the scene (i.e. users don't need to create it explicitly).
Just hope this PR doesn't keep growing than interface binding.

> - I've noticed something that might be a bug, so need to debug that first.

What's the potential bug?

torfjelde:
> > What do you think? Should it return PhasePoint or just thetas?
> 
> I think we should not change the return at least for this PR.

Sounds good; I'll get that sorted out.

> Plus now stuff like ComponentArrays is supported, you can even have simple "variable bindings".

Sure, but then we'd still need to depend on ComponentArrays.jl to be able to implement a `bundle_samples(..., MCMCChains.Chains)` method + assumes that the user is using it. IMO AMC shouldn't care, and this should be implemented somewhere else (alternatively we can provide an example somewhere).

> > Not happy with how we have a DifferentiableModel in this repo. This should preferably go somewhere else, e.g. AMCMC. Alternatively we can make use of LogDensityProblems.jl, which might be a good approach (me and @devmotion discussed this on slack a couple of weeks ago).
> 
> Both are OK but I don't think we have to make it perfect **within this PR** as long as it's used behind the scene (i.e. users don't need to create it explicitly).

True :+1: But currently the users do need to instantiate the `DifferentiableModel` (well, they don't _have_ to, but the intention is that they can use this), so it is exposed to the user.

I'll take a look at what it would take to use LogDensityProblems.jl's approach, and then make a decision from there. I agree that we don't need to solve this completely right now.

> > * I've noticed something that might be a bug, so need to debug that first.
> 
> What's the potential bug?

I don't think it is after all:) I observed some "strange" behavior in a problem I'm working on where the tree-depth during sampling was _always_ one less than the maximum tree-depth, while in adaptation you'd see the "expected" behavior where it changed between iterations. But I increased the adaptation phase and then observe the same behavior at the end of adaptation, so it seems like it's just an artifact of the particular problem I'm working on. You've seen this before? It just seems a bit strange to me that it's so "deterministic" (also this is a very high-dimensional problem).

xukai92:
> Sure, but then we'd still need to depend on ComponentArrays.jl

Aren't the implementation to bundle stuff is just for `AbstractArray`?

> currently the users do need to instantiate the DifferentiableModel

Can we keep the same interface as before and convert Hamiltonian to DifferentiableModel under the hood?

> You've seen this before?

It indeed sounds a bit weird but I don't have any idea why this could happen.
Did you try to increase the maximum and see if the depth still trys to "catch up" to max minus one?

yebai:
> where the tree-depth during sampling was always one less than the maximum tree-depth

@xukai92 I remember the counter for tree-depth was a bit confusing. Do we still record the actual tree-depth minus one internally? If so, it might be that NUTS is reaching the tree-depth limit during the sampling phase for Tor's problem.

torfjelde:
> Aren't the implementation to bundle stuff is just for AbstractArray?

Yes. My point is that we can't make use of `keys` in `bundle_samples` when we can't guarantee they exist. But I'm guessing you just meant that if we're using `ComponentArray`, and AHMC bundles these into a vector, then this can then be converted into a `Chains` later? If so I'm with you; I'm just saying that the last part shouldn't go into AHMC:)

> Can we keep the same interface as before and convert Hamiltonian to DifferentiableModel under the hood?

Yeah, so this is what we're doing right now. Both using `Hamiltonian` and `DifferentiableModel` is directly is possible.

> It indeed sounds a bit weird but I don't have any idea why this could happen.
> Did you try to increase the maximum and see if the depth still trys to "catch up" to max minus one?

Didn't, but will; thanks :+1:

sethaxen:
> @xukai92 I remember the counter for tree-depth was a bit confusing. Do we still record the actual tree-depth minus one internally? If so, it might be that NUTS is reaching the tree-depth limit during the sampling phase for Tor's problem.

We found it's a bit unclear in PPLs regarding what exactly `tree_depth` means and how it should relate to `n_steps`. e.g. `n_steps` and `tree_depth` seem to relate to each other the same way in Stan and here but differently than in PyMC3. See https://github.com/arviz-devs/arviz/pull/1063#discussion_r499258271. Is `tree_depth=1` a single-node tree, a 2-node tree, or a 4-node tree? Does a `tree_depth` of `n` mean that `n` tree doublings were accepted or proposed? perhaps here we are accidentally using two different definitions, one to count `tree_depth` and another to relate it to the maximum depth allowed?

yebai:
@sethaxen thanks for the information - we're happy to adopt a more standardised terminology if that helps. Also happy to collaborate more closely with ArviZ devs on improving MCMC workflows.

sethaxen:
> @sethaxen thanks for the information - we're happy to adopt a more standardised terminology if that helps.

I'm not certain there _is_ a standardised terminology here. The most important thing I think is to clearly define what the terms mean in this package and then include tests that the implied relationships are as expected.

> Also happy to collaborate more closely with ArviZ devs on improving MCMC workflows.

ArviZ's focus is mostly on analysis of posteriors, focusing on diagnostics and visualizations, while we devs individually contribute to different PPLs. What sorts of collaborations RE MCMC workflows did you have in mind?

yebai:
>  ArviZ's focus is mostly on analysis of posteriors, focusing on diagnostics and visualizations, while we devs individually contribute to different PPLs. What sorts of collaborations RE MCMC workflows did you have in mind?

For example, we are happy to transfer some parts of MCMCChains (e.g. diagnostics) to ArviZ. But ideally, we want to keep maintain them so that after transferring these implementations remain compatible with Turing. Additionally, it is helpful to have a suite of challenging tasks where current MCMC algorithms are known to perform poorly; such a suite is useful for testing the diagnostic tools developed by ArviZ, and for testing novel MCMC sampling methods. Just some initial thoughts. 

cc @devmotion

sethaxen:
> > ArviZ's focus is mostly on analysis of posteriors, focusing on diagnostics and visualizations, while we devs individually contribute to different PPLs. What sorts of collaborations RE MCMC workflows did you have in mind?
> 
> For example, we are happy to transfer some parts of MCMCChains (e.g. diagnostics) to ArviZ. But ideally, we are happy to keep maintain them so that after transferring these implementations remain compatible with Turing.

Yes, I proposed this in https://github.com/arviz-devs/ArviZ.jl/issues/129#issue-893730587 and https://github.com/TuringLang/MCMCChains.jl/issues/266#issuecomment-843010728, and I think this makes sense. And absolutely Turing devs should keep co-maintaining them with the ArviZ devs. A good way to ensure that works smoothly would be for one or more Turing core devs to join the ArviZ devs. There's no commitment or requirement there; it might just help facilitate better communication and integration between Turing and ArviZ, which I think would be a good thing anyways. Since @torfjelde has previously contributed to ArviZ.jl and @devmotion has a keen interest in the diagnostics and has already started work on https://github.com/devmotion/InferenceDiagnostics.jl, I'd nominate them, but perhaps that's something you can discuss amongst yourselves and who might be interested in joining.

> Additionally, it is helpful to have a suite of challenging tasks where current MCMC algorithms are known to perform poorly; such a suite is useful for testing the diagnostic tools developed by ArviZ, and for testing novel MCMC sampling methods. Just some initial thoughts.

I think this is definitely useful, and Julia PPLs should have something like this. A few such efforts exist: e.g. Stan has [posteriordb](https://github.com/stan-dev/posteriordb), TFP has [inference-gym](https://pypi.org/project/inference-gym/), and then there's [PPL Bench](https://github.com/facebookresearch/pplbench) for Jags, Pyro, NumPyro, PyMC3, and Stan. Necessarily, these efforts tend to be a bit PPL-specific, so I doubt there could be a formal collaboration here between ArviZ and Turing. But I would be interested in contributing to such a suite, and we could of course use ArviZ's diagnostics to critique the models.

Sorry @torfjelde for how off-topic this has gotten! Perhaps we should continue this discussion elsewhere.

xukai92:
> Yes. My point is that we can't make use of `keys` in `bundle_samples` when we can't guarantee they exist. But I'm guessing you just meant that if we're using `ComponentArray`, and AHMC bundles these into a vector, then this can then be converted into a `Chains` later? If so I'm with you; I'm just saying that the last part shouldn't go into AHMC:)

For this PR we don't worry about interfacing with `Chains`.
We don't even want to bundle them into a vector as for now;
the current interface doesn't do this and we should keep it the same.

HarrisonWilde:
Just wanted to note, when I dev this PR and try to use latest version of Turing, the AHMC samplers don't seem to work:
```julia
julia > using Turing
[ Info: Precompiling Turing [fce5fe82-541a-59a6-adf8-730c64b5f9a0]

julia > using StatsPlots

julia > # Define a simple Normal model with unknown mean and variance.
        @model function gdemo(x, y)
          s² ~ InverseGamma(2, 3)
          m ~ Normal(0, sqrt(s²))
          x ~ Normal(m, sqrt(s²))
          y ~ Normal(m, sqrt(s²))
        end
gdemo (generic function with 1 method)

julia > #  Run sampler, collect results
        chn = sample(gdemo(1.5, 2), HMC(0.1, 5), 1000)
Sampling 100%|███████████████████████████████████████████████████████████████████████████████████████████████| Time: 0:00:22
ERROR: MethodError: no method matching step(::Random._GLOBAL_RNG, ::AdvancedHMC.Hamiltonian{AdvancedHMC.UnitEuclideanMetric{Float64, Tuple{Int64}}, Turing.Inference.var"#logπ#52"{DynamicPPL.TypedVarInfo{NamedTuple{(:s², :m), Tuple{DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:s², Tuple{}}, Int64}, Vector{InverseGamma{Float64}}, Vector{AbstractPPL.VarName{:s², Tuple{}}}, Vector{Float64}, Vector{Set{DynamicPPL.Selector}}}, DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:m, Tuple{}}, Int64}, Vector{Normal{Float64}}, Vector{AbstractPPL.VarName{:m, Tuple{}}}, Vector{Float64}, Vector{Set{DynamicPPL.Selector}}}}}, Float64}, DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}}, DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}}, Turing.Inference.var"#∂logπ∂θ#51"{DynamicPPL.TypedVarInfo{NamedTuple{(:s², :m), Tuple{DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:s², Tuple{}}, Int64}, Vector{InverseGamma{Float64}}, Vector{AbstractPPL.VarName{:s², Tuple{}}}, Vector{Float64}, Vector{Set{DynamicPPL.Selector}}}, DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:m, Tuple{}}, Int64}, Vector{Normal{Float64}}, Vector{AbstractPPL.VarName{:m, Tuple{}}}, Vector{Float64}, Vector{Set{DynamicPPL.Selector}}}}}, Float64}, DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}}, DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}}}, ::AdvancedHMC.HMCKernel{AdvancedHMC.FullMomentumRefreshment, AdvancedHMC.Trajectory{AdvancedHMC.EndPointTS, AdvancedHMC.Leapfrog{Float64}, AdvancedHMC.FixedNSteps}}, ::AdvancedHMC.PhasePoint{Vector{Float64}, AdvancedHMC.DualValue{Float64, Vector{Float64}}})
Closest candidates are:
  step(::AdvancedHMC.AbstractLeapfrog{T}, ::AdvancedHMC.Hamiltonian, ::P) where {T<:(Union{AbstractVector{var"#s55"}, var"#s55"} where var"#s55"<:AbstractFloat), P<:AdvancedHMC.PhasePoint} at /Users/harrisonwilde/Library/Mobile Documents/com~apple~CloudDocs/PhD/Pumas/AdvancedHMC.jl/src/integrator.jl:58
  step(::AdvancedHMC.AbstractLeapfrog{T}, ::AdvancedHMC.Hamiltonian, ::P, ::Int64; fwd, full_trajectory) where {T<:(Union{AbstractVector{var"#s56"}, var"#s56"} where var"#s56"<:AbstractFloat), P<:AdvancedHMC.PhasePoint} at /Users/harrisonwilde/Library/Mobile Documents/com~apple~CloudDocs/PhD/Pumas/AdvancedHMC.jl/src/integrator.jl:58
Stacktrace:
  [1] initialstep(rng::Random._GLOBAL_RNG, model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, spl::DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}}, vi::DynamicPPL.TypedVarInfo{NamedTuple{(:s², :m), Tuple{DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:s², Tuple{}}, Int64}, Vector{InverseGamma{Float64}}, Vector{AbstractPPL.VarName{:s², Tuple{}}}, Vector{Float64}, Vector{Set{DynamicPPL.Selector}}}, DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:m, Tuple{}}, Int64}, Vector{Normal{Float64}}, Vector{AbstractPPL.VarName{:m, Tuple{}}}, Vector{Float64}, Vector{Set{DynamicPPL.Selector}}}}}, Float64}; init_params::Nothing, nadapts::Int64, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Turing.Inference ~/.julia/packages/Turing/TbEZL/src/inference/hmc.jl:198
  [2] initialstep(rng::Random._GLOBAL_RNG, model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, spl::DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}}, vi::DynamicPPL.TypedVarInfo{NamedTuple{(:s², :m), Tuple{DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:s², Tuple{}}, Int64}, Vector{InverseGamma{Float64}}, Vector{AbstractPPL.VarName{:s², Tuple{}}}, Vector{Float64}, Vector{Set{DynamicPPL.Selector}}}, DynamicPPL.Metadata{Dict{AbstractPPL.VarName{:m, Tuple{}}, Int64}, Vector{Normal{Float64}}, Vector{AbstractPPL.VarName{:m, Tuple{}}}, Vector{Float64}, Vector{Set{DynamicPPL.Selector}}}}}, Float64})
    @ Turing.Inference ~/.julia/packages/Turing/TbEZL/src/inference/hmc.jl:153
  [3] step(rng::Random._GLOBAL_RNG, model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, spl::DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}}; resume_from::Nothing, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ DynamicPPL ~/.julia/packages/DynamicPPL/emAbx/src/sampler.jl:87
  [4] step(rng::Random._GLOBAL_RNG, model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, spl::DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}})
    @ DynamicPPL ~/.julia/packages/DynamicPPL/emAbx/src/sampler.jl:62
  [5] macro expansion
    @ ~/.julia/packages/AbstractMCMC/ByHEr/src/sample.jl:123 [inlined]
  [6] macro expansion
    @ ~/.julia/packages/ProgressLogging/6KXlp/src/ProgressLogging.jl:328 [inlined]
  [7] (::AbstractMCMC.var"#21#22"{Bool, String, Nothing, Int64, Int64, Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}}, Random._GLOBAL_RNG, DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}}, Int64, Int64})()
    @ AbstractMCMC ~/.julia/packages/AbstractMCMC/ByHEr/src/logging.jl:11
  [8] with_logstate(f::Function, logstate::Any)
    @ Base.CoreLogging ./logging.jl:491
  [9] with_logger(f::Function, logger::LoggingExtras.TeeLogger{Tuple{LoggingExtras.EarlyFilteredLogger{TerminalLoggers.TerminalLogger, AbstractMCMC.var"#1#3"{Module}}, LoggingExtras.EarlyFilteredLogger{Logging.ConsoleLogger, AbstractMCMC.var"#2#4"{Module}}}})
    @ Base.CoreLogging ./logging.jl:603
 [10] with_progresslogger(f::Function, _module::Module, logger::Logging.ConsoleLogger)
    @ AbstractMCMC ~/.julia/packages/AbstractMCMC/ByHEr/src/logging.jl:34
 [11] macro expansion
    @ ~/.julia/packages/AbstractMCMC/ByHEr/src/logging.jl:10 [inlined]
 [12] mcmcsample(rng::Random._GLOBAL_RNG, model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, sampler::DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}}, N::Int64; progress::Bool, progressname::String, callback::Nothing, discard_initial::Int64, thinning::Int64, chain_type::Type, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ AbstractMCMC ~/.julia/packages/AbstractMCMC/ByHEr/src/sample.jl:114
 [13] sample(rng::Random._GLOBAL_RNG, model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, sampler::DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}}, N::Int64; chain_type::Type, resume_from::Nothing, progress::Bool, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Turing.Inference ~/.julia/packages/Turing/TbEZL/src/inference/Inference.jl:156
 [14] sample(rng::Random._GLOBAL_RNG, model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, sampler::DynamicPPL.Sampler{HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}}, N::Int64)
    @ Turing.Inference ~/.julia/packages/Turing/TbEZL/src/inference/Inference.jl:155
 [15] sample(rng::Random._GLOBAL_RNG, model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, alg::HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}, N::Int64; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Turing.Inference ~/.julia/packages/Turing/TbEZL/src/inference/Inference.jl:142
 [16] sample(rng::Random._GLOBAL_RNG, model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, alg::HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}, N::Int64)
    @ Turing.Inference ~/.julia/packages/Turing/TbEZL/src/inference/Inference.jl:142
 [17] sample(model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, alg::HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}, N::Int64; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Turing.Inference ~/.julia/packages/Turing/TbEZL/src/inference/Inference.jl:132
 [18] sample(model::DynamicPPL.Model{var"#5#6", (:x, :y), (), (), Tuple{Float64, Int64}, Tuple{}}, alg::HMC{Turing.Core.ForwardDiffAD{40}, (), AdvancedHMC.UnitEuclideanMetric}, N::Int64)
    @ Turing.Inference ~/.julia/packages/Turing/TbEZL/src/inference/Inference.jl:132
 [19] top-level scope
    @ REPL[4]:2

```
Not entirely sure what is going on, I have tried this with a clean setup so I am pretty sure it should be reproducible if someone else tries but I may just be being stupid, thought it was worth flagging up.

xukai92:
I believe we will need a patch on the Turing side as well to make this branch works.

torfjelde:
> I believe we will need a patch on the Turing side as well to make this branch work.

Yeah, there are breaking changes in the current master, unrelated to this PR.

theogf:
Hey! I have also been trying out this branch but Turing free. And I faced one main issue. I tried to use `ForwardDiff` and I defined my `DifferentiableDensityModel` myself by passing function and gradient, except I did it wrong by not expecting that `∂ℓπ∂θ` should return both the value and the gradient.
Also I noticed that the `Hamiltonian` created in the test is never used...
It would be nice to have constructors for `DifferentiableDensityModel` like the one for `Hamiltonian`, which given an AD package will do the right thing.

By the way the code failed because of the `DualValue` constructor expecting Scalar + Gradient which failed since `∂H∂θ` passed x[1], x[2]. Maybe one should check that the `res` from `∂H∂θ` is indeed a `Tuple`

Very cool stuff otherwise :D I am looking forward to see this merged

torfjelde:
Everything is ready to go now, once #270 is merged.

IMO, we should make it a breaking release since there might exist some code that expects to receive the old trajectory type but which now receives a `HMCKernel`, e.g. https://github.com/TuringLang/Turing.jl/pull/1660, even though AFAIK none of the exported functions are doing anything different. I've therefore bumped the minor version here.

test/abstractmcmc.jl
```diff
@@ -0,0 +1,29 @@
...
+@testset "`gdemo`" begin
+    rng = MersenneTwister(0)
+
+    n_samples = 5_000
+    n_adapts = 1_000
+
+    θ_init = randn(rng, 2)
+
+    metric = DiagEuclideanMetric(2)
+    h = Hamiltonian(metric, ℓπ_gdemo, ForwardDiff)
```
theogf, line 14: `h` is not used here. This is slightly misleading :)

torfjelde: Ah crap, this code isn't even runnable! E.g. where's `demo`? I've forgotten to include it in the `runtests.jl` :facepalm:

xukai92:
The `CI` workflow also fails here. Any idea?

torfjelde:
> The CI workflow also fails here. Any idea?

It might just be that the settings I have for the sampler is bad because it passes most of the time.

xukai92:
> It might just be that the settings I have for the sampler is bad because it passes most of the time.

They were fine in the other PR so I suppose you mean it's due to randomness?

Seems like there are also conflicts. 
Let's resolve them and see if the CI passes with a re-run.

torfjelde:
Btw, one annoying thing is that `find_good_stepsize` isn't supported by `DifferentiableDensityModel` yet. There's a bit too much functionality that uses the hamiltonian in there atm, so I haven't bothered trying to refactor that part yet. But this is something we can always do later anyways.

xukai92:
> Btw, one annoying thing is that `find_good_stepsize` isn't supported by `DifferentiableDensityModel` yet. There's a bit too much functionality that uses the hamiltonian in there atm, so I haven't bothered trying to refactor that part yet. But this is something we can always do later anyways.

Doing it in another PR sounds more reasonable. Can you create an issue for this along with other missing thing from `DifferentiableDensityModel` or even other alternatives (I believe we discussed a few).

xukai92:
How could `IntegrationTests` also pass even Turing is not compatible?

torfjelde:
> Doing it in another PR sounds more reasonable. Can you create an issue for this along with other missing thing from DifferentiableDensityModel or even other alternatives (I believe we discussed a few).

Will do :+1:

> How could IntegrationTests also pass even Turing is not compatible?

Because this is a new minor version for AHMC, which won't be compat with Turing.jl since it has compat-bounds on AHMC + it indicates it's a breaking release and so Turing.jl should be expected to have to make some changes to be compatible with the new version. If you look at the integration tests you see `Info: Not compatible with this release. No problem.`.

So it's just a question of whether we make the tests pass or fail in this case; I went with letting it pass since this is also what we do in other Turing.jl-packages, e.g. DPPL.

torfjelde:
Okay this is just strange. It fails on macOS but seems to do fine otherwise.

xukai92:
> So it's just a question of whether we make the tests pass or fail in this case; I went with letting it pass since this is also what we do in other Turing.jl-packages, e.g. DPPL. 

I see. Makes sense to me now.

> Okay this is just strange. It fails on macOS but seems to do fine otherwise.

Could be randomness again but I feel using the same parameters as the non-AMCMC version is enough. Otherwise it might indicate some issues. 

If I give same initial position and same random seed, do I expect to get exact same samples from the interface implementation? If so maybe we can add a test to check if samples are exactly the same.

torfjelde:
I realized what was causing the issue: I was including the adapted samples in the mean-computation:)

Now that those are dropped, it works. _But_ now there's a different issue? This one seems completely unrelated to the changes in this PR though. You got any clue @xukai92 ?

xukai92:
Looks like the tests are fine now. Is turning off the progress (the last commit) the fix?

torfjelde:
> Looks like the tests are fine now. Is turning off the progress (the last commit) the fix?

No, that was just to make the logs nicer. It seems like one of the adaptation tests can fail if we're unlucky with the random seed, but as I said, this has nothing to do with this PR so IMO we merge and I'll make an issue reminding us that the adaptation test is fickle.

yebai (approved):

PR #263, opened 2021-04-20
Fix type-issue in PhasePoint

torfjelde:
`Inf` is hardcoded as `Float64` in `PhasePoint` constructor, which causes issues, e.g.:
```julia
julia> import AdvancedHMC: PhasePoint, DualValue

julia> d1 = DualValue(Float32(Inf), [0f0])
DualValue{Float32, Vector{Float32}}(Inf32, Float32[0.0])

julia> d2 = DualValue(0f0, [0f0])
DualValue{Float32, Vector{Float32}}(0.0f0, Float32[0.0])

julia> PhasePoint([0f0], [0f0], d1, d2)
┌ Warning: The current proposal will be rejected due to numerical error(s).
│   isfinite.((θ, r, ℓπ, ℓκ)) = (true, true, false, true)
└ @ AdvancedHMC ~/Projects/public/AdvancedHMC.jl/src/hamiltonian.jl:47
ERROR: MethodError: Cannot `convert` an object of type 
  DualValue{Float64,Array{Float32{},1}} to an object of type 
  DualValue{Float32,Array{Float32{},1}}
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:205
Stacktrace:
 [1] PhasePoint(θ::Vector{Float32}, r::Vector{Float32}, ℓπ::DualValue{Float32, Vector{Float32}}, ℓκ::DualValue{Float32, Vector{Float32}})
   @ AdvancedHMC ~/Projects/public/AdvancedHMC.jl/src/hamiltonian.jl:51
 [2] top-level scope
   @ REPL[102]:1
```

Also, is there a reason why we're doing this `map` to check if something is finite and replace it with `inf` if it's not @xukai92 ? I mean, if `isfinite` returns `false` then the value should already be `inf`, no?

### DIFF
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -45,8 +45,9 @@ struct PhasePoint{T<:AbstractVecOrMat{<:AbstractFloat}, V<:DualValue}
         @argcheck length(θ) == length(r) == length(ℓπ.gradient) == length(ℓπ.gradient)
         if any(isfinite.((θ, r, ℓπ, ℓκ)) .== false)
             @warn "The current proposal will be rejected due to numerical error(s)." isfinite.((θ, r, ℓπ, ℓκ))
-            ℓπ = DualValue(map(v -> isfinite(v) ? v : -Inf, ℓπ.value), ℓπ.gradient)
-            ℓκ = DualValue(map(v -> isfinite(v) ? v : -Inf, ℓκ.value), ℓκ.gradient)
+            E = eltype(T)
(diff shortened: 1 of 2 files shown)


xukai92 (approved):

xukai92:
> I mean, if isfinite returns false then the value should already be inf, no?

`isfinite(NaN)` is `false` as well, in which case `NaN` is converted to `-Inf`.

xukai92:
Also positive infinity would be made negative for rejection as well, in this case.

torfjelde:
Hmm, aight :+1:

PR #267, opened 2021-06-04
PhasePoint constructor bug when using GPU

treigerm:
There seems to be a type-inference issue when using the PhasePoint constructor with CUDA.

MWE:
```julia
using CUDA
using AdvancedHMC: PhasePoint, DualValue

PhasePoint(
    CuArray([T(NaN) T(NaN)]),
    CuArray([T(NaN) T(NaN)]),
    DualValue(CuArray(zeros(T, 2)), CuArray(zeros(T, 1, 2))),
    DualValue(CuArray(zeros(T, 2)), CuArray(zeros(T, 1, 2)))
)
```
results in 
```
┌ Warning: The current proposal will be rejected due to numerical error(s).
│   isfinite.((θ, r, ℓπ, ℓκ)) = (false, false, true, true)
└ @ AdvancedHMC ~/dev/AdvancedHMC.jl/src/hamiltonian.jl:47
ERROR: LoadError: GPU broadcast resulted in non-concrete element type Any.
This probably means that the function you are broadcasting contains an error or type instability.
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] copy
   @ ~/.julia/packages/GPUArrays/Z5nPF/src/host/broadcast.jl:44 [inlined]
 [3] materialize(bc::Base.Broadcast.Broadcasted{CUDA.CuArrayStyle{1}, Nothing, AdvancedHMC.var"#71#73"{DataType}, Tuple{CuArray{Float32, 1}}})
   @ Base.Broadcast ./broadcast.jl:883
 [4] map(::Function, ::CuArray{Float32, 1})
   @ GPUArrays ~/.julia/packages/GPUArrays/Z5nPF/src/host/broadcast.jl:86
 [5] PhasePoint(θ::CuArray{Float32, 2}, r::CuArray{Float32, 2}, ℓπ::DualValue{CuArray{Float32, 1}, CuArray{Float32, 2}}, ℓκ::DualValue{CuArray{Float32, 1}, CuArray{Float32, 2}})
   @ AdvancedHMC ~/dev/AdvancedHMC.jl/src/hamiltonian.jl:49
 [6] top-level scope
   @ ~/dev/hmc_benchmark/ahmc/phasepoint_mwe.jl:4
 [7] include(fname::String)
   @ Base.MainInclude ./client.jl:444
 [8] top-level scope
   @ REPL[14]:1
 [9] top-level scope
   @ ~/.julia/packages/CUDA/3VnCC/src/initialization.jl:81
in expression starting at /home/reichelt/dev/hmc_benchmark/ahmc/phasepoint_mwe.jl:4
```

The fix is simple and I copied the PhasePoint tests to `test/cuda.jl` and adjusted them for the GPU.

### DIFF
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -45,6 +45,7 @@ struct PhasePoint{T<:AbstractVecOrMat{<:AbstractFloat}, V<:DualValue}
         @argcheck length(θ) == length(r) == length(ℓπ.gradient) == length(ℓπ.gradient)
         if any(isfinite.((θ, r, ℓπ, ℓκ)) .== false)
             @warn "The current proposal will be rejected due to numerical error(s)." isfinite.((θ, r, ℓπ, ℓκ))
+            # NOTE eltype has to be inlined to avoid type stability issue; see #267
             ℓπ = DualValue(
                 map(v -> isfinite(v) ? v : -eltype(T)(Inf), ℓπ.value), 
                 ℓπ.gradient
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -45,9 +45,14 @@ struct PhasePoint{T<:AbstractVecOrMat{<:AbstractFloat}, V<:DualValue}
         @argcheck length(θ) == length(r) == length(ℓπ.gradient) == length(ℓπ.gradient)
         if any(isfinite.((θ, r, ℓπ, ℓκ)) .== false)
             @warn "The current proposal will be rejected due to numerical error(s)." isfinite.((θ, r, ℓπ, ℓκ))
-            E = eltype(T)
-            ℓπ = DualValue(map(v -> isfinite(v) ? v : -E(Inf), ℓπ.value), ℓπ.gradient)
-            ℓκ = DualValue(map(v -> isfinite(v) ? v : -E(Inf), ℓκ.value), ℓκ.gradient)
+            ℓπ = DualValue(
+                map(v -> isfinite(v) ? v : -eltype(T)(Inf), ℓπ.value), 
+                ℓπ.gradient
+            )
+            ℓκ = DualValue(
+                map(v -> isfinite(v) ? v : -eltype(T)(Inf), ℓκ.value), 
+                ℓκ.gradient
+            )
         end
         new{T,V}(θ, r, ℓπ, ℓκ)
     end
--- a/test/cuda.jl
+++ b/test/cuda.jl
@@ -1,5 +1,6 @@
 using Test
 using AdvancedHMC
+using AdvancedHMC: DualValue, PhasePoint
 using CUDA
 
 @testset "AdvancedHMC GPU" begin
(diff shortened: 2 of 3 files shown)


yebai (dismissed):

yebai:
many thanks, @treigerm!

src/hamiltonian.jl
```diff
@@ -45,9 +45,14 @@ struct PhasePoint{T<:AbstractVecOrMat{<:AbstractFloat}, V<:DualValue}
...
-            ℓπ = DualValue(map(v -> isfinite(v) ? v : -E(Inf), ℓπ.value), ℓπ.gradient)
-            ℓκ = DualValue(map(v -> isfinite(v) ? v : -E(Inf), ℓκ.value), ℓκ.gradient)
+            ℓπ = DualValue(
+                map(v -> isfinite(v) ? v : -eltype(T)(Inf), ℓπ.value), 
+                ℓπ.gradient
+            )
+            ℓκ = DualValue(
+                map(v -> isfinite(v) ? v : -eltype(T)(Inf), ℓκ.value), 
+                ℓκ.gradient
+            )
```
xukai92, line 55: Maybe I'm missing something, but the new version looks equivalent to that original one?

treigerm: It is exactly the same but it seems that the GPU compiler has some problems inferring that the anonymous function is type-stable if `eltype(T)` is stored in a separate variable. I don't know enough about the GPUCompiler and Julia's type inference whether this is expected behaviour or not.

xukai92: This is interesting to know. 
Can you leave a short note on this so that no one is gonna "optimize" the code to revert this change?

treigerm: So I tried the following
```julia
function my_fun(arr::T) where {T<:Matrix{<:AbstractFloat}}
     E = eltype(T)
     @show Base.return_types(v -> isfinite(v) ? v : eltype(T)(-Inf), (eltype(arr),))
     @show Base.return_types(v -> isfinite(v) ? v : E(-Inf), (eltype(arr),))
end

my_fun(randn(4, 4))
```
which gives
```
Base.return_types((v->begin
            #= REPL[22]:3 =#
            if isfinite(v)
                v
            else
                (eltype(T))(-Inf)
            end
        end), (eltype(arr),)) = Any[Float64]
Base.return_types((v->begin
            #= REPL[22]:4 =#
            if isfinite(v)
                v
            else
                E(-Inf)
            end
        end), (eltype(arr),)) = Any[Any]
```
This suggests that the first version is type-stable, while the second version is not, right? I don't know that much about type stability in Julia so I am not that confident in interpreting these results. Also, I am using Julia 1.6.

xukai92: I agree with your interpretation.
But I don't know why this happens either.
Maybe @mohamed82008 knows?

mohamed82008: Probably the closure bug.

xukai92 (approved):

PR #270, opened 2021-07-14
Improved testing suite

torfjelde:
The testing suite is a bit messy, e.g. we depend on Turing.jl which in turn means that changes made here need to be propagated to Turing.jl before the tests will pass (example: #259).

This PR makes the following changes:
1. Turing.jl and MCMCDebugging.jl and the tests which have these dependencies are removed.
   - Reason: Though it provided a starting point for an interesting direction, AFAIK the additional test doesn't really add much to the testing suite but introduces these dependencies. I'd like to see something like MCMCDebugging.jl being used to test packages but it seems like it requires some additional work to get stuff going. Also, it might be better suited to target AbstractMCMC.jl now that AHMC.jl also will support this interface.
2. Integration tests for Turing.jl are added in `test/turing` but are not included in the default testing suite.
   - Reason: though AHMC should not be dependent on compat with Turing.jl for a PR to merge, it's nice to know whether a PR breaks Turing.jl-compatibility or not.


test/runtests.jl
```diff
@@ -1,33 +1,70 @@
...
+using Distributed, Test, CUDA, Pkg
+
+using AdvancedHMC: AdvancedHMC
 
 println("Envronment variables for testing")
 println(ENV)
 
+const DIRECTORY_AdvancedHMC = dirname(dirname(pathof(AdvancedHMC)))
+const DIRECTORY_Turing_tests = joinpath(DIRECTORY_AdvancedHMC, "test", "turing")
+const GROUP = get(ENV, "GROUP", "All")
```
xukai92, line 10: Can we name `GROUP` more explict, e.g. `AHMC_TEST_GROUP`?

test/runtests.jl
```diff
@@ -1,33 +1,70 @@
...
+            @eval module $(Symbol("Test_", t))
             include($t * ".jl")
+            end
+            return
+        end
+    end
+
+    if GROUP == "All" || GROUP == "Downstream"
+        @testset "turing" begin
+            try
```
xukai92, line 44: I suppose this block is to handle the change of how test-only dependencies are handled in `Project.toml`?

xukai92 (commented):
Cool. I like the proposal of removing MCMCDebugging and potentially consider adding it to AbstractMCMC.

.github/workflows/IntegrationTests.yml
```diff
@@ -0,0 +1,40 @@
+name: IntegrationTests
```
xukai92, line 1: How different is this workflow from that in `CI.yml`?

torfjelde: Absolutely no other difference:)

I just made it separate to make it easy to see which tests are suceeding/failing.

test/runtests.jl
torfjelde: Not sure I completely understand what you mean. But the reason why I introduced it is because there are some namespaces issues, e.g. if `using AdvancedHMC` occurs in any of the tests above, then `using Turing` which happens in `test/turing/runtests.jl` will cause a namespace clash, e.g. `Turing.NUTS` and `AdvancedHMC.NUTS`. This in turn leads to errors in the downstream tests since they expect `NUTS` to be available in the namespace + it refers to `Turing.NUTS`. Putting these in a module avoids this.

test/runtests.jl
torfjelde: Sure, I was just following the rest of the Turing-ecosystem, but I'm whatever about it :+1:

xukai92: Yeah I was worrying about potential conflicts with other testing suites so I thought an explict name might be helpful.

test/runtests.jl
xukai92: The namespace issue is handled from Line 58 right?
I'm referring to the codes before that.

torfjelde: Oh you're referring to the `try` block? It's to handle the case where we bump the minor or major version of AHMC, which then won't be compatible with Turing. In those cases we just have to go "np, it's a breaking change, so we're going to make changes to Turing anyways".

xukai92:
Is it expected that some of the CI runs fail?

torfjelde:
> Is it expected that some of the CI runs fail?

Yes this is why I made them separate. CI.yml should never fail, but IntegrationTests.yml are allowed to if it breaks Turing.jl. So the current master isn't compatible with Turing.jl, which is as expected.

xukai92:
Cool. 
Ideally we want something like "allow failure" for them but it seems that GitHub Actions doesn't support this feature (after a quick search).
I think this is fine for now as long as we (reviewers) know `IntegrationTests` is the one that allows to fail.

xukai92 (approved):

PR #276, opened 2021-08-27
Missing integrated tests for StrictGeneralisedNoUTurn

xukai92:
Looks like we missed `StrictGeneralisedNoUTurn` in the integrated test


torfjelde (approved):
Ah, great!:)

PR #284, opened 2022-01-19
Minor fix: missing `rng` for `init_params`

theogf:


cpfiffer (commented):
Could you increment the patch number as well?

xukai92 (dismissed):

sethaxen:
It would be nice also to have a test that would have broken before but now passes.

theogf:
@sethaxen Do you have something in mind?

xukai92:
I think it could be as simple as sampling twice with the same `rng` (for a very short chain) and check if the samples are exactly the same.

sethaxen:
I agree, we should have had that test anyways.

sethaxen (approved):
LGTM!

theogf:
Sorry the test I added was incorrect, it should be fixed now

test/abstractmcmc.jl
```diff
@@ -35,4 +35,23 @@ include("common.jl")
...
+    );
+    samples2 = AbstractMCMC.sample(
+        rng2, model, κ, metric, adaptor, 10;
+        nadapts = 0,
+        progress=false,
+        verbose=false
+    );
+    @test mapreduce(*, samples1, samples2) do s1, s2
+        s1.z.θ == s2.z.θ
+    end # Equivalent to using all, check that all samples are equal
```
sethaxen, line 56: Why not
```suggestion
    @test all(s1.z.θ .== s2.z.θ)
```
?

theogf: This would not work no?
And I thought my solution was prettier than 
```
@test all(getindex.(getindex.(samples1, :z), :\theta)) .== getindex.(getindex.(samples2, :z), :\theta))
```

theogf:
@sethaxen @xukai92 Is there anything missing from this PR?

xukai92 (approved):

PR #287, opened 2022-04-06
refactor: use ReTest

xukai92:
This PR replaces `Test` with `ReTest` (https://juliatesting.github.io/ReTest.jl/stable/) such that we can easily run a subset of tests.

Example
```
julia --project=@. -e 'using Pkg; Pkg.test(; test_args=ARGS)' "UnitEuclideanMetric/Leapfrog/Trajectory{MultinomialTS,FixedNSteps}/NoAdaptation"
```

Other benefits of `ReTest`
- It provides a better test summary with indication on which testsets miss a test.
- It provides a possible way to get rid of the hack to run AHMC tests vs Turing tests, but I will leave it for another PR.

Other changes include
- Removed `Distributed` (and related uses of `module`)---we actually never run the tests in a distributed manner and ReTest also supports a much simple way to handle this.
- Changed the default `AHMC_TEST_GROUP` to `AdvancedHMC`
- Added a simple CMD interface such that one can do `julia --projects=@. tests/runtest.jl "TestName"` to run dedicated tests.
- [Disable CI for `julia-nightly` on Windows](https://github.com/TuringLang/AdvancedHMC.jl/pull/287/commits/958ae572039292771e85c024e8b360399b1a3f71)


xukai92:
@yebai This is how the summary looks for test cases with no test:
image
Note the yellow/orange zeros.

yebai:
Hi, @xukai92 it looks good!

PR #294, opened 2022-09-06
Release new version

devmotion:
It seems the latest compat updates of SetFields and DocStringExtensions are not released yet.


yebai:
thanks @devmotion - since there are no changes here, I merged the PR manually.

PR #299, opened 2022-10-18
Introduce type for kinetic energy

xukai92:


src/contrib/ad.jl
```diff
@@ -1,9 +1,9 @@
 const ADSUPPORT = (:ForwardDiff, :Zygote)
 const ADAVAILABLE = Dict{Module, Function}()
 
-Hamiltonian(metric::M, ℓπ::T, m::Module) where {M<:AbstractMetric,T} = ADAVAILABLE[m](metric, ℓπ)
+Hamiltonian(metric::AbstractMetric, ℓπ::Function, m::Module) = ADAVAILABLE[m](metric, ℓπ)
```
sethaxen, line 4: Do we need this constraint to be a `Function`? This would prevent use of callables.

yebai:
@xukai92 some CI tests are failing - can you take a look?

src/contrib/ad.jl
yebai: It would be good to keep this more generic to support callables.

xukai92: I wasn’t aware of the callable use cases---will revert it.

yebai (approved):

PR #300, opened 2022-11-15
feat: relativistic HMC

xukai92:


src/AdvancedHMC.jl
```diff
@@ -30,7 +30,20 @@ include("utilities.jl")
 # r: momentum variables
 # z: phase point / a pair of θ and r
 
-include("kinetic.jl")
+# TODO Move it back to hamiltonian.jl after the rand interface is updated
+abstract type AbstractKinetic end
+
+struct GaussianKinetic <: AbstractKinetic end
+
+struct RelativisticKinetic{T} <: AbstractKinetic
```
yebai, line 38: Please consider keeping all these types and functions for relativistic HMC in `experimental/relativistic_hmc.jl`, so we have a clear separation of new research-lead code from the stable code.

src/hamiltonian.jl
```diff
@@ -34,9 +34,20 @@ function ∂H∂θ(h::Hamiltonian, θ::AbstractVecOrMat)
...
 end
 
-∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric}, r::AbstractVecOrMat) = copy(r)
-∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric}, r::AbstractVecOrMat) = h.metric.M⁻¹ .* r
-∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric}, r::AbstractVecOrMat) = h.metric.M⁻¹ * r
+∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric, <:GaussianKinetic}, r::AbstractVecOrMat) = copy(r)
+∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric, <:GaussianKinetic}, r::AbstractVecOrMat) = h.metric.M⁻¹ .* r
+∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric, <:GaussianKinetic}, r::AbstractVecOrMat) = h.metric.M⁻¹ * r
+
+function ∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric,<:RelativisticKinetic}, r::AbstractVecOrMat)
```
yebai, line 41: Similiarly, let us keep all these new algorithms under `experimental/relativistic_hmc.jl`

src/hamiltonian.jl
```diff
@@ -109,39 +120,60 @@ neg_energy(z::PhasePoint) = z.ℓπ.value + z.ℓκ.value
...
-    h::Hamiltonian{<:DenseEuclideanMetric},
+    h::Hamiltonian{<:DenseEuclideanMetric, <:GaussianKinetic},
     r::T,
     θ::T
 ) where {T<:AbstractVecOrMat}
     mul!(h.metric._temp, h.metric.M⁻¹, r)
     return -dot(r, h.metric._temp) / 2
 end
 
+# RelativisticKinetic
```
yebai, line 158: Same here.

src/metric.jl
```diff
@@ -106,13 +106,43 @@ end
...
     metric::DenseEuclideanMetric{T},
-    kinetic,
+    kinetic::GaussianKinetic,
 ) where {T}
     r = randn(rng, T, size(metric)...)
     ldiv!(metric.cholM⁻¹, r)
     return r
 end
 
+using AdaptiveRejectionSampling: RejectionSampler, run_sampler!
```
yebai, line 116: Same here.

test/hamiltonian.jl
```diff
@@ -12,6 +12,8 @@ include("common.jl")
     h1 = Hamiltonian(metric, f, g)
     h2 = Hamiltonian(metric, GaussianKinetic(), f, g)
     @test h1 == h2
+    h3 = Hamiltonian(metric, RelativisticKinetic(1.0, 1.0), f, g)
```
yebai, line 15: Let's keep these tests under `experimental/tests/runtests.jl`. We can then add this script to a new CI action.

Project.toml
```diff
@@ -4,6 +4,7 @@ version = "0.3.6"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
+AdaptiveRejectionSampling = "c75e803d-635f-53bd-ab7d-544e482d8c75"
```
yebai, line 7: I am not sure about this dependency, but don't want to be blocked by this at this stage. We can revisit this later.

src/AdvancedHMC.jl
```diff
@@ -30,7 +30,16 @@ include("utilities.jl")
...
-include("kinetic.jl")
+# TODO Move it back to hamiltonian.jl after the rand interface is updated
+abstract type AbstractKinetic end
+
+struct GaussianKinetic <: AbstractKinetic end
+
+export GaussianKinetic
+
+include("experimental/relativistic_hmc.jl")
+export RelativisticKinetic
```
yebai, line 41: Let's not export this, by default. 
```suggestion
```

test/experimental/relativistic_hmc.jl
```diff
@@ -0,0 +1,10 @@
+using ReTest, AdvancedHMC
+using LinearAlgebra: dot
+
+@testset "Hamiltonian" begin
+    f = x -> dot(x, x)
+    g = x -> 2x
+    metric = UnitEuclideanMetric(10)
+    h = Hamiltonian(metric, RelativisticKinetic(1.0, 1.0), f, g)
```
yebai, line 8:
```suggestion
    h = Hamiltonian(metric, AdvancedHMC.RelativisticKinetic(1.0, 1.0), f, g)
```

test/experimental/relativistic_hmc.jl
```diff
@@ -0,0 +1,10 @@
...
+    @test h.kinetic isa RelativisticKinetic
```
yebai, line 9:
```suggestion
    @test h.kinetic isa AdvancedHMC.RelativisticKinetic
```

yebai (dismissed):
Thanks, @xukai92. I left a few addition minor comments below. This PR should be ready to merge once all tests are passed.

xukai92:
Thanks for the review. I just recalled that we were intended to create a new branch for experimental stuff instead of merging it to the main branch. I will create a new PR for this instead and fix the CI.

yebai:
There are pros and cons for maintaining two branches -- BUG fixes and improvements will take a lot longer to propagate to the main branch. We observed that trend on `DynamicPPL`. 

 It's ok to merge this PR to the master branch I think. The new algorithm is already separated from the stable code, which is the main reason for a new branch.

xukai92:
I’m a bit concerned about the new dependency though. It makes no sense for all upstream packages to depend on a package that is not well maintained (last update was like 2 years ago). I could make it optional using Revise.jo. How do you think?

src/AdvancedHMC.jl
```diff
@@ -30,7 +30,16 @@ include("utilities.jl")
...
 
-include("kinetic.jl")
+# TODO Move it back to hamiltonian.jl after the rand interface is updated
+abstract type AbstractKinetic end
+
+struct GaussianKinetic <: AbstractKinetic end
+
+export GaussianKinetic
+
+include("experimental/relativistic_hmc.jl")
```
yebai, line 40: How about only loading experimental algorithms manually to avoid any additional dependency? For tests on CI, we can explicitly call `include("experimental/relativistic_hmc.jl")` from the test script. 
```suggestion
```

torfjelde:
IMO a different branch seems like the best option given the introduction of a new dep :confused:

xukai92:
@yebai I second to @torfjelde. Adding that dependency is a bad idea and managing it using Requires.jl is a bit cumbersome too. Since this is mostly for experimental features, I think we could just create a new branch called `experimental`, which we manually sync with the master. We just need to make sure we only introduce code change in the `src/experimental` folder; all other changes should come from merging with master.

yebai:
> @yebai I second to @torfjelde. Adding that dependency is a bad idea and managing it using Requires.jl is a bit cumbersome too. Since this is mostly for experimental features, I think we could just create a new branch called experimental, which we manually sync with the master. We just need to make sure we only introduce code change in the src/experimental folder; all other changes should come from merging with master.

Sure, that sounds good. Perhaps call that branch `hmc-research` instead?

xukai92:
> Sure, that sounds good. Perhaps call that branch `hmc-research` instead?

Sounds good. I will do this.

> EDIT: in that case, let's perhaps separate the changes to Kinect types into a new PR against master.

Do you mean undo the already merged PR (#299) in master?

src/AdvancedHMC.jl
```diff
@@ -30,7 +30,13 @@ include("utilities.jl")
 # r: momentum variables
 # z: phase point / a pair of θ and r
 
-include("kinetic.jl")
+# TODO Move it back to hamiltonian.jl after the rand interface is updated
+abstract type AbstractKinetic end
```
yebai, line 34: I am referring these changes -- they should be merged into `master`.

PR #301, opened 2022-11-17
Use LogDensityProblems.jl

torfjelde:
Turing.jl has somewhat "recently" started using `LogDensityProblems.jl` under the hood to simply all the handling of AD. This PR does the same for AdvancedHMC.jl, which in turn means that we can just plug a `Turing.LogDensityFunction` into AHMC and sample, in theory making the `HMC` implementation, etc. in Turing.jl redundant + we get additional AD-backends for free, e.g. Enzyme.

IMO we might even want to consider making a bridge-package between `AbstractMCMC.jl` and `LogDensityProblems.jl` simply containing this `LogDensityModel.jl`, given how this implementation would be very useful in other packages implementing samplers, e.g. `AdvancedMH.jl`. Thoughts on this @devmotion @yebai @xukai92 ?

A couple of (fixable) caveats:
- We're loosing the `ForwardDiff` implementation for `AbstractMatrix`.
- The `DiffResults.jl` helpers in `LogDensityProblems.jl` seems to be me a bit overly restrictive, e.g. https://github.com/tpapp/LogDensityProblems.jl/blob/a6a570751d0ee79345e92efd88062a0e6d59ef1b/src/DiffResults_helpers.jl#L14-L18 I believe will convert a `ComponentVector` into `Vector`, thus dropping the named dimensions. (@tpapp is there a reason why we can't just use `similar(x)`?)

EDIT: This now depends on https://github.com/TuringLang/AbstractMCMC.jl/pull/110.


tpapp:
@torfjelde: similar should work fine IMO. PRs to LogDensityProblems are most welcome and I will review them quickly.

yebai:
I fully support this idea. Also happy to consider making `AbstractMCMC` / `AbstractPPL` dependent on `LogDensityProblems.jl`.

yebai:
@tpapp would you keep `LogDensityProblems`'s dependency lightweight in the long term?

README.md
```diff
@@ -51,13 +51,19 @@ In this section we demonstrate a minimal example of sampling from a multivariate
   
 ```julia
 using AdvancedHMC, Distributions, ForwardDiff
+using LogDensityProblems
 using LinearAlgebra
 
+# Define the target distribution using the `LogDensityProblem` interface
+struct Problem
```
yebai, line 58:
```suggestion
struct LogTargetDensity
```

torfjelde:
> I fully support this idea. Also happy to consider making AbstractMCMC / AbstractPPL dependent on LogDensityProblems.jl.

The main argument against this is that LogDensityProblems.jl uses Requires.jl, and the idea behind AbstractMCMC is that it should be so lightweight that you should be able to just depend on it. This is why I'm in favour of a bridge-package (though I'm also not a big fan of the growing number of packages, but seems like it'll be difficult to find an alternative here).

README.md
```diff
@@ -51,13 +51,19 @@ In this section we demonstrate a minimal example of sampling from a multivariate
...
+end
+LogDensityProblems.logdensity(p::Problem, θ) = logpdf(MvNormal(zeros(p.dim), I), θ)
+LogDensityProblems.dimension(p::Problem) = p.dim
+
 # Choose parameter dimensionality and initial parameter value
 D = 10; initial_θ = rand(D)
-
-# Define the target distribution
-ℓπ(θ) = logpdf(MvNormal(zeros(D), I), θ)
+ℓπ = Problem(D)
```
yebai, line 66:
```suggestion
ℓπ = LogTargetDensity(D)
```

README.md
```diff
@@ -51,13 +51,19 @@ In this section we demonstrate a minimal example of sampling from a multivariate
...
 using AdvancedHMC, Distributions, ForwardDiff
+using LogDensityProblems
 using LinearAlgebra
 
+# Define the target distribution using the `LogDensityProblem` interface
+struct Problem
+    dim::Int
+end
+LogDensityProblems.logdensity(p::Problem, θ) = logpdf(MvNormal(zeros(p.dim), I), θ)
+LogDensityProblems.dimension(p::Problem) = p.dim
```
torfjelde, line 62:
```suggestion
LogDensityProblems.logdensity(p::LogTargetDensity, θ) = logpdf(MvNormal(zeros(p.dim), I), θ)
LogDensityProblems.dimension(p::LogTargetDensity) = p.dim
```

yebai:
It might be possible to move some core types and APIs from `LogDensityProblems` into `AbstractMCMC`, then make `LogDensityProblems` dependent on `AbstractMCMC`. But that requires @tpapp's support of course.

devmotion:
LogDensityProblems is much more lightweight than AbstractMCMC: The former depends on 4 packages (https://juliahub.com/ui/Packages/LogDensityProblems/wTQV3/1.0.2?page=1) whereas the latter has 7 direct and 35 indirect dependencies (https://juliahub.com/ui/Packages/AbstractMCMC/5x8OI/4.1.3?page=1). So I don't think it makes sense to make LogDensityProblems dependent on AbstractMCMC, the other way around seems more reasonable.

devmotion:
Already in https://github.com/TuringLang/Turing.jl/pull/1877 my plan was to move `Turing.LogDensityFunction` and Turing-specific AD stuff to separate packages. I just haven't had time to push it but I think it should not be part of Turing in the long run. For instance, to avoid redefining the AD parts and density functions etc. in other packages such as Bijectors, AdvancedVI etc. (and AdvancedHMC I assume).

src/AdvancedHMC.jl
```diff
@@ -49,6 +51,40 @@ export Trajectory, HMCKernel,
...
+    logdensity::L
+end
+
+LogDensityModel(logdensity, ad) = LogDensityModel(ADgradient(ad, logdensity))
+
+function LogDensityProblems.ADgradient(kind::Symbol, ℓ::LogDensityModel)
+    return LogDensityModel(LogDensityProblems.ADgradient(kind, ℓ.logdensity))
+end
+
+for kind in [:ForwardDiff, :ReverseDiff, :Zygote, :Tracker, :Enzyme]
```
devmotion, line 73: A bit annoying that this is hard-coded.

Do we actually need to implement the whole LogDensityProblems interface for the model? Is that supposed to be called by users? If not, we could just use and call whatever we need without making it a complete wrapper.

Similar to how in Turing we don't define any of this for `Model` but only (and only for own AD types!) functions such as `ADgradient` for the density.

tpapp:
@yebai: Yes, LogDensityProblems will be kept very light. I am not planning to add any more dependencies. It currently uses Requires.jl for loading backend-specific code for AD, I am hoping to replace it with AbstractDifferentiation.jl when that matures.

yebai:
@tpapp many thanks for confirming. Yes, I am slightly worried about depending on `Requires`.

src/AdvancedHMC.jl
torfjelde: I'm a bit uncertain what you mean by "is it supposed to be called by the user". Are you referring to `ADgradient`? If so, I would say we should allow it. IMO if we're buying into the LogDensityProblems.jl interface, we should adopt it properly so that the users can move between different packages, e.g. AdvancedHMC.jl and AdvancedMH.jl, without having to redefine their model (if they're not using a Turing model).

But I agree, it's annoying that it's hardcoded. Would even be nice if LogDensityProblems.jl just had a list of AD packages that it supports as a variable so we could do this programmatically. Or is that a bad idea? @tpapp

tpapp: 1. Generally, the intention is that users only call via the `LogDensityProblems` interface when debugging or optimizing code (eg profiling, benchmarking).

2. I would be happy to add a variable with the currently supported interfaces *for use in CI*.

tpapp: I opened https://github.com/tpapp/LogDensityProblems.jl/issues/96 for discussion.

devmotion:
I also think that AbstractDifferentation could resolve this but to me it seems still not mature enough. From the ouside (maybe someone knows it better) it seems that unfortunately nobody is pushing it right now and nobody is working on fixing the different issues it has. Possibly one reason for that state could be that it might require some design changes/decisions, and at least to me it's a bit unclear who's deciding what to do in the end.

src/AdvancedHMC.jl
torfjelde: > I would be happy to add a variable with the currently supported interfaces for use in CI.

So you're not a big fan of what I've done here? I basically want `LogDensityModel` to do nothing but propagate everything to the underlying implemented of the LogDensityProblems.jl interface. But to do this, without causing method ambiguities, I need to overload each of the `ADgradient` methods.

devmotion: My main point above was: Why does the model have to support the LogDensityProblems interface at all? When do users manually transform the model? For instance, currently in Turing we do not support these things at all for models - we only implemented them for the `Turing.LogDensityFunction`. Users specify their AD backend either globally or as part of the sampler, and then we handle everything internally.

devmotion: I think this also makes a bit of sense conceptually - AD choice is not really something model-specific but is a sampler-specific thing. And some samplers don't need to differentiate the log density function at all.

torfjelde:
I honestly think at this point I'd be happy to add the implementation of `LogDensityModel` from this PR to AbstractMCMC .jland depend on LogDensityProblems.jl. We've had plans for ages of adding some "simple"/default models, etc. to AbstractMCMC.jl which could be used by many of the sampler packages without them having to define their own models, _but_ these plans/discussions usually grind to a halt because we're never quite satisfied with the design. I think LogDensityProblems.jl does a good enough job for it to be worth just going with that approach for now. Then packages such as AdvancedHMC.jl and AdvancedMH.jl can both just use the LogDensityProblems.jl interface to interact with the model, which in turn has some nice implications:
1. Anything that implements `LogDensityProblems.logdensity`, `LogDensityProblems.dimension`, and `LogDensityProblems.capabilties` can be used, i.e. the user can use the same model across the different packages.
2. We can move the non-AD stuff for `Turing.LogDensityFunction` to DynamicPPL.jl.
3. We can easily apply samplers _not_ from Turing.jl to models from DynamicPPL.jl without any further changes (if the samplers start making use of the LogDensityProblems-interface).

(3) was actually my original intention for this PR. Using AdvancedHMC.jl directly gives me much better control of the sampler than the implementation of `HMC` present in Turing.jl, and after this PR I can just make a `Turing.LogDensityFunction` from my model and then AdvancedHMC.jl just works.

yebai:
@torfjelde that sounds like a sensible plan. I am happy to make `AbstractMCMC` depend on `LogDensityProblems`. 

For the slightly longer term, I hope that `LogDensityProblems` can separate the AD stuff into a separate package so `LogDensityProblems`  is truly lightweight.

tpapp:
@yebai: Yes, that is the plan. I created an issue for further discussion/suggestions about moving AD backends out of the package: https://github.com/tpapp/LogDensityProblems.jl/issues/97

yebai:
@tpapp my impression is that `AbstractDifferentiation` will likely take a very long time to mature if that happens at all. 

Would you be happy if @torfjelde helped to separate the AD code in `LogDensityProblems` into a new package now-ish? This would allow `AbstractMCMC` to depend on `LogDensityProblems` without introducing a dependency on `Requires` which we tried hard to avoid.

devmotion:
AdvancedHMC, Turing, etc. all depend on Requires anyway, don't they? But maybe it would be a good idea anyway to move them into separate (sub-)packages.

I think one other point is that it would be nice to be able to define AD backend types as well (maybe could be done even without hiding it behind Requires), similar to the backend types in AbstractDifferentiation and Turing: That would allow to pass the AD type and its option around in a somewhat clean and standardized way (cf. https://github.com/TuringLang/AdvancedHMC.jl/pull/301#discussion_r1029169062).

tpapp:
I separated the AD backend code to https://github.com/tpapp/LogDensityProblemsAD.jl. Once it registers, I will deprecate `LogDensityProblems.ADgradient` and remove code from there and the latter package will no longer depend on Requires.jl.

Further suggestions for AD backends are most welcome, but please open an issue there so they don't get lost.

src/AdvancedHMC.jl
torfjelde: > Generally, most samplers don't care about AD. It's mainly HMC and similar samplers, isn't it? So baking it into the model seems weird.

But it _also_ does make sense to go "the model is differentiable", no? And "baking the AD into the model" only occurs if, well, we _need_ to AD, not for _any_ model. 

> Just having a default model type that contains a log density function seems sufficient to be able to pass it around. The only requirement would be that the log density function (not the model!) supports the LogDensityProblems interface.

But this is what `LogDensityModel` from this PR achieves, no? This is literally just passing the log density function around but allowing us to also pass it through methods which require `AbstractMCMC.AbstractModel`.

torfjelde: > The reason for not using just Val(:ForwardDiff) etc. in Turing but defining our own AD backends is that in that way we can collect additional options such as chunk size, tag, etc. in a clean way and provide then these information to ADgradient. It also allows us to define a backend for e.g. Zygote without actually every loading or depending on Zygote. Maybe all these things could be resolved though if LogDensityProblems/AbstractDifferentiation would support these additional options. At least in Turing we could e.g. already use AbstractDifferentiation.ForwardDiffBackend instead of Turing.ForwardDiffAD - if it would support custom tags as well. We could then just call ADgradient(Val(:ForwardDiff), ...) in the same way but without our custom backend. 

This I'm 100% behind! I'm not arguing against this:)

> These backends are hidden behind a Requires block though in AbstractDifferentiation - maybe that should and could be changed?

Indeed. This (+ IIRC, lack of support for compiled ReverseDiff) is what stopped me the last time I looked into whether AbstractDifferentiation. The AD types need to be defined prior to the Requires block so we can actually dispatch on them in downstream packages without first having to load the corresponding AD package.

tpapp:
I have now factored out all the AD code from LogDensityProblems.  A new release (2.0) will happen as soon as the registry automerges (thanks for your patience, this has been a busy time for me so I only got around to it now).

I plan to use weak dependencies instead of Requires.jl in LogDensityProblemsAD. Follow https://github.com/tpapp/LogDensityProblemsAD.jl/issues/2 if you want to keep an eye on this.

test/common.jl
```diff
@@ -11,6 +11,16 @@ const DETATOL = 1e-3 * D * TRATIO
...
 
+# Convenience
+using Distributions: Distributions
+using Bijectors: Bijectors
+LogDensityProblems.dimension(d::Distributions.Distribution) = length(d)
+function LogDensityProblems.logdensity(d::Distributions.Distribution, y)
+    b = Bijectors.inverse(Bijectors.bijector(d))
+    x, logjac = Bijectors.with_logabsdet_jacobian(b, y)
+    return logpdf(d, x) + logjac
+end
```
torfjelde, line 22: This is maybe not the greatest idea :confused: 

Should I replace this with a "local" definition of `logdensity`?

test/abstractmcmc.jl
```diff
@@ -10,7 +10,7 @@ include("common.jl")
 
     θ_init = randn(rng, 2)
 
-    model = AdvancedHMC.DifferentiableDensityModel(ℓπ_gdemo, ForwardDiff)
+    model = AdvancedHMC.LogDensityModel(ℓπ_gdemo, ForwardDiff)
```
devmotion, line 13: This should be
```suggestion
    model = AdvancedHMC.LogDensityModel(ADgradient(Val(:forwarddiff), ℓπ_gdemo))
```
(or something similar, don't remember the API by heart), shouldn't it?

README.md
```diff
@@ -51,13 +51,19 @@ In this section we demonstrate a minimal example of sampling from a multivariate
...
 ```julia
 using AdvancedHMC, Distributions, ForwardDiff
+using LogDensityProblems
 using LinearAlgebra
 
+# Define the target distribution using the `LogDensityProblem` interface
+struct LogTargetDensity
+    dim::Int
+end
+LogDensityProblems.logdensity(p::LogTargetDensity, θ) = logpdf(MvNormal(zeros(p.dim), I), θ)
```
devmotion, line 61: More efficient alternatives would be
```suggestion
LogDensityProblems.logdensity(p::LogTargetDensity, θ) = logpdf(MvNormal(Zeros(p.dim), I), θ)
```
or
```suggestion
LogDensityProblems.logdensity(p::LogTargetDensity, θ) = logpdf(MvNormal(Eye(p.dim)), θ)
```
Both require `using FillArrays`.

test/common.jl
devmotion: No, I agree, I think we should remove this type piracy if possible. Where is it used?

README.md
torfjelde: Am aware of this but do we want to the introduce this additional "complexity"? I feel like if a user is familiar with FillArrays.jl, they'll immediately know to use that implementation instead, while if you're not then maybe this is not the place to familiarize yourself with it?

test/abstractmcmc.jl
torfjelde: Ah, good catch; thanks!

test/common.jl
torfjelde: Cool, will do that then. It's just used a couple of places in the tests.

torfjelde: Addressed it.

README.md
devmotion: Hmm maybe you're right. On the other hand, maybe the README would be a good opportunity to teach people that don't know these more efficient implementations.

src/AdvancedHMC.jl
```diff
@@ -139,9 +143,22 @@ include("sampler.jl")
...
-include("contrib/ad.jl")
+Hamiltonian(metric::AbstractMetric, ℓ::LogDensityModel) = Hamiltonian(
+    metric,
+    Base.Fix1(LogDensityProblems.logdensity, ℓ.logdensity),
+    Base.Fix1(LogDensityProblems.logdensity_and_gradient, ℓ.logdensity),
+)
+function Hamiltonian(metric::AbstractMetric, ℓπ::LogDensityModel, kind::Union{Symbol,Val})
+    ℓ = LogDensityModel(LogDensityProblemsAD.ADgradient(kind, ℓπ.logdensity))
+    return Hamiltonian(metric, ℓ)
+end
```
devmotion, line 155: Should this be removed? Or error if the density already has a gradient defined (can be checked with capabilities IIRC)? Otherwise we might override an analytical gradient.

src/AdvancedHMC.jl
```diff
@@ -139,9 +143,22 @@ include("sampler.jl")
...
+function Hamiltonian(metric::AbstractMetric, ℓπ, kind::Union{Symbol,Val})
+    ℓ = LogDensityModel(LogDensityProblemsAD.ADgradient(kind, ℓπ))
+    return Hamiltonian(metric, ℓ)
+end
+Hamiltonian(metric::AbstractMetric, ℓπ, m::Module) = Hamiltonian(metric, ℓπ, Val(Symbol(m)))
+Hamiltonian(metric::AbstractMetric, ℓπ) = Hamiltonian(metric, ℓπ, Val{:ForwardDiff}())
```
devmotion, line 161: A bit shorter:
```suggestion
function Hamiltonian(metric::AbstractMetric, ℓπ, kind::Union{Symbol,Val} = Val{:ForwardDiff}())
    ℓ = LogDensityModel(LogDensityProblemsAD.ADgradient(kind, ℓπ))
    return Hamiltonian(metric, ℓ)
end
Hamiltonian(metric::AbstractMetric, ℓπ, m::Module) = Hamiltonian(metric, ℓπ, Val(Symbol(m)))
```

test/adaptation.jl
```diff
@@ -142,8 +141,7 @@ end
                 Σ = m' * m
 
                 # Correlated Gaussian
-                target = MvNormal(zeros(D), Σ)
-                ℓπ = θ -> logpdf(target, θ)
+                ℓπ = LogDensityDistribution(MvNormal(zeros(D), Σ))
```
devmotion, line 144:
```suggestion
                ℓπ = LogDensityDistribution(MvNormal(Σ))
```

README.md
sethaxen: > On the other hand, maybe the README would be a good opportunity to teach people that don't know these more efficient implementations.

I don't think that should be a goal of this README. In fact, wouldn't it be simpler to just drop Distributions altogether and just use 
```suggestion
LogDensityProblems.logdensity(p::LogTargetDensity, θ) = -sum(abs2, θ) / 2  # standard multivariate normal
```
?

test/integrator.jl
```diff
@@ -106,8 +106,14 @@ using Statistics: mean
...
     end
 
-    @testset "Analyitcal solution to Eq (2.11) of Neal (2011)" begin
-        negU = q -> -dot(q, q) / 2
+    @testset "Analytical solution to Eq (2.11) of Neal (2011)" begin
+        struct NegU
+            dim::Int
+        end
+        (::NegU)(q) = -dot(q, q) / 2
+        LogDensityProblems.logdensity(d::NegU, x) = d(x)
```
devmotion, line 114:
```suggestion
        LogDensityProblems.logdensity(d::NegU, x) = -dot(x, x) / 2
```

README.md
torfjelde: > I don't think that should be a goal of this README.

Exactly, and in particular with `using` and exported functionality, a user might not even be able to identify what's from which package. 

> In fact, wouldn't it be simpler to just drop Distributions altogether and just use

Was thinking the same, but figured I'd leave it as is in an attempt to avoid this such a discussion to clog the PR :sweat_smile: But I'm happy to make that change if we want. @xukai92 you okay with that?

src/AdvancedHMC.jl
```diff
@@ -139,9 +143,22 @@ include("sampler.jl")
...
 
 include("abstractmcmc.jl")
-export DifferentiableDensityModel
 
-include("contrib/ad.jl")
+Hamiltonian(metric::AbstractMetric, ℓ::LogDensityModel) = Hamiltonian(
+    metric,
+    Base.Fix1(LogDensityProblems.logdensity, ℓ.logdensity),
+    Base.Fix1(LogDensityProblems.logdensity_and_gradient, ℓ.logdensity),
+)
```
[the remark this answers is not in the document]
torfjelde: Agreed:)

src/AdvancedHMC.jl
torfjelde: Hmm, though I agree with you, if the user actively calls `Hamiltonian(mertic, model, kind)` then the currently expected behavior is that you use `kind` for the gradient computation. That is, the user shouldn't really be calling `Hamiltonian(metric, model, kind)` if they don't want to actually use `kind` for the gradient computation, no? 

And IMO there are def cases where you might want to override a manual impl of the gradient, e.g. if you want to check your own impl vs. "truth" (AD).

test/common.jl
```diff
@@ -11,6 +11,20 @@ const DETATOL = 1e-3 * D * TRATIO
 # Random tolerance
 const RNDATOL = 5e-2 * D * TRATIO * 2
 
+# Convenience
+using Distributions: Distributions
+using Bijectors: Bijectors
+struct LogDensityDistribution{D<:Distributions.Distribution}
```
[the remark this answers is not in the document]
torfjelde: Yeaaah I was also thinking this would be nice to have somewhere it's more widely available.

But is AHMC really the right place? :confused:

test/common.jl
```diff
@@ -24,6 +38,9 @@ end
 
 ℓπ_gaussian(m, s, x) = ℓπ_gaussian(m .- x, s)
 
+LogDensityProblems.dimension(g::Gaussian) = dim(g.m)
+LogDensityProblems.logdensity(g::Gaussian, x) = ℓπ_gaussian(g.m. g.s, x)
```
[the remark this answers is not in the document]
torfjelde: Probably but I didn't want to start messing around with that right now tbh.

src/AdvancedHMC.jl
devmotion: Seems reasonable.

test/common.jl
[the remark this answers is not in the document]
devmotion: Since it depends on Bijectors, maybe it could be defined in Bijectors?

torfjelde:
This is causing issues: https://github.com/tpapp/LogDensityProblemsAD.jl/blob/79e245e1bb8e904087af9de284e2796e51eb55c4/src/AD_ForwardDiff.jl#L21-L23

`LogDensityProblemsAD.ADgradient` requires `ForwardDiff.GradientConfig` and so the type of the dual vector will be tied to the density problem.

devmotion:
Where/why is that problematic? There are no issues with that design in Turing (in fact, being able to specify the GradientConfig is very useful for working with custom tags: https://github.com/TuringLang/Turing.jl/blob/61b06f642872522ca14133bb5c86fc603841dbba/src/essential/ad.jl#L89-L100).

torfjelde:
It's not necessarily _problematic_, but it does mean this change will be breaking. Currently AHMC won't require specification of the gradient config to work with something like ComponentArray.

torfjelde:
Btw, @devmotion now I remembered why I didn't use the shorter definition
```julia
function Hamiltonian(metric::AbstractMetric, ℓπ, kind::Union{Symbol,Val} = Val{:ForwardDiff}())
    ℓ = LogDensityModel(LogDensityProblemsAD.ADgradient(kind, ℓπ))
    return Hamiltonian(metric, ℓ)
end
```

It overrides the default implementation
```julia
Hamiltonian(metric::AbstractMetric, ℓ::LogDensityModel) = Hamiltonian(
    metric,
    Base.Fix1(LogDensityProblems.logdensity, ℓ.logdensity),
    Base.Fix1(LogDensityProblems.logdensity_and_gradient, ℓ.logdensity),
)
```

:facepalm:

test/demo.jl
```diff
@@ -42,15 +48,31 @@ end
...
+    ℓπ_with_grad = LogDensityProblemsAD.ADgradient(
+        Val(:ForwardDiff),
+        ℓπ;
+        # NOTE: Have to manually construct the `gradientconfig` to ensure we're not using the default
+        # buffer in the config (which won't be of type `ComponentArray`).
+        gradientconfig = ForwardDiff.GradientConfig(
+            Base.Fix1(LogDensityProblems.logdensity, ℓπ),
+            similar(p1),
+            LogDensityProblemsAD._default_chunk(ℓπ),
+        )
```
devmotion, line 73: As you said, that's not convenient for users. Could we just add a method
```julia
Hamiltonian(metric, logdensity, ::Val{:ForwardDiff}; kwargs...)
```
and perform the gradient construction with the correct config there automatically? With the `kwargs` we could allow some customization by users.

Generally, I guess it would be useful to forward `kwargs` to `ADgradient` also for other backends.

devmotion: Generally, maybe that should also be fixed upstream: Similar to the ReverseDiff options (https://github.com/tpapp/LogDensityProblemsAD.jl/blob/79e245e1bb8e904087af9de284e2796e51eb55c4/src/AD_ReverseDiff.jl#L32), one could add a `x` keyword argument (default: `nothing`) that would allow to specify the type of input vector.

torfjelde: You can't specify the underlying type used in the `GradientConfig` using a `kwarg` for `ADgradient` though :confused: That is, we the "best" simplification we can do (without upstream changes) is

```julia
Hamiltonian(metric, logdensity, ::Val{:ForwardDiff}; gradientconfig)
```

and the user still has to construct `gradientconfig` as above.

We can of course also make some helper method for constructing `GradientConfig` for a particular type, or, maybe better, add an additional argument to `LogDensityProblemsAD.ADgradient` which is passed to `LogDensityProblemsAD._default_gradientconfig`.

devmotion: > maybe better, add an additional argument to LogDensityProblemsAD.ADgradient which is passed to LogDensityProblemsAD._default_gradientconfig.

That's what I would suggest :slightly_smiling_face: 

> one could add a x keyword argument (default: nothing) that would allow to specify the type of input vector.

However, even if that's not available yet - we can use in `Hamiltonian` whatever keyword arguments we want, can't we? So we could add such a `x` keyword argument at least there right away and then define a gradient config based on it.

torfjelde: > That's what I would suggest

https://github.com/tpapp/LogDensityProblemsAD.jl/pull/3

> However, even if that's not available yet - we can use in Hamiltonian whatever keyword arguments we want, can't we? So we could add such a x keyword argument at least there right away and then define a gradient config based on it.

Yeah, but if @tpapp is okay with the PR me might as well just use that to avoid (potentially) more breaking changes, e.g. if we choose a diff kwarg than LogDensityProblemsAD does.

tpapp: I am OK with the PR, conditional on changes requested by @devmotion.

src/AdvancedHMC.jl
```diff
@@ -139,9 +143,21 @@ include("sampler.jl")
...
+function Hamiltonian(metric::AbstractMetric, ℓπ::LogDensityModel, kind::Union{Symbol,Val})
+    ℓ = LogDensityModel(LogDensityProblemsAD.ADgradient(kind, ℓπ.logdensity))
```
devmotion, line 153: Forward keyword arguments?
```suggestion
function Hamiltonian(metric::AbstractMetric, ℓπ::LogDensityModel, kind::Union{Symbol,Val}; kwargs...)
    ℓ = LogDensityModel(LogDensityProblemsAD.ADgradient(kind, ℓπ.logdensity; kwargs...))
```

src/AdvancedHMC.jl
```diff
@@ -139,9 +143,21 @@ include("sampler.jl")
...
+    return Hamiltonian(metric, ℓ)
+end
+function Hamiltonian(metric::AbstractMetric, ℓπ, kind::Union{Symbol,Val} = Val{:ForwardDiff}())
+    ℓ = LogDensityModel(LogDensityProblemsAD.ADgradient(kind, ℓπ))
```
devmotion, line 157:
```suggestion
function Hamiltonian(metric::AbstractMetric, ℓπ, kind::Union{Symbol,Val} = Val{:ForwardDiff}(); kwargs...)
    ℓ = LogDensityModel(LogDensityProblemsAD.ADgradient(kind, ℓπ; kwargs...))
```

src/AdvancedHMC.jl
```diff
@@ -139,9 +143,21 @@ include("sampler.jl")
...
+    return Hamiltonian(metric, ℓ)
+end
+Hamiltonian(metric::AbstractMetric, ℓπ, m::Module) = Hamiltonian(metric, ℓπ, Val(Symbol(m)))
```
devmotion, line 160:
```suggestion
Hamiltonian(metric::AbstractMetric, ℓπ, m::Module; kwargs...) = Hamiltonian(metric, ℓπ, Val(Symbol(m)); kwargs...)
```

test/demo.jl
```diff
@@ -42,15 +48,31 @@ end
...
+    )
+    hamiltonian = Hamiltonian(metric, AbstractMCMC.LogDensityModel(ℓπ_with_grad))
```
devmotion, line 75: Requires LogDensityProblemsAD 1.1.0 but only in the tests it seems:
```suggestion
    hamiltonian = Hamiltonian(metric, ℓπ, Val(:ForwardDiff); x = p1)
```

README.md
torfjelde: I made the change suggested by @sethaxen

test/common.jl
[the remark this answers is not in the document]
torfjelde: I'm a bit uncertain if it's worth including since it's of little practical usage (beyond testing samplers). For now I've added a comment for future reference, but leaving it as is for now.

torfjelde:
This PR should now be good to go as soon as tests pass:)

src/AdvancedHMC.jl
```diff
@@ -139,9 +143,21 @@ include("sampler.jl")
 export sample
 
 include("abstractmcmc.jl")
-export DifferentiableDensityModel
 
-include("contrib/ad.jl")
+Hamiltonian(metric::AbstractMetric, ℓ::LogDensityModel) = Hamiltonian(
+    metric,
+    Base.Fix1(LogDensityProblems.logdensity, ℓ.logdensity),
+    Base.Fix1(LogDensityProblems.logdensity_and_gradient, ℓ.logdensity),
```
devmotion, line 150: One final suggestion: I guess it could be useful to check here if the gradient is actually defined with the LogDensityProblems API and in case it is missing one could throw an error with an explanation that e.g. one can choose an AD backend by using the other constructor.

src/AdvancedHMC.jl
```diff
@@ -139,9 +143,21 @@ include("sampler.jl")
...
+)
```
devmotion, line 151:
```suggestion
function Hamiltonian(metric::AbstractMetric, ℓ::LogDensityModel)
    ℓπ = ℓ.logdensity
    cap = LogDensityProblems.capabilities(ℓπ)
    if cap === nothing
        throw(ArgumentError("The log density function does not support the LogDensityProblems.jl interface"))
    end
    if cap === LogDensityProblems.LogDensityOrder{0}()
        throw(ArgumentError("The gradient of the log density function is not defined: Implement `LogDensityProblems.logdensity_and_gradient` or use automatic differentiation by calling `Hamiltionian(metric, logdensity, AD; kwargs...)` where AD is one of the backends supported by LogDensityProblemsAD.jl"))
    return Hamiltonian(
        metric,
        Base.Fix1(LogDensityProblems.logdensity, ℓ.logdensity),
        Base.Fix1(LogDensityProblems.logdensity_and_gradient, ℓ.logdensity),
    )
end
```

The first check should maybe moved to AbstractMCMC.

devmotion (approved):

PR #302, opened 2022-11-28
Relativistic HMC

xukai92:
I duplicated the current `master` as `hmc-research`. This PR is same as #300 but towards the new branch.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -33,6 +33,16 @@ if GROUP == "All" || GROUP == "AdvancedHMC"
     end
 end
 
+if GROUP == "All" || GROUP == "AdvancedHMC"
+    using ReTest
+
+    include(joinpath("experimental", "relativistic_hmc.jl"))
+
+    @main function runtests(patterns...; dry::Bool=false)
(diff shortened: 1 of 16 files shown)


yebai:
thanks, @xukai92 -- can we transfer the changes to `hamiltonian.jl` and `metric.jl` to #303 too?

test/runtests.jl
```diff
@@ -33,6 +33,16 @@ if GROUP == "All" || GROUP == "AdvancedHMC"
     end
 end
 
+if GROUP == "All" || GROUP == "Experimental"
```
yebai, line 36: Let's create a new file `test/experimental/runtests.jl` and transfer these test changes there. The idea is that on the `hmc-research` branch, we only change files under `src/experimental` and `tests/experimental`. Everything else should remain the same with the `master` branch.

xukai92:
> can we transfer the changes to `hamiltonian.jl` and `metric.jl` to https://github.com/TuringLang/AdvancedHMC.jl/pull/303 too?

Done.

PR #304, opened 2022-11-30
Moved all experimental code including tests into research folder. 

yebai:


research/src/relativistic_hmc.jl
```diff
@@ -1,3 +1,5 @@
+using AdvancedHMC
```
xukai92, line 1: The CI is breaking as we will need to `using` internal names within `AdvancedHMC` inside here, such as `using AdvancedHMC: AbstractKinetic`. And potentially for all other names that we do not export but actually need to use or extend.

xukai92:
@yebai If we want to do this setup, we will need to carefully `using` internal names from AHMC. The CI is broken because of this; see more at https://github.com/TuringLang/AdvancedHMC.jl/pull/304#discussion_r1036362118.

xukai92:
Just to add my preference here: I feel this setup makes experimenting a bit hard as we need to do those extra management. In fact, if we want to separate the code like this, we may well just keep the research code in a separate repo. The main benefit of having it here is that we can change AHMC itself more in an invasive manner with an implementation that could be easily cherry picked into the main branch.

research/src/relativistic_hmc.jl
yebai, line 1:
```suggestion
using AdvancedHMC
import AdvancedHMC: ∂H∂r, neg_energy, AbstractKinetic
```

yebai:
> 
Just to add my preference here: I feel this setup makes experimenting a bit hard as we need to do those extra management. In fact, if we want to separate the code like this, we may well just keep the research code in a separate repo. The main benefit of having it here is that we can change AHMC itself more in an invasive manner with an implementation that could be easily cherry picked into the main branch.

The idea is to use a standalone Julia environment (project) fro each experimental project. These environments manage their own dependencies like the `tests` folder. In addition, we can work with one branch `master` for research code too, since the algorithms in the `research` folder is not loaded by default: users have to explicitly instantiate them.

test/runtests.jl
```diff
@@ -34,7 +34,12 @@ if GROUP == "All" || GROUP == "AdvancedHMC"
 end
 
 if GROUP == "All" || GROUP == "Experimental"
-    include(joinpath("experimental", "runtests.jl"))
+    using Pkg
+     # activate separate test environment
+     Pkg.activate(joinpath(DIRECTORY_AdvancedHMC, "research/tests"))
```
yebai, line 39:
```suggestion
     Pkg.activate(joinpath(DIRECTORY_AdvancedHMC, "research"))
```

yebai:
FYI: https://github.com/JuliaLang/julia/pull/47695 will make it possible to have conditional dependency soon, so keeping everything on the `master` branch will be possible without separate environments for the research folder -- it also means we can load the `research` folder by default. So having a clear separation is now only for clarity rather than avoiding unnecessary dependencies needed by research code in the stable releases.

PR #305, opened 2022-12-07
Implement Riemannian HMC

xukai92:
This is a draft to implement Riemannian HMC. There are many things to discuss. I put the high-level points here while leaving more specific ones in the code.
- Do we want to unify all function signatures for Hamiltonian to a non-separable one (i.e. position dependent)?
- Do we need to update our metric abstraction? Looks like we should decouple unit/diagonal/dense vs Euclidean/Riemannian. 
- Where should SoftAbs best locate? I current implemented it as an external function to provide `G` to a generic Riemannian HMC implementation. 
- The efficiency is pretty bad and there is fewer optimization opportunities due to our decoupled abstraction (e.g. compared to what derived in [1]).

To-dos
- The current SoftAbs implementation has numerical issues when running on Neal's funnel. I need to double check the if the manual Jacobian implementation is correct, or we could register `softabs` with ReverseDiff to see if this solves the issue. I could use someone else help on this.

### How to play with this PR
I provided a notebook (which contains the same content as the `test/experimental/riemannian_hmc.jl` file) to play with the code. The notebook has some simple validation on the implementation and also shows the current numerical issue of SoftAbs. *I highly recommend you to try this.*

---
[1] Betancourt, M., 2013, August. A general metric for Riemannian manifold Hamiltonian Monte Carlo. In International Conference on Geometric Science of Information (pp. 327-334). Springer, Berlin, Heidelberg.


src/experimental/riemannian_hmc.jl
```diff
@@ -0,0 +1,268 @@
+### integrator.jl
```
xukai92, line 1: I left a note for each code chunk on where it should if this was implemented in AHMC directly.

src/experimental/riemannian_hmc.jl
```diff
@@ -0,0 +1,268 @@
...
+    res = if FullTraj
+        Vector{P}(undef, n_steps)
+    else
+        z
+    end
+
+    for i = 1:n_steps
+        θ_init, r_init = z.θ, z.r
+        # Tempering
+        #r = temper(lf, r, (i=i, is_half=true), n_steps)
```
xukai92, line 46: I commented out tempering to reduce extra complexity of this PR.

src/experimental/riemannian_hmc.jl
```diff
@@ -0,0 +1,268 @@
+### integrator.jl
+
+import AdvancedHMC: ∂H∂θ, ∂H∂r, DualValue, PhasePoint, phasepoint, step
+using AdvancedHMC: @unpack, TYPEDEF, TYPEDFIELDS, AbstractScalarOrVec, AbstractLeapfrog, step_size
+
+"""
+$(TYPEDEF)
+
+Generalized leapfrog integrator with fixed step size `ϵ`.
```
xukai92, line 9:
```suggestion
Generalized leapfrog integrator with fixed step size `ϵ` and number of fixed-point iteration `n`.
```

src/experimental/riemannian_hmc.jl
```diff
@@ -0,0 +1,268 @@
...
+refresh(
+    rng::Union{AbstractRNG, AbstractVector{<:AbstractRNG}},
+    ref::PartialMomentumRefreshment,
+    h::Hamiltonian,
+    z::PhasePoint,
+) = phasepoint(h, z.θ, ref.α * z.r + sqrt(1 - ref.α^2) * rand(rng, h.metric, h.kinetic, z.θ))
+
+# To change L146 of metric.jl
+# Ignore θ by default (i.e. not position-dependent)
+Base.rand(rng::AbstractRNG, metric::AbstractMetric, kinetic, θ) = _rand(rng, metric, kinetic)    # this disambiguity is required by Random.rand
```
xukai92, line 124: This is what I mean by making everything position-dependent by default.

src/experimental/riemannian_hmc.jl
```diff
@@ -0,0 +1,268 @@
...
+
+Base.rand(rng::AbstractRNG, metric::AbstractRiemannianMetric, kinetic, θ) = _rand(rng, metric, kinetic, θ)
+Base.rand(rng::AbstractVector{<:AbstractRNG}, metric::AbstractRiemannianMetric, kinetic, θ) = _rand(rng, metric, kinetic, θ)
+
+### hamiltonian.jl
+
+import AdvancedHMC: phasepoint, neg_energy, ∂H∂θ, ∂H∂r
+using LinearAlgebra: logabsdet, tr
+
+# QUES Do we want to change everything to position dependent by default?
```
xukai92, line 186: See my QUES for another position-dependent change

src/experimental/riemannian_hmc.jl
```diff
@@ -0,0 +1,268 @@
...
+function neg_energy(
+    h::Hamiltonian{<:DenseRiemannianMetric},
+    r::T,
+    θ::T
+) where {T<:AbstractVecOrMat}
+    G = h.metric.G(θ)
+    D = size(G, 1)
+    # Need to consider the normalizing term as it is no longer same for different θs
+    lad, s = logabsdet(G)
+    if s == -1
```
xukai92, line 206: I added this trick to reject when the determinant of G is negative.

src/experimental/riemannian_hmc.jl
```diff
@@ -0,0 +1,268 @@
...
+            ∂G∂θᵢ = ∂G∂θ[:,:,i]
+            ∂ℓπ∂θ[i] - 1 / 2 * tr(invG * ∂G∂θᵢ) + 1 / 2 * r' * invG * ∂G∂θᵢ * invG * r
+        end,
+    )
+end
+
+function ∂H∂r(h::Hamiltonian{<:DenseRiemannianMetric}, θ::AbstractVecOrMat, r::AbstractVecOrMat)
+    G = h.metric.G(θ)
+    # return inv(G) * r
+    return G \ r
```
xukai92, line 236: I change some `inv(G) * r` to `G \ r` just so it's more numerically stable.

test/experimental/riemannian_hmc.jl
```diff
@@ -0,0 +1,132 @@
...
+    
+    G = x -> hess_func(x)[3] + hps.λ * I
+    _∂G∂θ = gen_∂H∂x(neg_ℓπ, initial_θ)
+    ∂G∂θ = x -> reshape_J(copy(_∂G∂θ(x)))
+    
+    G_softabs = x -> softabs(hess_func(x)[3] + hps.λ * I, hps.α)
+    # NOTE The implementation below attempts to diff through the softabs function which involves eigen that is not implemented in RD
+    #_∂G∂θ_softabs = gen_∂H∂x(neg_ℓπ, initial_θ; f=(X -> softabs(X + hps.λ * I, hps.α)))
+    _∂G∂θ_softabs = gen_∂H∂x(neg_ℓπ, initial_θ)
+    ∂G∂θ_softabs = x -> make_J(G_softabs(x), hps.α) .* reshape_J(copy(_∂G∂θ_softabs(x)))
```
xukai92, line 59: I'm not sure if this is correct.

test/experimental/riemannian_hmc.jl
```diff
@@ -0,0 +1,132 @@
...
+    _hess_func = VecTargets.gen_hess(neg_ℓπ, initial_θ) # x -> (value, gradient, hessian)
+    hess_func = x -> copy.(_hess_func(x))
+    
+    G = x -> hess_func(x)[3] + hps.λ * I
+    _∂G∂θ = gen_∂H∂x(neg_ℓπ, initial_θ)
+    ∂G∂θ = x -> reshape_J(copy(_∂G∂θ(x)))
+    
+    G_softabs = x -> softabs(hess_func(x)[3] + hps.λ * I, hps.α)
+    # NOTE The implementation below attempts to diff through the softabs function which involves eigen that is not implemented in RD
+    #_∂G∂θ_softabs = gen_∂H∂x(neg_ℓπ, initial_θ; f=(X -> softabs(X + hps.λ * I, hps.α)))
```
xukai92, line 57: See the in-line comment above.

torfjelde:
One thing that is worth discussing wrt. this PR is how much of this functionality we should put into the metric types themselves vs. just implementing it all as a change of variables.

If we instead consider these position-dependent metrics as a change of variables for the potential instead, we don't have to deal with non-separable Hamiltonians internally and hence should hopefully make hte internals simpler. 

Are there any obvious downsides to this?

xukai92:
> One thing that is worth discussing wrt. this PR is how much of this functionality we should put into the metric types themselves vs. just implementing it all as a change of variables.
> If we instead consider these position-dependent metrics as a change of variables for the potential instead, we don't have to deal with non-separable Hamiltonians internally and hence should hopefully make hte internals simpler.
> Are there any obvious downsides to this?

I also think so as I put a note here: https://github.com/TuringLang/AdvancedHMC.jl/blob/master/src/metric.jl#L116.
Ideally we could interface Bijectors.jl here and do more than just affine transformation.

With this being said, the best route to implement this in AHMC is to do an internal refactoring PR first and then implement here. I feel that would require quite a bit of discussion.

Besides, I think most of the callback change to handle non-separable Hamiltonians that introduced in this PR is still needed there.

torfjelde:
> I also think so as I put a note here

Ah didn't catch that; cool! Btw, are we aware of any downsides to just doing a reparameterization for something like RHMC, given that the original paper didn't do this but instead went the route of Genearlized Leapfrog, which requires a root-solve in to determine the implictly defined step?

xukai92:
> Btw, are we aware of any downsides to just doing a reparameterization for something like RHMC, given that the original paper didn't do this but instead went the route of Genearlized Leapfrog, which requires a root-solve in to determine the implictly defined step?

I don't quite get the question. What the original paper does can also be seen as a form of reparameterization and we are doing the same thing together with genearlized Leapfrog, for which the implict updates are done by fixed-point iterations.

xukai92:
Updates as of ed15539
1. Introduced types to control how Hessian is mapped (https://github.com/TuringLang/AdvancedHMC.jl/blob/kx/rhmc-draft/research/src/riemannian_hmc.jl#L136-L155). This gives an easy way to access internal parameters of the SoftAbs map and specialize the internals based on different maps (`IdentityMap` or `SoftAbsMap`). 
    - PS: I made this choice mostly because of the easiness of implementation and I'm happy to consider other designs.
2. Fixed a bug in constructing `J` (https://github.com/TuringLang/AdvancedHMC.jl/blob/kx/rhmc-draft/research/src/riemannian_hmc.jl#L258) by adding a missing coefficient to `csch`
    - With 1 and 2, I refactored `∂H∂θ` for the SoftAbs metric (https://github.com/TuringLang/AdvancedHMC.jl/blob/kx/rhmc-draft/research/src/riemannian_hmc.jl#L271-L300) and the sampling results visually look OK on Neal's funnel now:
image
3. The incorrect way of handling numerical issue is removed and any handling is done by users (e.g. https://github.com/TuringLang/AdvancedHMC.jl/blob/kx/rhmc-draft/research/tests/riemannian_hmc.jl#L40)
4. Performed Geweke test for RHMC (), which indicates that there are still bugs in the RHMC sampler with SoftAbs metric:
image
The test is done by the model below
```julia
@model function TuringFunnel(θ=missing, x=missing)
    if ismissing(θ)
        θ = Vector(undef, 2)
    end
    θ[1] ~ Normal(0, 3)
    s = exp(θ[1] / 2)
    θ[2] ~ Normal(0, s)
    x ~ Normal(0, s)
    return θ, x
end
```
with [MCMCDebugging.jl](https://github.com/TuringLang/MCMCDebugging.jl).

PS: For reference, the test results of a normal HMC sampler under the same model is below:
image

I'm looking into the codes to see where the bug might be but if anyone has some ideas for debugging or intuition of where the bug might be, please let me know.

xukai92:
Note of meeting with Hong on 23 Feb

Potential optimization
- `G(θ)` is unchanged in the fixed-point iteration for (16). We don't need to compute it for every call of `∂H∂θ`.
- The first call of `∂H∂r` is unchanged during (17). We can put it outside of the loop.

Tests
- Use finite differentiation to test our Hessian/metric computation and partial derivatives
- Test `neg_energy` using MvNormal
    - Double check the usage

PR #307, opened 2022-12-29
Improve Hamiltonian constructors

devmotion:
This PR proposes some changes of the `Hamiltonian` constructors:
- For log density functions that already support computation of the gradient as well (via `logdensity_and_gradient`) an AD backend is not used automatically but only if explicitly specified
- `LogDensityModel`s are handled in the same way as if only their log density would have been provided, ie, in particular ForwardDiff is used automatically if the log density does not support computation of the gradient

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,7 @@ StatsFuns = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 UnPack = "3a884ed6-31ef-47d7-9d2a-63182c4928ed"
 
 [compat]
-AbstractMCMC = "4.2.1"
+AbstractMCMC = "4.2"
 ArgCheck = "1, 2"
 DocStringExtensions = "0.8, 0.9"
 InplaceOps = "0.3"
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "AdvancedHMC"
 uuid = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
-version = "0.4"
+version = "0.4.1"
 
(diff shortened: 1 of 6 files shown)


yebai (approved):

torfjelde:
Sorry, but isn't this a breaking change? Prior to this PR, you didn't _have_ to implement the LogDensityProblems-interface; you could still just provide a callable that represented the logdensity. It seems like this is no longer possible?

devmotion:
No, I don't think so. Already prior to this PR in AdvancedHMC 0.4, `Hamiltonian(metric, lp)` required that `lp` supported `LogDensityProblems.logdensity`. Arbitrary callables were only supported in the `Hamiltonian(metric, lp, lp_and_grad)` constructor which is not affected by this PR.

The main changes in this PR were:
- Handling `model::LogDensityModel` in exactly the same way as if `model.logdensity` would have been provided
  - previously `model.logdensity` required support of gradient evaluation whereas it was fine to provide a "vanilla" log density without gradient capabilities
  - previously only `Hamiltonian(metric, model::LogDensityModel, ::Union{Val,Symbol})` was supported whereas `Hamiltonian(metric, logdensity, ::Union{Val,Symbol,Module})` also supported `Module`s
- Checking if the log density function actually supports the LogDensityProblems interface with `LogDensityProblems.capabilities` to provide more meaningful error messages.

In principle, in a follow-up PR we could re-add support for callables which was removed in AdvancedHMC 0.4: If `LogDensityProblems.capabilities(logdensity) === nothing`, we could assume the provided `logdensity` is a function.

PR #309, opened 2023-01-02
Compatibility with MCMCChains

kaandocal:
This is a very minor pull request for conveniently bundling samples from `AdvancedHMC.jl` into an `Chains` object from `MCMCChains.jl`. Very similar to the interface in `AdvancedMH.jl`, it should make it easier to switch from one to the other.


devmotion:
I think ultimately we should try to improve the packages and interfaces such that downstream packages get this for free/automatically or by implementing a more minimal interface (at least don't have to depend on MCMCChains/Requires). Weak dependencies could help with this in Julia >= 1.9 but somehow it feels unsatisfying if similar code is added to every sampler package.

yebai (approved):
This PR looks useful. It's probably ok to merge this before we figure out a better interface in `AbstractMCMC`.

src/mcmcchains-connect.jl
```diff
@@ -0,0 +1,32 @@
...
+    state,
+    chain_type::Type{Chains};
+    discard_initial=0,
+    thinning=1,
+    param_names=missing,
+    bijector=identity,
+    kwargs...
+)
+    # Turn all the transitions into a vector-of-vectors.
+    vals = [vcat(bijector(t.z.θ), t.z.ℓπ.value) for t in ts]
```
sethaxen, line 17: It's important to return the other statistics as well:
```suggestion
    tstat = merge((; lp = t.z.ℓπ.value), stat(t))
    tstat_names = collect(keys(tstat))
    vals = [vcat(bijector(t.z.θ), collect(tstat)) for t in ts]
```

src/mcmcchains-connect.jl
```diff
@@ -0,0 +1,32 @@
...
+    if ismissing(param_names)
+        param_names = [Symbol(:param_, i) for i in 1:length(keys(ts[1].z.θ))]
+    else
+        # Generate new array to be thread safe.
+        param_names = Symbol.(param_names)
+    end
+
+    # Bundle everything up and return a Chains struct.
+    return Chains(
+        vals, vcat(param_names, [:lp]), (parameters = param_names, internals = [:lp],);
```
sethaxen, line 29:
```suggestion
        vals, vcat(param_names, tstat_names), (parameters = param_names, internals = tstat_names,);
```

test/mcmcchains.jl
```diff
@@ -0,0 +1,38 @@
...
+
+    θ_init = randn(rng, 2)
+
+    model = AdvancedHMC.LogDensityModel(LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓπ_gdemo))
+    init_eps = Leapfrog(1e-3)
+    κ = NUTS(init_eps)
+    metric = DiagEuclideanMetric(2)
+    adaptor = StanHMCAdaptor(MassMatrixAdaptor(metric), StepSizeAdaptor(0.8, κ.τ.integrator))
+
+    samples = AbstractMCMC.sample(
```
sethaxen, line 19: Can you add a test in which `bijector` is used?

sethaxen (changes requested):

sethaxen:
Documenting this with a usage example would also be great.

test/mcmcchains.jl
kaandocal: Done. I think I got a bit confused as to where bijectors belong, this is probably not the right place.

src/mcmcchains-connect.jl
```diff
@@ -0,0 +1,34 @@
...
+    state,
+    chain_type::Type{Chains};
+    discard_initial=0,
+    thinning=1,
+    param_names=missing,
+    bijector=identity,
+    kwargs...
+)
+    # Turn all the transitions into a vector-of-vectors.
+    tstat = merge((; lp = t.z.ℓπ.value), stat(t))
```
yebai, line 17:
```suggestion
    Tables.columntable([merge((; lp = t.z.ℓπ.value), stat(t)) for t in ts])
```

yebai (dismissed):

src/mcmcchains-connect.jl
```diff
@@ -0,0 +1,35 @@
...
+    thinning=1,
+    param_names=missing,
+    bijector=identity,
+    kwargs...
+)
+    # Turn all the transitions into a vector-of-vectors.
+    t = ts[1]
+    tstat = merge((; lp = t.z.ℓπ.value), stat(t))
+    tstat_names = collect(keys(tstat))
+    vals = [vcat(bijector(t.z.θ), collect(tstat)) for t in ts]
```
sethaxen, line 20: Currently this adds the statistics from the first transition to the values of every transition, but most of the statistics change with every transition. Something like this should be right:
```suggestion
    vals = [vcat(bijector(t.z.θ), t.z.ℓπ.value, collect(values(stat(t)))) for t in ts]
```

sethaxen (changes requested):

yebai:
thanks @kaandocal and @sethaxen!

PR #318, opened 2023-03-08
Use SimpleUnPack

devmotion:
Since `UnPack.@unpack` is based on `unpack(x, Val(key))` it can lead to a lot of undesired specializations, see e.g. https://github.com/SciML/OrdinaryDiffEq.jl/pull/1893. Therefore I put together a more lightweight macro in https://github.com/devmotion/SimpleUnPack.jl (by design, it has less features than UnPack.jl): the expression is just rewritten to a sequence of `getproperty` calls without any `Val` on older Julia versions and to the Base destructuring syntax on Julia >= 1.7.


src/adaptation/stepsize.jl
```diff
@@ -133,7 +133,10 @@ function adapt_stepsize!(
         @unpack m, ϵ, x_bar, H_bar = state
     end
 
-    @pack! state = m, ϵ, x_bar, H_bar
+    state.m = m
+    state.ϵ = 
+    state.x_bar = x_bar
+    state.H_bar = H_bar
```
devmotion, line 139: I plan to add a `@pack!` macro based on `setproperty` this week (possibly earlier), so in case you don't want the hand-written assignments you could wait a few days.

src/adaptation/stepsize.jl
```diff
@@ -133,7 +133,10 @@ function adapt_stepsize!(
         @unpack m, ϵ, x_bar, H_bar = state
     end
 
-    @pack! state = m, ϵ, x_bar, H_bar
+    state.m = m
+    state.ϵ = 
+    state.x_bar = x_bar
```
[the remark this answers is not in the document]
devmotion: Oh, no that's not a good suggestion :smile:

src/adaptation/stepsize.jl
```diff
@@ -133,7 +133,10 @@ function adapt_stepsize!(
         @unpack m, ϵ, x_bar, H_bar = state
     end
 
-    @pack! state = m, ϵ, x_bar, H_bar
+    state.m = m
+    state.ϵ =
```
devmotion, line 137:
```suggestion
    state.ϵ = ϵ
```

PR #319, opened 2023-03-09
No glue code

JaimeRZP:


JaimeRZP:
I have now coded the changes into AdvancedHMC src.
Users should be able to sample a Turing model using AdvancedHMC directly by using:
```Julia
n_samples, n_adapts = 10_000, 1_000
sample(model, metric, proposal, initial_θ, n_samples, adaptor, n_adapts)
```
or even simpler
```Julia
samples, stats = sample(model, 0.1, 0.95, n_samples, n_adapts; initial_θ=initial_θ)
```
which will use NUTS by defautl.

The unbounded to bounded space transforms are still missing.

Issue #313, opened 2023-01-16
More friendly default `sample` interface 


yebai:
We should consider supporting the following `AbstractMCMC.sample` interface (defined [here](https://turing.ml/library/AbstractMCMC/dev/api/#StatsBase.sample-Tuple{AbstractRNG,%20Any,%20AbstractMCMC.AbstractSampler,%20Any}))

```julia
sample(
    rng::Random.AbstractRNG=Random.default_rng(),
    logdensity::LogDensityProblem,
    sampler::AbstractSampler,
    N_or_isdone;
    kwargs...,
)
```

This allows users to pass a `LogDensityProblem` to the `sample` function instead of a `hamiltonian` object. We could default to Gaussian Kinect energy and canonical moment refreshment scheme. Moreover, we should define a few off-the-shelf `sampler` types rather than requiring the user to construct an `HMC` kernel from technically frightening components like `leapfrog` and trajectory `termination` types. 

Such an intuitive `sample` interface can even be helpful for `Turing` models, allowing users to run MCMC without using out-of-date glue interfaces (correctly) criticised [here](https://github.com/TuringLang/Turing.jl/pull/1818). 

- Related PR for `AdvancedMH`: https://github.com/TuringLang/AdvancedMH.jl/pull/76
- Related [code](https://github.com/TuringLang/AdvancedHMC.jl/blob/master/src/abstractmcmc.jl)

yebai:
@sethaxen any thoughts on how to design this API better?

yebai:
cc @devmotion

torfjelde:
But that interface _is_ supported, no? After #301 ?

yebai:
I think we still need to explicitly construct the `Hamiltonian` type and pass it to the `sample` function. Some cosmetic changes are needed in the `abstractmcmc` interface implementation to support an intuitive interface like the one in https://github.com/TuringLang/AdvancedMH.jl/pull/76.

torfjelde:
> I think we still need to explicitly construct the Hamiltonian type and pass it to the sample function

Nah, you can just construct the `HMCSampler` and pass that it:)

sethaxen:
@torfjelde can you give an example of what you mean?

In my view a user should be able to entirely configure the warm-up period and sampling period without requiring they specify initializations for step size, metric, and initial point, but allowing for the possibility that they specialize the initializations.

[...]

PR #325, opened 2023-06-28
Convinience constructors

### LINKED ISSUE #314: Make `HMCState` stores `rng` in `AbstractMCMC` interface.
https://github.com/TuringLang/AdvancedHMC.jl/blob/82de3ff64c88316f7ebdf1e407e3077b3870c8af/src/abstractmcmc.jl#L166

We should add `rng` to the HMC state for continuing HMC sampling from a previous stopping point.

JaimeRZP:
Fix #314. Fix https://github.com/TuringLang/AdvancedHMC.jl/issues/313

### DIFF
--- a/src/abstractmcmc.jl
+++ b/src/abstractmcmc.jl
@@ -286,8 +286,6 @@ function make_step_size(
     return spl.κ.τ.integrator.ϵ
 end
 
-#########
-
 make_integrator(spl::HMCSampler, ϵ::Real) = spl.κ.τ.integrator
 make_integrator(spl::AbstractHMCSampler, ϵ::Real) = make_integrator(spl.integrator, ϵ)
 make_integrator(i::AbstractIntegrator, ϵ::Real) = i
@@ -312,10 +310,6 @@ function make_metric(spl::AbstractHMCSampler, logdensity)
     d = LogDensityProblems.dimension(logdensity)
     T = get_type_of_spl(spl)
     return make_metric(spl.metric, T, d)
-end
-
-function make_metric(spl::HMCSampler, logdensity)
-    return spl.metric
 end
 
 #########


src/abstractmcmc.jl
```diff
@@ -224,6 +202,78 @@ function AbstractMCMC.step(
...
+### Turing ###
+##############
+
+function AbstractMCMC.step(
+    rng::AbstractRNG,
+    model::LogDensityModel,
+    spl::AbstractMCMC.AbstractSampler;
+    init_params = nothing,
+    kwargs...,
+)
```
devmotion, line 215: This definition is too generic, isn't it? We surely don't want this method be hit for every sampler, RNG, and log density model.

devmotion (commented):
A few high-level comments:
- It's still not clear to me why a `CustomHMC` is needed if there is already a `HMCSampler`. The only difference seems to be the missing integrator but that could just be set to a default value (`nothing`?) and a constructor could be added that allows to not specify an integrator.
- The `step` implementation seems to be too generic and hit cases that should not be covered here.
- I think the design mixes up the notion of algorithms and samplers. IMO algorithms should not be `AbstractSampler`s.

So I would propose:
- Remove `CustomHMC` and only use `HMCSampler` + a constructor `HMCSampler(...)` that does not define an integrator but sets it to `nothing` (if needed).
- Add common fields `init_eps`, `integrator_method` and `metric_type` to `HMCSampler`.
- Define `NUTSOptions` with `n_adapts`, `delta`, `max_depth`, `Delta_max`, `HMCOptions` with `n_leapfrog` and `HMCDAOptions` with `n_adapts`, `delta`, and `lambda`
- Add a field `opts` to `HMCSampler`
- Define aliases `const NUTS = HMCSampler{NUTSOptions}` etc.
- Define custom aliases `NUTS(...) = HMCSampler(NUTSOptions(...), ...)`
- Define `sample`, `step` etc. only for `HMCSampler`

An alternative to the last few points (still removing `CustomHMC`) would be to
- define `NUTS` etc. as algorithms, i.e., remove the AbstractSampler supertype
- define a `HMCSampler` constructor which takes these algorithms to create a `HMCSampler` (which is a `AbstractSampler`)

JaimeRZP:
Hi @devmotion!

Thank you a lot for having a look at this!
I think the approach you are suggesting is very similar to what I origally proposed in this [commit](https://github.com/TuringLang/AdvancedHMC.jl/blob/1af622cd1854f1bbf629015fe303c9aef58828fb/src/constructors.jl).
After some back and forth with @torfjelde we ended up going down this route but maybe we should have a call about it all the three of us?

torfjelde:
This is looking good @JaimeRZP :) Will have a proper look on Friday :+1:

README.md
```diff
@@ -89,6 +93,35 @@ adaptor = StanHMCAdaptor(MassMatrixAdaptor(metric), StepSizeAdaptor(0.8, integra
 #   - `stats` will store diagnostic statistics for each sample
 samples, stats = sample(hamiltonian, proposal, initial_θ, n_samples, adaptor, n_adapts; progress=true)
 ```
+## Using AdvancedHMC with Turing
```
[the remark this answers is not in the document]
JaimeRZP: done!

src/AdvancedHMC.jl
```diff
@@ -103,20 +81,9 @@ struct StaticTrajectory{TS} end
     Trajectory{EndPointTS}(Leapfrog(ϵ), FixedNSteps(L)),
 )
 
-struct HMCDA{TS} end
```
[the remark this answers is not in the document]
JaimeRZP: Is there a particular way of doing so?
Otherwise, I have added a comment to the index.

src/abstractmcmc.jl
```diff
@@ -60,36 +35,23 @@ A convenient wrapper around `AbstractMCMC.sample` avoiding explicit construction
 """
 function AbstractMCMC.sample(
```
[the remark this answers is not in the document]
JaimeRZP: Done!

src/abstractmcmc.jl
```diff
@@ -109,9 +71,7 @@ end
 
 function AbstractMCMC.sample(
```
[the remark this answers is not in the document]
JaimeRZP: Done!

src/abstractmcmc.jl
```diff
@@ -41,14 +14,16 @@ struct HMCState{
     TKernel<:AbstractMCMCKernel,
     TAdapt<:Adaptation.AbstractAdaptor,
 }
+    "Random number of the state"
+    rng::Random.AbstractRNG
```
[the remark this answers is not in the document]
JaimeRZP: The original motivation was indeed reproducibility.
I also agree that this will not cut it as we discussed in the other PR. 
Nonetheless, I would leave it in for this PR until we figure out how deal with it upstream.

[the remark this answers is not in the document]
torfjelde: See the discussion in #314 . I've made the same point as @devmotion ; IMO there's no point in doing this. Moreover, the `rng` is available to `callback`, so if it's about reproducibility, then we just make a callback that saves the state + the `rng`.

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+- `n_adapts::Int` : The number of samples to use with adaptation.
+- `δ::Float64` : Target acceptance rate for dual averaging.
+- `max_depth::Int` : Maximum doubling tree depth.
+- `Δ_max::Float64` : Maximum divergence during doubling tree.
+- `init_ϵ::Float64` : Initial step size; 0 means automatically searching using a heuristic procedure.
+
+"""
+Base.@kwdef struct NUTS <: AbstractHMCSampler
+    n_adapts::Int                      # number of samples with adaption for ϵ
+    δ::Float64                         # target accept rate
```
torfjelde, line 65: This is too strict. It should be possible to use, say, `Float32`. Similar with all the other fields restricted to `Float64`.

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+    max_depth::Int = 10                # maximum tree depth
+    Δ_max::Float64 = 1000.0            # maximum error
+    init_ϵ::Float64 = 0.0              # (initial) step size
+    integrator_method = Leapfrog       # integrator method
+    metric_type = DiagEuclideanMetric  # metric type
+end
```
torfjelde, line 71:
```suggestion
Base.@kwdef struct NUTS{T<:AbstractFloat} <: AbstractHMCSampler
    n_adapts::Int
    δ::T
    max_depth::Int = 10
    Δ_max::T = T(1000)
    init_ϵ::T = zero(T)
    integrator_method = Leapfrog
    metric_type = DiagEuclideanMetric
end
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+    adaptor::A
+    "Adaptation steps if any"
+    n_adapts::Int = 0
+end
+
+############
+### NUTS ###
+############
+"""
+    NUTS(n_adapts::Int, δ::Float64; max_depth::Int=10, Δ_max::Float64=1000.0, init_ϵ::Float64=0.0)
```
torfjelde, line 43:
```suggestion
    NUTS(n_adapts::Int, δ::Real; max_depth::Int=10, Δ_max::Real=1000, init_ϵ::Real=0)
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+
+```julia
+NUTS()            # Use default NUTS configuration.
+NUTS(1000, 0.65)  # Use 1000 adaption steps, and target accept ratio 0.65.
+```
+
+Arguments:
+
+- `n_adapts::Int` : The number of samples to use with adaptation.
+- `δ::Float64` : Target acceptance rate for dual averaging.
```
torfjelde, line 57:
```suggestion
- `δ::Real` : Target acceptance rate for dual averaging.
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+- `max_depth::Int` : Maximum doubling tree depth.
+- `Δ_max::Float64` : Maximum divergence during doubling tree.
+- `init_ϵ::Float64` : Initial step size; 0 means automatically searching using a heuristic procedure.
```
torfjelde, line 60:
```suggestion
- `Δ_max::Real` : Maximum divergence during doubling tree.
- `init_ϵ::Real` : Initial step size; 0 means automatically searching using a heuristic procedure.
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+    init_ϵ::Float64 = 0.0              # (initial) step size
+    integrator_method = Leapfrog       # integrator method
+    metric_type = DiagEuclideanMetric  # metric type
+end
+
+###########
+### HMC ###
+###########
+"""
+    HMC(ϵ::Float64, n_leapfrog::Int)
```
torfjelde, line 77:
```suggestion
    HMC(ϵ::Real, n_leapfrog::Int)
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+
+Hamiltonian Monte Carlo sampler with static trajectory.
+
+Arguments:
+
+- `ϵ::Float64` : The leapfrog step size to use.
```
torfjelde, line 83:
```suggestion
- `ϵ::Real` : The leapfrog step size to use.
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+# Reduced step size
+sample(gdemo([1.5, 2]), HMC(0.01, 10), 1000)
+```
+"""
+Base.@kwdef struct HMC <: AbstractHMCSampler
+    init_ϵ::Float64                    # leapfrog step size
+    n_leapfrog::Int                    # leapfrog step number
+    integrator_method = Leapfrog       # integrator method
+    metric_type = DiagEuclideanMetric  # metric type
+end
```
torfjelde, line 109:
```suggestion
Base.@kwdef struct HMC{T<:AbstractFloat} <: AbstractHMCSampler
    init_ϵ::T
    n_leapfrog::Int
    integrator_method = Leapfrog
    metric_type = DiagEuclideanMetric
end
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+
+#############
+### HMCDA ###
+#############
+"""
+    HMCDA(n_adapts::Int, δ::Float64, λ::Float64; ϵ::Float64=0.0)
```
torfjelde, line 115:
```suggestion
    HMCDA(n_adapts::Int, δ::Real, λ::Real; ϵ::Real=0)
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+```julia
+HMCDA(200, 0.65, 0.3)
+```
+
+Arguments:
+
+- `n_adapts::Int` : Numbers of samples to use for adaptation.
+- `δ::Float64` : Target acceptance rate. 65% is often recommended.
+- `λ::Float64` : Target leapfrog length.
+- `ϵ::Float64=0.0` : Initial step size; 0 means automatically search by Turing.
```
torfjelde, line 130:
```suggestion
- `δ::Real` : Target acceptance rate. 65% is often recommended.
- `λ::Real` : Target leapfrog length.
- `ϵ::Real=0` : Initial step size. If 0, then it is automatically determined.
```

torfjelde: This is AdvancedHMC, not Turing:)

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+  Research 15, no. 1 (2014): 1593-1623.
+"""
+Base.@kwdef struct HMCDA <: AbstractHMCSampler
+    n_adapts::Int                      # number of samples with adaption for ϵ
+    δ::Float64                         # target accept rate
+    λ::Float64                         # target leapfrog length
+    init_ϵ::Float64 = 0.0              # (initial) step size
+    integrator_method = Leapfrog       # integrator method
+    metric_type = DiagEuclideanMetric  # metric type
+end
```
torfjelde, line 145:
```suggestion
Base.@kwdef struct HMCDA{T<:AbstractFloat} <: AbstractHMCSampler
    n_adapts::Int
    δ::T
    λ::T
    init_ϵ::T = zero(T)
    integrator_method = Leapfrog
    metric_type = DiagEuclideanMetric
end
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+
+export HMCSampler, HMC, NUTS, HMCDA
```
torfjelde, line 147: Move this to the `src/AdvancedHMC.jl` under the `include` statement, as is done for all other `export` in this package.

docs/src/turing.md
```diff
@@ -0,0 +1,29 @@
+## Using AdvancedHMC with Turing
```
torfjelde, line 1: I'm a bit uncertain if we want this here; it seems more suitable to be put in Turing.jl, no?

And, if we keep it here, we should at least make the code blocks below `@example turing` or something so that they're run, given that the `externalsampler` is quite a new features and thus prone to breaking changes.

src/AdvancedHMC.jl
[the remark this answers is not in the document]
torfjelde: I've added suggestions to do so.

src/abstractmcmc.jl
```diff
@@ -303,3 +239,91 @@ function (cb::HMCProgressCallback)(rng, model, spl, t, state, i; nadapts = 0, kw
...
+
+#########
+
+function get_nadapts(spl::Union{HMCSampler,NUTS,HMCDA})
+    return spl.n_adapts
+end
+
+function get_nadapts(spl::HMC)
+    return 0
+end
```
torfjelde, line 311:
```suggestion
get_nadapts(spl::Union{HMCSampler,NUTS,HMCDA}) = spl.n_adapts
get_nadapts(spl::HMC) = 0
```

src/constructors.jl
```diff
@@ -0,0 +1,147 @@
...
+Usage:
+
+```julia
+NUTS()            # Use default NUTS configuration.
+NUTS(1000, 0.65)  # Use 1000 adaption steps, and target accept ratio 0.65.
+```
+
+Arguments:
+
+- `n_adapts::Int` : The number of samples to use with adaptation.
```
torfjelde, line 56: Generally speaking, I'd recommend just moving these docstrings to the fields themselves and use the $(FIELDS) to interpolate these directly into the docstring, rather than keeping track of this manually (see how it's done for `HMCSampler`)

docs/src/turing.md
JaimeRZP: Done!

src/constructors.jl
```diff
@@ -0,0 +1,172 @@
...
+$(FIELDS)
+
+# Usage:
+
+```julia
+NUTS(n_adapts=1000, δ=0.65)  # Use 1000 adaption steps, and target accept ratio 0.65.
+```
+"""
+struct NUTS{T<:AbstractFloat,I,D} <: AbstractHMCSampler
+    "`n_adapts::Int` : Number of adaptation steps."
```
devmotion, line 58: Just
```suggestion
    "Number of adaptation steps."
```
should be sufficient. Similar suggestions apply to the other fields below 🙂

src/constructors.jl
```diff
@@ -0,0 +1,172 @@
...
+    Δ_max::T
+    "`init_ϵ::Real` : Initial step size; 0 means automatically searching using a heuristic procedure."
+    init_ϵ::T
+    "[`AbstractIntegrator`](@ref)."
+    integrator_method::I
+    "[`AbstractMetric`](@ref)."
+    metric_type::D
+end
+
+function NUTS(
```
devmotion, line 74: Is this constructor actually desired now that you switched to `@kwdef`? Or should we not use `@kwdef`?

src/constructors.jl
```diff
@@ -0,0 +1,172 @@
...
+
+$(FIELDS)
+
+# Usage:
+
+```julia
+HMC(init_ϵ=0.05, n_leapfrog=10)
+```
+"""
+struct HMC{T<:AbstractFloat,I,D} <: AbstractHMCSampler
```
devmotion, line 104: Same here, just
```suggestion
struct HMC{T<:Real,I,D} <: AbstractHMCSampler
```
?

src/constructors.jl
JaimeRZP: Hi! We no longer use `@kwdef`. That's why we got the constructor.

src/constructors.jl
```diff
@@ -0,0 +1,172 @@
...
+    "`init_ϵ::Real` : Initial step size; 0 means automatically searching using a heuristic procedure."
+    init_ϵ::T
+    "`n_leapfrog::Int` : Number of leapfrog steps."
+    n_leapfrog::Int
+    "[`AbstractIntegrator`](@ref)."
```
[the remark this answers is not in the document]
JaimeRZP: We agreed to make these symbols such as the user provides `:LeapFrog` and then during the sampling the steps functions turn it into a type `integrator = eval(:LeapFrog)`. This should help users save the settings of the sampler more easily.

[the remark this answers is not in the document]
JaimeRZP: I also changed the doc string.

src/constructors.jl
```diff
@@ -0,0 +1,172 @@
...
+"""
+struct NUTS{T<:AbstractFloat,I,D} <: AbstractHMCSampler
+    "`n_adapts::Int` : Number of adaptation steps."
+    n_adapts::Int
+    "`δ::Real` : Target acceptance rate for dual averaging."
+    δ::T
+    "`max_depth::Int` : Maximum doubling tree depth."
+    max_depth::Int
+    "`Δ_max::Real` : Maximum divergence during doubling tree."
+    Δ_max::T
```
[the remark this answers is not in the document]
JaimeRZP: Done! Let me know if this is the best way of doing the promotion,

src/constructors.jl
torfjelde: We switched away from `@kwdef` because it doens't allow you to specify a default type parameter.

src/constructors.jl
```diff
@@ -0,0 +1,177 @@
...
+"""
+struct HMC{T<:Real,I,D} <: AbstractHMCSampler
+    "Initial step size; 0 means automatically searching using a heuristic procedure."
+    init_ϵ::T
+    "Number of leapfrog steps."
+    n_leapfrog::Int
+    "Choice of integrator method given as a symbol"
+    integrator_method::I
+    "Choice of metric type as given a symbol"
+    metric_type::D
```
torfjelde, line 113:
```suggestion
    "Choice of integrator, specified either using a `Symbol` or [`AbstractIntegrator`](@ref)"
    integrator::I
    "Choice of metric, specified either using a `Symbol` or `AbstractMetric`"
    metric::D
```

src/abstractmcmc.jl
```diff
@@ -303,3 +237,88 @@ function (cb::HMCProgressCallback)(rng, model, spl, t, state, i; nadapts = 0, kw
...
+    rng::Random.AbstractRNG,
+    spl::Union{HMC,NUTS,HMCDA},
+    hamiltonian::Hamiltonian,
+    init_params,
+)
+    if iszero(spl.init_ϵ)
+        ϵ = find_good_stepsize(rng, hamiltonian, init_params)
+        @info string("Found initial step size ", ϵ)
+    end
+    integrator = eval(spl.integrator_method)
```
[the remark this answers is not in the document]
torfjelde: Can also rename the above to `determine_integrator_constuctor`, to be more explicit.

torfjelde (changes requested):

src/abstractmcmc.jl
```diff
@@ -303,3 +237,88 @@ function (cb::HMCProgressCallback)(rng, model, spl, t, state, i; nadapts = 0, kw
...
+end
+
+function make_integrator(
+    rng::Random.AbstractRNG,
+    spl::HMCSampler,
+    hamiltonian::Hamiltonian,
+    init_params,
+)
+    # rerturns a dummy integrator
+    return Leapfrog(0.0)
```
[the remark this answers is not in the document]
JaimeRZP: It shouldn't because it is not used anywhere when a `HMCSampler `is provided.
I have made it return a `AbstractIntegrator `now which should be more general.

src/abstractmcmc.jl
```diff
@@ -303,3 +237,88 @@ function (cb::HMCProgressCallback)(rng, model, spl, t, state, i; nadapts = 0, kw
...
+end
+
+#########
+
+function make_metric(spl::Union{HMC,NUTS,HMCDA}, logdensity)
+    d = LogDensityProblems.dimension(logdensity)
+    metric = eval(spl.metric_type)
+    return metric(d)
```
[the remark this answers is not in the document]
JaimeRZP: Float64 for what?

src/abstractmcmc.jl
```diff
@@ -303,3 +239,138 @@ function (cb::HMCProgressCallback)(rng, model, spl, t, state, i; nadapts = 0, kw
...
 end
+
+#############
+### Utils ###
+#############
+
+function get_type_of_spl(spl::AbstractHMCSampler)
+    T = collect(typeof(spl).parameters)[1]
+    return T
+end
```
devmotion, line 250: This seems too hacky for my taste. According to a quick check, (unsurprisingly) it also can't be inferred. It seems this could be fixed by removing `collect` - but IMO the correct approach would be to define dispatches such as `get_type_of_spl(::NUTS{T}) = T` - or more generically define `abstract type AbstractHMCSampler{T<:Real} <: AbstractSampler end` and define `NUTS{T<:Real} <: AbstractHMCSampler{T} ...`.

src/abstractmcmc.jl
```diff
@@ -303,3 +239,138 @@ function (cb::HMCProgressCallback)(rng, model, spl, t, state, i; nadapts = 0, kw
...
+        error("Integrator $integrator not supported.")
+    end
+
+    return SYMBOL_TO_INTEGRATOR_TYPE[integrator]
+end
+
+# If it's the "constructor" of an integrator or instantance of an integrator, do nothing.
+determine_integrator_constructor(x::AbstractIntegrator) = x
+determine_integrator_constructor(x::Type{<:AbstractIntegrator}) = x
+determine_integrator_constructor(x) = error("Integrator $x not supported.")
```
devmotion, line 269: Can we use a different design? It seems quite weird... Why do we need to retrieve the constructor? Couldn't we just use something like `get_integrator(Val(spl.integrator), eps)` where we dispatch on the first argument?

src/constructors.jl
devmotion: `HMCSampler` still uses `@kwdef` it seems?

src/abstractmcmc.jl
JaimeRZP: Thanks David! I was trying to get something like this to work for 1h yesterday until I exasperated.

src/abstractmcmc.jl
JaimeRZP: Ok let me try!

JaimeRZP: I think we still want to keep the look up table design but I have disentangled `make_step_size `from `make_integrator`.
@torfjelde?

src/abstractmcmc.jl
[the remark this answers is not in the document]
JaimeRZP: Solved!

src/abstractmcmc.jl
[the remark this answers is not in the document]
JaimeRZP: I found a more elegant solution!
I now simply returns the integrator provided by the user, even though it is technically not used anywhere.

src/abstractmcmc.jl
JaimeRZP: I like this solution a lot!! I have given the same treatment to the integrator and the metric. 
There's a little bit of nesting going on here but basically if the user provides an `AbstractMetric`/`AbstractIntegrator ` the code simly returns it. If on the other hand, the user provides a symbol it gets wrapped in  `Val(::Symbol)` and then we do your solution.

This allows the user to use the NUTS constructor with their own metric/integrator without having to build the whole sampler themselves which is something I always wanted to code into these constructors. 

```julia
make_integrator(spl::HMCSampler, ϵ::Real) = spl.κ.τ.integrator
make_integrator(spl::AbstractHMCSampler, ϵ::Real) = make_integrator(spl.integrator, ϵ)
make_integrator(i::AbstractIntegrator, ϵ::Real) = i
make_integrator(i::Type{<:AbstractIntegrator}, ϵ::Real) = i
make_integrator(i::Symbol, ϵ::Real) = make_integrator(Val(i), ϵ)
make_integrator(i...) = error("Integrator $(typeof(i)) not supported.")
make_integrator(i::Val{:leapfrog}, ϵ::Real) = Leapfrog(ϵ)
make_integrator(i::Val{:jitteredleapfrog}, ϵ::Real) = JitteredLeapfrog(ϵ)
make_integrator(i::Val{:temperedleapfrog}, ϵ::Real) = TemperedLeapfrog(ϵ)

#########

make_metric(i...) = error("Metric $(typeof(i)) not supported.")
make_metric(i::Symbol, T::Type, d::Int) = make_metric(Val(i), T, d)
make_metric(i::AbstractMetric, T::Type, d::Int) = i
make_metric(i::Type{AbstractMetric}, T::Type, d::Int) = i
make_metric(i::Val{:diagonal}, T::Type, d::Int) = DiagEuclideanMetric(T, d)
make_metric(i::Val{:unit}, T::Type, d::Int) = UnitEuclideanMetric(T, d)
make_metric(i::Val{:dense}, T::Type, d::Int) = DenseEuclideanMetric(T, d)
```

src/abstractmcmc.jl
```diff
@@ -303,3 +239,120 @@ function (cb::HMCProgressCallback)(rng, model, spl, t, state, i; nadapts = 0, kw
...
+    hamiltonian::Hamiltonian,
+    init_params,
+)
+    return spl.κ.τ.integrator.ϵ
+end
+
+make_integrator(spl::HMCSampler, ϵ::Real) = spl.κ.τ.integrator
+make_integrator(spl::AbstractHMCSampler, ϵ::Real) = make_integrator(spl.integrator, ϵ)
+make_integrator(i::AbstractIntegrator, ϵ::Real) = i
+make_integrator(i::Type{<:AbstractIntegrator}, ϵ::Real) = i
```
devmotion, line 292: Are both needed?

src/abstractmcmc.jl
```diff
@@ -303,3 +239,120 @@ function (cb::HMCProgressCallback)(rng, model, spl, t, state, i; nadapts = 0, kw
...
+make_integrator(i::Symbol, ϵ::Real) = make_integrator(Val(i), ϵ)
+make_integrator(i...) = error("Integrator $(typeof(i)) not supported.")
```
devmotion, line 294: This won't work as desired IIUC:
```julia
julia> f(x...) = typeof(x)

julia> f(3)
Tuple{Int64}

julia> f(1.0, "a")
Tuple{Float64, String}
```
IMO one could simply use
```suggest
make_integrator(@nospecialize(i), ::Real) = error("Integrator $i not supported.")
```

PR #336, opened 2023-07-26
Minor improvements to constructors and related tests

torfjelde:
Some additional changes I was going to request in #325


torfjelde:
For future reference: let the person who opened the PR merge the PR:) For example, in this scenario, I thinking if maybe I should add a few more tests for the constructors.

PR #341, opened 2023-07-27
Add back the `HMC(eps, n_leapfrog)` as a convenience constructor

yebai:
Our doc string still contains the old `HMC(eps, N)` constructor, so let's actually define it. 

https://github.com/TuringLang/AdvancedHMC.jl/blob/84290771ada04536c5247ac226cb1baed043dbc4/src/constructors.jl#L109


torfjelde (dismissed):
Can you also add a test for this? See `test/constructors.jl` and add it to the list there.

JaimeRZP (approved):

Issue #344, opened 2023-08-01
Component Arrays with DenseEuclideanMetric fails


erathorn:
Dear Team,

Thak you for your amazing package!

I tried to play around with your wonderful package. However, I noticed that using ComponentArrays and a DenseEuclideanMetric does not work. 
Here is a quote from the tests.

```Julia
using AdvancedHMC, ComponentArrays, ForwardDiff

p1 = ComponentVector(μ = 2.0, σ = 1)
struct DemoProblemComponentArrays end

function LogDensityProblems.logdensity(::DemoProblemComponentArrays, p::ComponentArray)
    return -((1 - p.μ) / p.σ)^2
end
LogDensityProblems.dimension(::DemoProblemComponentArrays) = 2
LogDensityProblems.capabilities(::Type{DemoProblemComponentArrays}) =
    LogDensityProblems.LogDensityOrder{0}()

ℓπ = DemoProblemComponentArrays()

# Define a Hamiltonian system
D = length(p1)          # number of parameters
metric = DenseEuclideanMetric(D) # !!! IN THE ORIGINAL TESTS, THIS IS DiagEuclideanMetric !!!

# choose AD framework or provide a function manually
hamiltonian = Hamiltonian(metric, ℓπ, Val(:ForwardDiff); x = p1)

# Define a leapfrog solver, with initial step size chosen heuristically
initial_ϵ = find_good_stepsize(hamiltonian, p1)
```
This code fails with:
```Julia
ERROR: MethodError: no method matching AdvancedHMC.PhasePoint(::ComponentVector{Float64, Vector{Float64}, Tuple{Axis{(μ = 1, σ = 2)}}}, ::ComponentVector{Float64, Vector{Float64}, Tuple{Axis{(μ = 1, σ = 2)}}}, ::AdvancedHMC.DualValue{Float64, ComponentVector{Float64, Vector{Float64}, Tuple{Axis{(μ = 1, σ = 2)}}}}, ::AdvancedHMC.DualValue{Float64, Vector{Float64}})
Closest candidates are:
  AdvancedHMC.PhasePoint(::T, ::T, ::V, ::V) where {T, V} at ~/JULIAPACKAGES/103/packages/AdvancedHMC/jiCaS/src/hamiltonian.jl:53
```
I think the problem is here: 
[...]

PR #345, opened 2023-08-03
DenseMetric and Component arrays (solve #344)

erathorn:
This PR attempts to solve #344 

I went for the solution to preallocate the result in `∂H∂r` such that the type of the input `r` matches the type of the output. 
I added tests that not only check the correct numerical output but also check the type.

Additionally, I found a typo in the inner constructor of PhasePoint, where it originally was `length(ℓπ.gradient) == length(ℓπ.gradient)`.

### DIFF
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -42,8 +42,11 @@ end
 ∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) = copy(r)
 ∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) =
     h.metric.M⁻¹ .* r
-∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) =
-    h.metric.M⁻¹ * r
+function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat)
+    out = similar(r) # Make sure the output of this function is of the same type as r
+    mul!(out, h.metric.M⁻¹, r)
+    out
+end
 
 struct PhasePoint{T<:AbstractVecOrMat{<:AbstractFloat},V<:DualValue}
     θ::T  # Position variables / model parameters.
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -51,7 +51,7 @@ struct PhasePoint{T<:AbstractVecOrMat{<:AbstractFloat},V<:DualValue}
     ℓπ::V # Cached neg potential energy for the current θ.
     ℓκ::V # Cached neg kinect energy for the current r.
     function PhasePoint(θ::T, r::T, ℓπ::V, ℓκ::V) where {T,V}
-        @argcheck length(θ) == length(r) == length(ℓπ.gradient) == length(ℓπ.gradient)
+        @argcheck length(θ) == length(r) == length(ℓπ.gradient) == length(ℓk.gradient)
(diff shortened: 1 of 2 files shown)


src/hamiltonian.jl
```diff
@@ -42,16 +42,19 @@ end
 ∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) = copy(r)
 ∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) =
     h.metric.M⁻¹ .* r
-∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) =
-    h.metric.M⁻¹ * r
+function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat)
+    out = similar(r) # Make sure the output of this function is of the same type as r
+    mul!(out, h.metric.M⁻¹, r)
+    out
+end
```
torfjelde, line 49: I'm a bit uncertain about this change as it "complicates" code to mainly just stay compatible with ComponentArrays.jl, and thus I'd be more in favour of just making it an extension instead, I think :confused: Then in the extension, we just overload whatever we need to be compatible.

Also, will this code break if, say, `h.metric.M⁻¹` has eltype `Float64` but `r` has eltype `Float32`, rather than just promoting, as is current behavior?

test/hamiltonian.jl
```diff
@@ -49,18 +50,50 @@ end
...
+        end
+    end
+end
+
+@testset "Metric ComponentArrays" begin
+    n_tests = 10
+    for T in [Float32, Float64]
+        for _ = 1:n_tests
+            θ_init = ComponentArray(a = randn(T, D), b= randn(T, D))
+            r_init = ComponentArray(a = randn(T, D), b= randn(T, D))
```
torfjelde, line 77:
```suggestion
            θ_init = ComponentArray(a = randn(T, D), b = randn(T, D))
            r_init = ComponentArray(a = randn(T, D), b = randn(T, D))
```

src/hamiltonian.jl
```diff
@@ -42,16 +42,19 @@ end
 ∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) = copy(r)
 ∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) =
```
torfjelde, line 43: Also, does this not suffer from the same issue? Is ComponentArrays.jl "clever" enough to realize this is a diagonal matrix and so there's no "mixing" between the components?

torfjelde (changes requested):
Thanks @erathorn ! And nice catch on the constructor for `Phasepoint`!

I'm wondering if we should just make compat with ComponentArrays.jl an extension instead of complicating the existing code, and then we can just overload whatever we need there. Thoughts?

src/hamiltonian.jl
erathorn: The diagonal matrix case does not have the problem, because of the broadcasted multiplication. The type of `r` is maintained in this case. The test in line 89 of /test/hamiltonian.jl should cover this case.

src/hamiltonian.jl
erathorn:
```Julia
p1 = ComponentArray(m=one(Float32), s = one(Float32))
r = similar(p1)
M = diagm(randn(Float64, 2))
mul!(r, M, p1)
```

This works on my machine, and returns `r` as a component array of eltype `Float32` as expected.

torfjelde:
> The tests fail at this line:

This is indeed strange given that the CI on master is working just fine :confused:

PR #348, opened 2023-08-18
Find step_size if zero

yebai:
Determine initial leapfrog step size via `find_step_size` automatically when a user passes `Leapfrog(0.0)`.


xukai92 (commented):
can we do something to avoid the code duplication?

torfjelde (commented):
I'd also suggest refraining from comparing floats to 0 as this can sometimes fail, but I guess that's something for another PR :confused:

PR #353, opened 2023-10-27
Deprecate `init_params` which is no longer in AbstractMCMC

torfjelde:


torfjelde:
Maybe we shouldn't do this actually. We should probably just bump minor version and drop support for `init_params` completely, as we've done in other packages: https://github.com/TuringLang/AdvancedMH.jl/pull/92#event-10797486064

Issue #350, opened 2023-09-28
`AbstractMCMC.step` interface broken


devmotion:
https://github.com/TuringLang/AdvancedHMC.jl/pull/332 broke the `AbstractMCMC.step` interface: It's not possible anymore to initialize the sampler, obtain the initial state, and then continue sampling with `AbstractMCMC.step`.

The problem is
https://github.com/TuringLang/AdvancedHMC.jl/blob/04ce92ab167df3c2a309e845911129d3a1d43aa7/src/abstractmcmc.jl#L161
It assumes that `nadapts` is provided as a keyword argument to the `step` function by the user. This is not part of the AbstractMCMC.step interface requirements. If AdvancedHMC wants to use such a keyword argument in the function body it should be possible to omit it and have a default value. I think that's still a bad idea though since then users/developers have to know about how AdvancedHMC internally (ab)uses keyword arguments to be able to pass around this information in the same way. IMO it would be better to put this into the state of the sampler if it's not intended to be fixed a priori when constructing the sampler.

devmotion:
Just remembered https://github.com/TuringLang/AbstractMCMC.jl/pull/124.

I still think this should be fixed in AdvancedHMC rather than AbstractMCMC - we can't add all downstream options to AbstractMCMC and it means that custom implementations based on `step` (as in my use case) will always use a default value of 0, if they don't know about and reimplement the logic in AdvancedHMC.

PR #359, opened 2023-11-03
Fix Step

JaimeRZP:
Answer to #350

### DIFF
--- a/src/abstractmcmc.jl
+++ b/src/abstractmcmc.jl
@@ -141,6 +141,7 @@ function AbstractMCMC.step(
     model::AbstractMCMC.LogDensityModel,
     spl::AbstractHMCSampler,
     state::HMCState;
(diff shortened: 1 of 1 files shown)


devmotion:
This will fix the error I ran into, but the more general problem still exists: Since AdvancedHMC adds special keyword arguments to `sample`, `sample` and the `step` and `steps` iterators/transducers become inconsistent and generic calls of `sample` in other packages will perform sampling differently from AdvancedHMC because in general `n_adapts` etc. are not supported by `sample` and hence can't be provided in such generic situations.

That is, I think to fix these inconsistencies customizations of the `sample` call should mainly be done via the `sampler` (e.g., by adding fields that change the behaviour of `step`).

devmotion (approved):
I think it's not the ideal approach but it avoids the nasty error at least.

PR #366, opened 2024-04-23
Initial `AbstractMCMC.step` should not sample

torfjelde:
Generally speaking, the first `AbstractMCMC.step` should just *construct* the initial transition and state, not actually perform any sampling. If it performs sampling in the initial step, it becomes increasingly difficult to test that arguments such as `initial_params` and `initial_state` are properly respected (since we can't just compare the resulting `initial_params` with the ones produced by the first step). Moreover, it can hide certain behaviors from the user since a traceplot will show the "starting value" as something different than what they set.

This PR just makes the initial `AbstractMCMC.step` do exactly this.

Should be a quick merge. @devmotion @yebai @xukai92


torfjelde:
Hmm, so I'm realizing that it seems _most_ packages in fact do some sampling in the initial step :confused: 

I'm not uncertain if this is desirable given that it makes testing correctness, etc. as mentioned above somewhat difficult..

devmotion (approved):

mhauru:
Is this good to merge? Except that we should also bump a breaking version number.

torfjelde:
I think there are still some tests that need to be updated, but otherwise, yeah it should be good:)

torfjelde:
Accidentally merged this 🤦 Will re-open

sunxd3:
I wasn't prompted by @MateusMaiaDS to look into if we could make sure that the initialization is included in the chain for debugging purposes. @penelopeysm points this PR to me (I remembered I read it time ago.)

There is an old issue at https://github.com/TuringLang/Turing.jl/issues/1282 also related to this.

Looking at main, I think the merge here was reverted because main branch doesn't contain the code. Thoughts on should we still push this through @torfjelde @yebai @devmotion ?

PR #369, opened 2024-07-04
feat: support position-dependent kinetic

xukai92:
extend the interface with position-independent fallback as default

### DIFF
--- a/src/metric.jl
+++ b/src/metric.jl
@@ -141,8 +141,12 @@ Base.rand(metric::AbstractMetric, kinetic::AbstractKinetic) =
     rand(GLOBAL_RNG, metric, kinetic)
 
 # ignore θ by default unless defined by the specific kinetic (i.e. not position-dependent)
-Base.rand(rng::AbstractRNG, metric::AbstractMetric, kinetic::AbstractKinetic, θ::AbstractVecOrMat) =
-    rand(rng, metric, kinetic)    # this disambiguity is required by Random.rand
+Base.rand(
+    rng::AbstractRNG,
+    metric::AbstractMetric,
+    kinetic::AbstractKinetic,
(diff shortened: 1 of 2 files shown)


xukai92:
I accepted your suggestions. do you want to merge this PR then?

src/metric.jl
```diff
@@ -129,13 +129,15 @@ function _rand(
     return r
 end
 
-# TODO The rand interface should be updated by rand from momentum distribution + optional affine transformation by metric
-Base.rand(rng::AbstractRNG, metric::AbstractMetric, kinetic::AbstractKinetic) =
+# TODO The rand interface should be updated as "rand from momentum distribution + optional affine transformation by metric"
+# ignore θ by default unless defined by the specific kinetic (i.e. not position-dependent)
+Base.rand(rng::AbstractRNG, metric::AbstractMetric, kinetic::AbstractKinetic, θ) =
```
mhauru, line 134: In this and the below methods, would it make sense to apply similar type bounds to `θ` as above in e.g. `∂H∂θ`?

mhauru (commented):
Do I understand correctly that `θ` are the "position" parameters and `r` are the corresponding canonical momentum, and the idea here is to extend the interface so that the metric could depend on `θ`, although that is still unimplemented?

The code looks fine to me from a software point of view. The main issue I have is that the module in general relies on short, obscure variable names with special meanings to be readable. A lot of functions and structs would benefit form docstrings explaining what they do and what the fields and arguments are. This is of course not this PR's fault, but if you feel like starting the work of documenting things then it would definitely be welcome.

I'll leave approving to others because I don't understand the wider context here, but I don't see anything wrong. I had one minor comment about possible type bounds to consider.

PR #372, opened 2024-07-25
general relativistic HMC

xukai92:


yebai (commented):
I left some comments above. Will continue later.

yebai (changes requested):
@xukai92 I added some further comments. Please take a look. 

A high-level comment: 

- Perhaps we should add a folder `general_relavistic` (or `riemannian_relativistic`), and move all general relavistic code there:  `general_relavistic/hamiltonian.jl` and `general_relavistic/metric.jl`.

nsiccha:
Leaving a comment to remember to benchmark my upcoming implementation against this one, @yebai.

PR #374, opened 2024-10-03
Use `Base.RefValue`

devmotion:
`Ref{Int}` is an abstract type:

```julia
julia> typeof(Ref(0))
Base.RefValue{Int64}

julia> typeof(Ref(0)) <: Ref{Int}
true

julia> isabstracttype(Ref{Int})
true

julia> isabstracttype(Base.RefValue{Int})
false
```


torfjelde (approved):
LGTM!

PR #375, opened 2024-10-03
Use `n_adapts` instead of `nadapts`

devmotion:
Some code, tests, and even the README used `nadapts` instead of `n_adapts`. I replaced all occurrences of `nadapts` with `n_adapts`. Additionally, I thought it would be good to error out loudly if a user or package actually uses the non-functional `nadapts` keyword argument.


torfjelde (dismissed):
Awesome; thanks @devmotion :)

Maybe bump patch version?

devmotion:
Still fine @torfjelde?

torfjelde (approved):
Aye, very much so:)

PR #376, opened 2024-10-03
Require AbstractMCMC 5

devmotion:
Since AdvancedHMC uses the keyword argument `initial_params` and AbstractMCMC 4 used `init_params`, I think we should require AbstractMCMC 5 which also uses `initial_params`.


torfjelde (approved):
Lovely:)

research/Project.toml
```diff
@@ -1,31 +1,8 @@
 [deps]
-AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
```
torfjelde, line 2: This I'm uncertain about; did this occur automatically or was it intended? I think this is something maybe @xukai92 @yebai should confirm is okay?

devmotion: The test was failing in the first commit since this Project.toml file restricted AbstractMCMC to < 5. However, I noticed that AbstractMCMC is neither loaded nor accessed in the files in this folder at all. Which made me curious, so I searched for all packages in the Project.toml file and removed all of them that do not show up anywhere in research/

torfjelde: Lovely:)

PR #377, opened 2024-10-04
Update GH actions and add dependabot

devmotion:
Github is already complaining about some actions that still use Node 12 but are forced to run with Node 16 by now.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+# https://docs.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+version: 2
+updates:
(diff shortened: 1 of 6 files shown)


Project.toml
```diff
@@ -48,7 +48,7 @@ StatsBase = "0.31, 0.32, 0.33, 0.34"
 StatsFuns = "0.8, 0.9, 1"
 LinearAlgebra = "1.6"
 Random = "1.6"
-julia = "1.6"
+julia = "1.6.7"
```
devmotion, line 51: Tests seem to pass with 1.6.7 but didn't with 1.6.0. Not sure which dependency is the problem.

PR #378, opened 2024-10-21
Add `getparams` and `setparams!!` following AbstractMCMC v5.5 and v5.6

sunxd3:
ref: https://github.com/TuringLang/Turing.jl/issues/2367


sunxd3:
the failed test is not related to this PR, @yebai maybe you already have an answer to why it fails?

src/abstractmcmc.jl
```diff
@@ -30,6 +30,15 @@ getadaptor(state::HMCState) = state.adaptor
 getmetric(state::HMCState) = state.metric
 getintegrator(state::HMCState) = state.κ.τ.integrator
 
+function AbstractMCMC.getparams(state::HMCState)
+    # TODO(sunxd): should we return a copy?
+    return state.transition.z.θ
+end
+
+function AbstractMCMC.setparams!!(state::HMCState, θ)
```
torfjelde, line 38: One dangerous aspect there is that `state.transition` _caches_ the logjoint and gradient computations. So when you use the `@set` macro here, you're going to also keep the cached log-joint and gradient computation, which will then be out of sync with the parameters.

If you then naively pass this transition somewhere, say, into the next `step` call, AHMC.jl will use the incorrect logjoint eval in the MH step.

IMO the safe way is to use the explicit constructor of `PhasePoint` I believe without passing in the cached values. This should result in receomputation of this.

torfjelde (changes requested):
The current impl is going to lead to incorrect log-density values being used in the transition. IMO we shouuld do the safe thing by default here.

src/abstractmcmc.jl
sunxd3: just realized that `model` is not passed into `setparams!!`, I think for now, we only set the parameters, then when use the transition, logp should be recomputed. (we can also later introduce some like `setlogp` or `compute_logp` etc.) I'll add a comment in the code

sunxd3:
let's delay a bit to see through https://github.com/TuringLang/AbstractMCMC.jl/pull/150

torfjelde (changes requested):
The fix to `setparams!!` doesn't work unfortauntely. We need access to the `model` here, which we don't have.

IMO we should add `model` as an argument to the AbstractMCMC.jl interface. Otherwise there's no way we can do stuff like update the gradient information, which is an issue we're facing both here and in AdvancedMH.jl.

torfjelde (approved):
Good stuff @sunxd3 :)

PR #387, opened 2025-02-18
Fix `rand` and `randn` type piracy

devmotion:
I just came across a seemingly severe case of type piracy in AdvancedHMC. As a first step I just removed it to see what parts are currently making use of it.

Edit: I removed the definition of `Base.rand` (replaced with branches in the two function where it was used) and added an internal `_randn` to fix the type piracy of `Base.randn` (falls back to `Base.randn` for single RNGs and keeps the current behaviour for vectors of RNGs). Additionally, I renamed the existing internal `_rand` function to `rand_momentum` to make it clearer that it's just about sampling a momentum and not similar to `_randn`.


src/utilities.jl
```diff
@@ -80,8 +96,7 @@ function randcat(
     rng::Union{AbstractRNG,AbstractVector{<:AbstractRNG}},
     P::AbstractMatrix{T},
 ) where {T}
-    u = rand(rng, T, size(P, 2))
+    u = _rand(rng, T, size(P, 2))
     C = cumsum(P; dims = 1)
-    indices = convert.(Int, vec(sum(C .< u'; dims = 1))) .+ 1
-    return max.(indices, 1)  # prevent numerical issue for Float32
+    return max.(vec(count(C .< u'; dims = 1))) .+ 1, 1)  # prevent numerical issue for Float32
```
devmotion, line 101:
```suggestion
    return max.(vec(count(C .< u'; dims = 1)) .+ 1, 1)  # prevent numerical issue for Float32
```

penelopeysm:
@devmotion Could I get you to fix the conflicts? Then I can take a look again.

devmotion:
Done, I fixed the merge conflict.

research/src/riemannian_hmc.jl
```diff
@@ -142,9 +142,9 @@ refresh(
 # To change L146 of metric.jl
 # Ignore θ by default (i.e. not position-dependent)
 Base.rand(rng::AbstractRNG, metric::AbstractMetric, kinetic, θ) =
-    _rand(rng, metric, kinetic)    # this disambiguity is required by Random.rand
+    rand_momentum(rng, metric, kinetic)    # this disambiguity is required by Random.rand
 Base.rand(rng::AbstractVector{<:AbstractRNG}, metric::AbstractMetric, kinetic, θ) =
-    _rand(rng, metric, kinetic)
+    rand_momentum(rng, metric, kinetic)
 Base.rand(metric::AbstractMetric, kinetic, θ) = rand(Random.default_rng(), metric, kinetic)
```
penelopeysm, line 148: > Additionally, I renamed the existing internal _rand function to rand_momentum to make it clearer that it's just about sampling a momentum

I'm always hugely in favour of disambiguating names so very glad to see this.

There seem to be a few of these definitions

```julia
Base.rand(args...) = rand_momentum(fewer_args...)
```

Would it make sense to rename those `Base.rand` to `rand_momentum` as well? Happy to leave it to another time too, just curious if there's a reason you didn't do that.

devmotion: I was about to remove them but then decided to keep them in this PR because it could (should?) be considered part of the public API whereas `_rand` etc are surely internal.

penelopeysm: Hmm, I thought yes that makes sense, then looking at it more closely I realised the `research` subdirectory is just a bunch of scripts that use AdvancedHMC rather than part of the module itself. So I guess the name doesn't really matter so much. Feel free to hit merge if you're happy

PR #388, opened 2025-02-18
Reduce allocations in `all` + `any` and remove `isfinite` type piracy

devmotion:
The type piracy is detected and tested by #395.

### DIFF
--- a/src/trajectory.jl
+++ b/src/trajectory.jl
@@ -275,7 +275,7 @@ function transition(
             hamiltonian_energy = H,
             hamiltonian_energy_error = H - H0,
             # check numerical error in proposed phase point. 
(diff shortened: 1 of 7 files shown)


src/trajectory.jl
```diff
@@ -275,7 +275,7 @@ function transition(
             hamiltonian_energy = H,
             hamiltonian_energy_error = H - H0,
             # check numerical error in proposed phase point. 
-            numerical_error = isfinite(H′),
+            numerical_error = !all(isfinite, H′),
```
penelopeysm, line 278: 🤨 at the previous lack of negation

penelopeysm (approved):

PR #390, opened 2025-02-18
Reduce allocations in stepsize.jl

devmotion:


yebai:
@devmotion can you fix the merge clash before I review this PR?

src/adaptation/stepsize.jl
```diff
@@ -8,7 +8,7 @@ mutable struct DAState{T<:AbstractScalarOrVec{<:AbstractFloat}}
     H_bar::T
 end
 
-computeμ(ϵ::AbstractScalarOrVec{<:AbstractFloat}) = log.(10 * ϵ)
+computeμ(ϵ::AbstractFloat) = log(10 * ϵ)
```
yebai, line 11: Caution is required here: these support the vectorised version of `HMC`. Do you know how `map` would differ from broadcasting here?

src/adaptation/stepsize.jl
```diff
@@ -17,22 +17,34 @@ end
 
 function DAState(ϵ::AbstractVector{T}) where {T}
     n = length(ϵ)
-    μ = computeμ(ϵ)
+    μ = map(computeμ, ϵ)
```
yebai, line 20: See the above comment.

src/adaptation/stepsize.jl
```diff
@@ -17,22 +17,34 @@ end
...
+    das.H_bar = zero(T)
+    return das
 end
 
 function reset!(das::DAState{<:AbstractVector{T}}) where {T<:AbstractFloat}
     das.m = 0
-    das.μ .= computeμ(das.ϵ)
-    das.x_bar .= zero(T)
-    return das.H_bar .= zero(T)
+    map!(computeμ, das.μ, das.ϵ)
```
yebai, line 34: Isn't the `.=` statement storing results in LHS in place too? Can you explain the difference between `.=` and `map!`? Sorry, I haven't used `map!` or `fill!` extensively before.

src/adaptation/stepsize.jl
```diff
@@ -151,9 +166,12 @@ function adapt!(
...
-reset!(da::NesterovDualAveraging) = reset!(da.state)
+function reset!(da::NesterovDualAveraging)
+    reset!(da.state)
+    return da
+end
 
 function finalize!(da::NesterovDualAveraging)
-    da.state.ϵ = exp.(da.state.x_bar)
-    return nothing
+    finalize!(da.state)
```
yebai, line 175: nice improvement!

yebai (commented):
Thanks @devmotion -- I left a few questions below.

I'll have to take another closer look later this week.

src/adaptation/stepsize.jl
devmotion: The results of the calculations won't be affected by this change, but using the non-broadcasted formulation for scalars and `map` for vectors of floats will remove the broadcasting overhead and reduce stress on the compiler, i.e., generally it reduces compilation time. Sometimes it also helps type inference (but this case is too simple for this effect I assume).

In my experience, broadcasting is useful if one's actually broadcasting values of different size and dimensions but otherwise often a suboptimal choice.

yebai, line 11: I'd suggest that we define a vectorised version along with your proposal: 

```suggestion
computeμ(ϵ::AbstractFloat) = log(10 * ϵ)
computeμ(ϵ::AbstractVector{<:AbstractFloat}) = map(computeμ, ϵ)
```

The motivation is to keep the vectorised interface unchanged - we could refactor this interface in a separate effort.

src/adaptation/stepsize.jl
yebai, line 20:
```suggestion
    μ = computeμ(ϵ)
```

src/adaptation/stepsize.jl
yebai, line 34: Let's keep this as-is for now. We could refactor the vectorised HMC interface, but better to do it seprately in a concerted effort: 
```suggestion
    das.μ .= computeμ(das.ϵ)
```

src/adaptation/stepsize.jl
```diff
@@ -95,35 +117,28 @@ end
...
     δ::T, ϵ::VT
 ) where {T<:AbstractFloat,VT<:AbstractScalarOrVec{T}}
-    return NesterovDualAveraging(T(0.05), T(10.0), T(0.75), δ, ϵ)
+    return NesterovDualAveraging(T(1//20), T(10), T(3//4), δ, ϵ)
 end
 
 # Ref: https://github.com/stan-dev/stan/blob/develop/src/stan/mcmc/stepsize_adaptation.hpp
 # Note: This function is not merged with `adapt!` to empahsize the fact that
 #       step size adaptation is not dependent on `θ`.
 function adapt_stepsize!(
```
yebai, line 126:
```suggestion
# Note 2: `da.state` and `α` support vectorised HMC but should do so together. 
function adapt_stepsize!(
```

yebai (changes requested):
Thanks, @devmotion. Nice improvements. I left a few comments below, mostly about whether we should refactor the vectorised HMC implementation in a concerted effort separately to avoid inconsistency.

src/adaptation/stepsize.jl
devmotion: This suggestion would go against the main intention of the PR, reducing unnecessary allocations: With `map!` (or `das.μ .= computeμ.(das.ϵ)`, but the broadcasting is more stressful for the compiler) no intermediate array would be created in this line, whereas with the suggestion on the right-hand side a new array is allocated that is then copied to `das.μ` (as a side remark, for the compiler `copyto!` should be simpler than broadcasting).

penelopeysm (dismissed):

yebai (approved):

yebai:
Feel free to merge once CI passes!

PR #391, opened 2025-02-18
FIx incorrect ratios of divergent transitions

### LINKED ISSUE #373: Progress message `ratio_divergent_transitions_during_adaption` bug
This is difficult to demonstrate, but I believe the value of `ratio_divergent_transitions_during_adaption` printed during sampling when `progress=true` is currently incorrect. 

The behaviour I've seen is:

1. During adaptation, numerical errors are encountered causing `ratio_divergent_transitions_during_adaption` to increase
2. By the end of adaptation, the numerical errors stop
3. Sampling continues after adaptation and `ratio_divergent_transitions` stays zero.
4. `ratio_divergent_transitions_during_adaption` starts to drop (even though adaptation is now finished!) until it reaches a lower value, eg 50% lower if there are equal numbers of samples during and after adaptation. 

I couldn't find the right location in the code, but I believe that `ratio_divergent_transitions_during_adaption` might be calculating the ratio using the total number of samples so far, and not stopping updating after adaptation completes.

devmotion:
Fixes #373.

### DIFF
--- a/src/abstractmcmc.jl
+++ b/src/abstractmcmc.jl
@@ -265,7 +265,8 @@ function (cb::HMCProgressCallback)(
 
     # Update progress meter
     if pm !== nothing
-        percentage_divergent_transitions = cb.num_divergent_transitions[] / max(i - n_adapts, 1)
+        percentage_divergent_transitions =
+            cb.num_divergent_transitions[] / max(i - n_adapts, 1)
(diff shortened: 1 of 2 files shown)


yebai (approved):
Thanks @devmotion -- please note this is minorlly breaking change.

PR #398, opened 2025-02-20
Reduce allocations of `pm_next!`

devmotion:


src/sampler.jl
```diff
@@ -94,7 +94,7 @@ end
 Progress meter update with all trajectory stats, iteration number and metric shown.
 """
 function pm_next!(pm, stat::NamedTuple)
-    ProgressMeter.next!(pm; showvalues = [tuple(s...) for s in pairs(stat)])
+    ProgressMeter.next!(pm; showvalues = map(tuple, values(stat), keys(stat)))
```
yebai, line 97: It looks good; why does the current code allocate more?

devmotion: It creates an array instead of a tuple.

PR #399, opened 2025-02-22
Fix fields with abstract types

devmotion:
I went through all `struct`s in AdvancedHMC and fixed all fields that are currently not concretely typed.


devmotion:
Bump 🙂

mhauru (commented):
Looks great. Any indication of how this affects compile time and performance? Happy to merge without benchmarking, but would be interested to hear if you've tried it out.

Needs a breaking version bump since many of the types are exported.

devmotion:
I updated the version number.

PR #401, opened 2025-03-10
Remove SimpleUnpack

ErikQQY:
Since there is a standard destructuring syntax in Julia, we can remove SimpleUnpack then.

This PR also set Julia compat to 1.10, together with LinearAlgebra, Random etc.


ext/AdvancedHMCMCMCChainsExt.jl
```diff
@@ -15,21 +15,21 @@ function AbstractMCMC.bundle_samples(
     sampler::AbstractMCMC.AbstractSampler,
     state,
     chain_type::Type{Chains};
-    discard_initial = 0,
-    thinning = 1,
-    param_names = missing,
-    bijector = identity,
+    discard_initial=0,
+    thinning=1,
```
yebai, line 19: @ErikQQY, consider avoiding these styling changes in PRs.  It's okay to do it via a separate PR if we decide to follow a different style.

Project.toml
```diff
@@ -36,17 +35,16 @@ DocStringExtensions = "0.8, 0.9"
...
-SimpleUnPack = "1.1"
 Statistics = "1.6"
 StatsBase = "0.31, 0.32, 0.33, 0.34"
 StatsFuns = "0.8, 0.9, 1"
-LinearAlgebra = "1.6"
-Random = "1.6"
-julia = "1.6.7"
+LinearAlgebra = "1.10"
+Random = "1.10"
+julia = "1.10"
```
yebai, line 47: @devmotion, will dropping Julia's pre-1.10 support cause an issue?

Project.toml
```diff
@@ -36,17 +35,16 @@ DocStringExtensions = "0.8, 0.9"
 LogDensityProblems = "2"
 LogDensityProblemsAD = "1"
 MCMCChains = "5, 6"
-OrdinaryDiffEq = "6"
+OrdinaryDiffEq = "6.92"
```
devmotion, line 38: This seems unnecessary?

```suggestion
OrdinaryDiffEq = "6"
```

Project.toml
```diff
@@ -36,17 +35,16 @@ DocStringExtensions = "0.8, 0.9"
...
 ProgressMeter = "1"
 Requires = "0.5, 1"
 Setfield = "0.7, 0.8, 1"
-SimpleUnPack = "1.1"
 Statistics = "1.6"
 StatsBase = "0.31, 0.32, 0.33, 0.34"
 StatsFuns = "0.8, 0.9, 1"
-LinearAlgebra = "1.6"
```
devmotion, line 47: It's unnecessary to update stdlib compats in this way. In fact, due to potential issues with test environments the current recommendation is to use
```suggestion
LinearAlgebra = "<0.1, 1"
```

devmotion (commented):
I'm fine with updating the min compat to 1.10. I don't see any need for it though - the package works fine on >= 1.6, and there's no add maintenance burden on < 1.10 currently. SimpleUnpack also just uses the Julia-native destructuring on Julia >= 1.7 (where it was introduced).

ext/AdvancedHMCMCMCChainsExt.jl
mhauru: I think autoformatting would make these same changes if we copied the `.JuliaFormatter.toml` from here: https://github.com/TuringLang/Turing.jl/blob/main/.JuliaFormatter.toml

I think that would be a good PR to make, but I agree that it should be a separate PR, because it's hard and a bit risky to read diffs that mix actual changes made by a human with autoformatting changes in unrelated parts of the codebase.

ErikQQY: Yeah, I was trying to format before push habitually but found out the formatting style has not been configured yet.

yebai:
@ErikQQY I'm happy with the changes. Can you separate the styling change and removal of `SimpleUnpack` into two PRs? In addition, can you please set up auto-formatting CI for AHMC?

src/adaptation/stepsize.jl
```diff
@@ -130,10 +130,13 @@ function adapt_stepsize!(
...
         # FIXME: this revert is buggy for batch mode
-        @unpack m, ϵ, x_bar, H_bar = state
+        (; m, ϵ, x_bar, H_bar) = state
     end
 
-    @pack! state = m, ϵ, x_bar, H_bar
+    state.m = m
+    state.ϵ = ϵ
+    state.x_bar = x_bar
+    state.H_bar = H_bar
```
devmotion, line 139: Does this change the return value from `state` to `H_bar`?

ErikQQY: Yes, the `DAState` is a mutable struct, so this is viable, but it’s not good for performance, we need to find a way to avoid such mutable struct if possible.

ErikQQY: Yeah, this is an inplace function, just return nothing

src/adaptation/stepsize.jl
```diff
@@ -88,7 +88,7 @@ NesterovDualAveraging(
     t_0::T,
     κ::T,
     δ::T,
-    ϵ::VT,
+    ϵ::VT
```
devmotion, line 91: Unrelated change it seems:
```suggestion
    ϵ::VT,
```

devmotion (approved):

yebai:
Thanks @ErikQQY!

PR #405, opened 2025-03-16
Allow ADTypes AD backend selection in Hamiltonian

ErikQQY:
While LogDensityProblemsAD.jl supports the AD backend from ADTypes.jl, we can also allow this for an unifying interface


ext/AdvancedHMCADTypesExt.jl
```diff
@@ -0,0 +1,41 @@
+module AdvancedHMCADTypesExt
+
+if isdefined(Base, :get_extension)
```
devmotion, line 3: This is unnecessary, it's always true on Julia >= 1.9.

ext/AdvancedHMCADTypesExt.jl
```diff
@@ -0,0 +1,41 @@
...
+    using AdvancedHMC:
+        AbstractMetric,
+        Hamiltonian,
+        LogDensityModel,
```
devmotion, line 7: It would be cleaner to directly import it from AbstractMCMC instead of relying on it being imported in AdvancedHMC.

ext/AdvancedHMCADTypesExt.jl
```diff
@@ -0,0 +1,41 @@
...
+        LogDensityProblems,
```
devmotion, line 8: The extension should depend on LogDensityProblems if it loads it, indirect imports should be avoided.

ext/AdvancedHMCADTypesExt.jl
```diff
@@ -0,0 +1,41 @@
...
+        LogDensityProblemsAD
```
devmotion, line 9: Same here, the extension should depend on LogDensityProblemsAD.

Project.toml
```diff
@@ -18,16 +18,19 @@ StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
...
 OrdinaryDiffEq = "1dea7af3-3e70-54e6-95c3-0bf5283fa5ed"
 
 [extensions]
+AdvancedHMCADTypesExt = "ADTypes"
 AdvancedHMCCUDAExt = "CUDA"
 AdvancedHMCMCMCChainsExt = "MCMCChains"
 AdvancedHMCOrdinaryDiffEqExt = "OrdinaryDiffEq"
 
 [compat]
+ADTypes = "1.14"
```
devmotion, line 33: I assume

```suggestion
ADTypes = "1"
```

would be sufficient? More relevant is here the LogDensityProblemsAD compat though - it has to be updated to the version that started to support ADTypes.

test/demo.jl
```diff
@@ -105,3 +105,58 @@ end
     @test "μ" ∈ labels
     @test "σ" ∈ labels
 end
+
+# -- test ADTypes compatibility
+
```
devmotion, line 110:
```suggestion
```

test/demo.jl
```diff
@@ -105,3 +105,58 @@ end
...
+@testset "Demo with ADTyes" begin
```
devmotion, line 111:
```suggestion
@testset "ADTyes" begin
```

test/demo.jl
```diff
@@ -105,3 +105,58 @@ end
...
+    # Define the target distribution using the `LogDensityProblem` interface
```
devmotion, line 112: Can't we just reuse existing models? The main thing to test is just that you can use (different) ADTypes instead of Val to specify the backend. It shouldn't be necessary to design a custom model.

test/demo.jl
```diff
@@ -105,3 +105,58 @@ end
...
+    D = 10
+    initial_θ = rand(D)
+    ℓπ = DemoProblem(D)
+
+    # Set the number of samples to draw and warmup iterations
+    n_samples, n_adapts = 2_000, 1_000
+
+    # Define a Hamiltonian system
+    metric = DiagEuclideanMetric(D)
+    hamiltonian = Hamiltonian(metric, ℓπ, AutoForwardDiff())
```
devmotion, line 132: This is the only relevant part in the test, and it misses the case where the logdensity is provided as a LogDensityModel.

Project.toml
ErikQQY: Yeah, setting ADTypes to "1" would be sufficient

ext/AdvancedHMCADTypesExt.jl
```diff
@@ -0,0 +1,26 @@
...
+end
+function Hamiltonian(metric::AbstractMetric, ℓπ, kind::AbstractADType; kwargs...)
+    if LogDensityProblems.capabilities(ℓπ) === nothing
+        throw(
+            ArgumentError(
+                "The log density function does not support the LogDensityProblems.jl interface",
+            ),
+        )
+    end
+    ℓ = LogDensityProblemsAD.ADgradient(kind, ℓπ; kwargs...)
```
yebai, line 22: Let's use `DifferentiationInterface` to stay consistent with the rest of the TuringLang ecosystem. You could try to use `AbstractMCMC.LogDensityModel` and define `LogDensityProblem` interface on it in a similar way to [`DynamicPPL.LogDensityFunction`](https://github.com/TuringLang/DynamicPPL.jl/blob/main/src/logdensityfunction.jl).

test/demo.jl
```diff
@@ -1,6 +1,6 @@
 using ReTest
-using AdvancedHMC, Distributions, ForwardDiff, ComponentArrays, AbstractMCMC
-using LinearAlgebra
+using AdvancedHMC, Distributions, ForwardDiff, Zygote, ComponentArrays, AbstractMCMC
+using LinearAlgebra, ADTypes, LogDensityProblemsAD
```
yebai, line 3: Let's replace LDP with DI. 

```suggestion
using LinearAlgebra, ADTypes, DifferentiationInterface
```

ext/AdvancedHMCADTypesExt.jl
yebai: Yes, I am aware of that. DynamicPPL has dropped `LogDensityProblemsAD` to use `DifferentiationInterface` directly at suggestions by @willtebbutt

ErikQQY: Now DifferentiationInterface is supported

src/AdvancedHMC.jl
```diff
@@ -146,7 +152,12 @@ function Hamiltonian(metric::AbstractMetric, ℓ; kwargs...)
     ℓπ = if cap === LogDensityProblems.LogDensityOrder{0}()
         # In this case ℓ does not support evaluation of the gradient of the log density function
         # We use ForwardDiff to compute the gradient
-        LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓ; kwargs...)
+        _logdensity = Base.Fix1(LogDensityProblems.logdensity, ℓ)
+        _logdensity_and_gradient = function (x)
+            prep = DI.prepare_gradient(_logdensity, adtype, x)
```
gdalle, line 157: Preparing outside the closure would be more efficient, but users have to be warned I guess

src/AdvancedHMC.jl
```diff
@@ -146,7 +152,12 @@ function Hamiltonian(metric::AbstractMetric, ℓ; kwargs...)
     ℓπ = if cap === LogDensityProblems.LogDensityOrder{0}()
         # In this case ℓ does not support evaluation of the gradient of the log density function
         # We use ForwardDiff to compute the gradient
-        LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓ; kwargs...)
+        _logdensity = Base.Fix1(LogDensityProblems.logdensity, ℓ)
```
gdalle, line 155: One can also use `DI.Constant` for `ℓ` if it contains no differentiable storage

src/AdvancedHMC.jl
```diff
@@ -158,26 +169,6 @@ function Hamiltonian(metric::AbstractMetric, ℓ; kwargs...)
     )
 end
 
-## With explicit AD specification
-function Hamiltonian(
-    metric::AbstractMetric, ℓπ::LogDensityModel, kind::Union{Symbol,Val,Module}; kwargs...
-)
```
gdalle, line 164: I guess removing this is a breaking change?

test/contrib.jl
```diff
@@ -4,8 +4,8 @@ using ReTest, AdvancedHMC, ForwardDiff, Zygote
     @testset "ad" begin
         metric = UnitEuclideanMetric(D)
         h_hand = Hamiltonian(metric, ℓπ, ∂ℓπ∂θ)
-        h_forwarddiff = Hamiltonian(metric, ℓπ, ForwardDiff)
-        h_zygote = Hamiltonian(metric, ℓπ, Zygote)
+        h_forwarddiff = Hamiltonian(metric, ℓπ; adtpye=AutoForwardDiff())
```
gdalle, line 7:
```suggestion
        h_forwarddiff = Hamiltonian(metric, ℓπ; adtype=AutoForwardDiff())
```

test/demo.jl
```diff
@@ -105,3 +105,37 @@ end
     @test "μ" ∈ labels
     @test "σ" ∈ labels
 end
+
+@testset "ADTyes" begin
```
gdalle, line 109:
```suggestion
@testset "ADTypes" begin
```

gdalle (changes requested):

src/AdvancedHMC.jl
ErikQQY: I agree we should prepare outside the closure, but since we need the log density and its gradient of the model as a function in Hamiltonian, how can we prepare in advance?

gdalle: just move the `prep` one line up?

gdalle: Do you have access to a "typical input" somewhere?

ErikQQY: AFAIK, the `Hamiltonian` is basically used for keeping the log density and its gradient function, and "typical input" is used in the subsequent sampling process which may not be possible to pass in here.

Project.toml
```diff
@@ -3,14 +3,16 @@ uuid = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
...
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
 ArgCheck = "dce04be8-c92d-5529-be00-80e4d2c0e197"
+DifferentiationInterface = "a0c0ee7d-e4b9-4e03-894e-1c5f64a51d63"
 DocStringExtensions = "ffbed154-4ef7-542d-bbb7-c09d3a79fcae"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 LogDensityProblems = "6fdf6af0-433a-55f7-b3ed-c6c6e0b8df7c"
-LogDensityProblemsAD = "996a588d-648d-4e1f-a8f0-a84b347e47b1"
 ProgressMeter = "92933f4c-e287-5a05-a399-4b506db050ca"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
```
devmotion, line 15: Unconditional reexports are a code smell IMO - why can't users just load ADTypes? By unconditional reexporting everything you completely lose control over what's defined in your package and automatically the whole API of the reexported package (and hence all breaking changes etc.) become part of the API of your package.

src/AdvancedHMC.jl
```diff
@@ -130,10 +133,12 @@ export HMCSampler, HMC, NUTS, HMCDA
...
 ## Without explicit AD backend
-function Hamiltonian(metric::AbstractMetric, ℓ::LogDensityModel; kwargs...)
-    return Hamiltonian(metric, ℓ.logdensity; kwargs...)
+function Hamiltonian(
+    metric::AbstractMetric, ℓ::LogDensityModel; adtype=AutoForwardDiff(), kwargs...
+)
+    return Hamiltonian(metric, ℓ.logdensity; adtype, kwargs...)
 end
-function Hamiltonian(metric::AbstractMetric, ℓ; kwargs...)
+function Hamiltonian(metric::AbstractMetric, ℓ; adtype=AutoForwardDiff(), kwargs...)
```
devmotion, line 141:
```suggestion
function Hamiltonian(metric::AbstractMetric, ℓ; adtype::ADType=AutoForwardDiff(), kwargs...)
```

src/AdvancedHMC.jl
```diff
@@ -130,10 +133,12 @@ export HMCSampler, HMC, NUTS, HMCDA
 include("abstractmcmc.jl")
 
 ## Without explicit AD backend
-function Hamiltonian(metric::AbstractMetric, ℓ::LogDensityModel; kwargs...)
-    return Hamiltonian(metric, ℓ.logdensity; kwargs...)
+function Hamiltonian(
+    metric::AbstractMetric, ℓ::LogDensityModel; adtype=AutoForwardDiff(), kwargs...
```
devmotion, line 137:
```suggestion
    metric::AbstractMetric, ℓ::LogDensityModel; adtype::ADType=AutoForwardDiff(), kwargs...
```

src/AdvancedHMC.jl
```diff
@@ -146,7 +151,12 @@ function Hamiltonian(metric::AbstractMetric, ℓ; kwargs...)
     ℓπ = if cap === LogDensityProblems.LogDensityOrder{0}()
         # In this case ℓ does not support evaluation of the gradient of the log density function
         # We use ForwardDiff to compute the gradient
-        LogDensityProblemsAD.ADgradient(Val(:ForwardDiff), ℓ; kwargs...)
```
devmotion, line 149: Why not just
```suggestion
        LogDensityProblemsAD.ADgradient(adtype, ℓ; kwargs...)
```
?

src/AdvancedHMC.jl
```diff
@@ -146,7 +151,12 @@ function Hamiltonian(metric::AbstractMetric, ℓ; kwargs...)
...
+        _logdensity = Base.Fix1(LogDensityProblems.logdensity, ℓ)
+        _logdensity_and_gradient = function (x)
+            prep = DI.prepare_gradient(_logdensity, adtype, x)
```
devmotion, line 156: What's the point of the preparation step in this function? Every time you'd evaluate logdensity + gradient for an input `x`, it would run preparation for `x` and then immediately discard it after the computation. If you just use `LogDensityProblemsAD.ADgradient` (see above), then users could provide a typical input `x` as keyword argument to optimize the AD computations of the hamiltonian (or otherwise they'd get the slower fallback).

src/AdvancedHMC.jl
```diff
@@ -158,26 +168,6 @@ function Hamiltonian(metric::AbstractMetric, ℓ; kwargs...)
     )
 end
 
-## With explicit AD specification
-function Hamiltonian(
-    metric::AbstractMetric, ℓπ::LogDensityModel, kind::Union{Symbol,Val,Module}; kwargs...
-)
```
devmotion, line 164: No need to remove these?

src/AdvancedHMC.jl
gdalle:
```suggestion
function Hamiltonian(metric::AbstractMetric, ℓ; adtype=ADTypes.AutoForwardDiff(), kwargs...)
```

src/AdvancedHMC.jl
gdalle:
```suggestion
    metric::AbstractMetric, ℓ::LogDensityModel; adtype=ADTypes.AutoForwardDiff(), kwargs...
```

src/AdvancedHMC.jl
gdalle: If preparation is not reused, then it is not beneficial. But as per my [comment above](https://github.com/TuringLang/AdvancedHMC.jl/pull/405#discussion_r2008703659), if it can be reused because we have access to a typical input, then it works basically the same as a `LogDensityProblemsAD.ADgradient` (except that Turing no longer depends on LogDensityProblemsAD` IIUC)

Project.toml
ErikQQY: Yeah, users who want to try other AD backends should just load ADTypes by themselves then

src/AdvancedHMC.jl
ErikQQY: Changing to DifferentiationInterface is just a step forward for the mutualized AD maintenance in Turing, and seems if we want to use DI to compute the gradient, we will have to use the preparation mechanism. As for the performance aspects, I will test the difference between using DifferentiationInterface and LogDensityProblemsAD, if no difference between those two options, we can safely change to DI then

yebai:
@ErikQQY, let's only introduce `ADTypes` support in this PR and switch to `DifferentiationInterface` via a separate PR.

ErikQQY:
As per @yebai's suggestion, let's only focus on ADTypes support in this PR and integrate DifferentiationInterface in another PR

ext/AdvancedHMCADTypesExt.jl
```diff
@@ -0,0 +1,25 @@
+module AdvancedHMCADTypesExt
+
+using AdvancedHMC: AbstractMetric, Hamiltonian, LogDensityModel
+using LogDensityProblems: LogDensityProblems
+using LogDensityProblemsAD: LogDensityProblemsAD
```
devmotion, line 5: These have to be dependencies of the extension or imported via AdvancedHMC.

devmotion (approved):

PR #407, opened 2025-03-17
Add ComponentArrays extension

ErikQQY:
Replace: #345 
Fix: #344 

So in  #344, the reason why ComponentArray failed with `DenseEuclideanMetric` is that the dense matrix in `DenseEuclideanMetric` doesn't have the same axes as the input, and when we do linear algebra operations on two arrays with different axes, the result is just an ordinary Julia array, for example:

```julia
julia> r_init = ComponentArray(; a=randn(5), b=randn(5))
ComponentVector{Float64}(a = [-0.8058038744185181, -0.08413831281388894, 0.7453256076416268, -0.03469137039683826, -0.2517275575798834], b = [-1.5630129100605425, 1.7031219617336555, -1.0997337141360415, 0.4409537319539482, 0.19944559484881833])

julia> getaxes(r_init)
(Axis(a = ViewAxis(1:5, Shaped1DAxis((5,))), b = ViewAxis(6:10, Shaped1DAxis((5,)))),)

julia> M = rand(10, 10);

julia> axes(M)
(Base.OneTo(10), Base.OneTo(10))

julia> M * r_init # clearly the axes of M and r_init don't match
10-element Vector{Float64}:
 -0.06957219975578224
 -0.24499462421604584
  0.26535313109443504
 -0.05499884350331631
 -0.11871623121978324
 -1.132138726898246
 -0.6221912360430668
 -0.7051768915767198
  1.3780763510373664
 -0.9263200146682737
```

But it's basically user's obligation to ensure consistent input data structure, so this PR ensures the input from user are axes-matching.

Some notes: 
- See the tests for [example uses](https://github.com/TuringLang/AdvancedHMC.jl/blob/459ebb8a10cc1bc7dbbc27ed79afa796c607697a/test/hamiltonian.jl#L77-L100) of `ComponentArrays`. 
- The code changes in `src/adaptation/massmatrix.jl` are not related to fixing `ComponentArrays`.


yebai (commented):
I'm not familiar with `ComponentArrays`. Can @mhauru, @penelopeysm, or @sunxd3 take a look, please?

ext/AdvancedHMCComponentArraysExt.jl
```diff
@@ -0,0 +1,27 @@
...
+function AdvancedHMC.∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric,<:GaussianKinetic}, r::ComponentArray)
+    out = similar(r)
+    out .= h.metric.M⁻¹ .* r
+    return out
+end
+function AdvancedHMC.∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::ComponentArray)
+    out = similar(r)
+    mul!(out, h.metric.M⁻¹, r)
+    return out
+end
```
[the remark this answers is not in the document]
yebai: We might be able to use `AbstractVector` (i.e. `AbstractArray{T, 1} where T`) here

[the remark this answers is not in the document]
ErikQQY: Yeah, instead of having separate extensions for ComponentArrays.jl, we can fix this issue by improving to more generic definitions

src/hamiltonian.jl
```diff
@@ -41,12 +41,16 @@ function ∂H∂θ(h::Hamiltonian, θ::AbstractVecOrMat)
     return DualValue(res[1], -res[2])
 end
 
-∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) = copy(r)
-function ∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat)
-    return h.metric.M⁻¹ .* r
+∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric,<:GaussianKinetic}, r::AbstractArray) = copy(r)
```
devmotion, line 44: This is probably not necessary?
```suggestion
∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) = copy(r)
```

src/hamiltonian.jl
```diff
@@ -41,12 +41,16 @@ function ∂H∂θ(h::Hamiltonian, θ::AbstractVecOrMat)
...
+function ∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric,<:GaussianKinetic}, r::AbstractArray)
+    out = similar(r)
+    out .= h.metric.M⁻¹ .* r
+    return out
```
devmotion, line 48: According to the issue only `DenseEuclideanMetric` was problematic. Was this also broken? If not, the current code is actually better IMO - it avoids changing from e.g. `SMatrix` to `MMatrix` and the new code would be broken for vectors with element types that the result can't be written into (probably e.g. `Vector{Int}`). So generally IMO the old code is better if it works.

src/hamiltonian.jl
```diff
@@ -41,12 +41,16 @@ function ∂H∂θ(h::Hamiltonian, θ::AbstractVecOrMat)
...
 end
-function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat)
-    return h.metric.M⁻¹ * r
+function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::AbstractArray)
```
devmotion, line 50:
```suggestion
function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat)
```

src/hamiltonian.jl
```diff
@@ -41,12 +41,16 @@ function ∂H∂θ(h::Hamiltonian, θ::AbstractVecOrMat)
...
+    out = similar(r)
+    mul!(out, h.metric.M⁻¹, r)
```
devmotion, line 52: Generally this is not sufficient, you need something like
```suggestion
    (; M⁻¹) = h.metric
    out = similar(r, promote_type(eltype(r),eltype(M⁻¹)))
    mul!(out, M⁻¹, r)
```

But the question is still: Why does the previous code - which in principle is much nicer since it doesn't require type heuristics - not work? Is it an upstream bug in ComponentArrays? Or because the hamiltonian was set up incorrectly?

ErikQQY: I believe the main issue here is that when we try to compute the product `Matrix{Float64}*ComponentVector{Float64}`, we would have `Vector{Float64}`, so I think maybe it's a ComponentArrays.jl thing that missing dispatches for such linear algebra operations?

```julia
julia> x = ComponentArray(a=5, b=[(a=20., b=0), (a=33., b=0), (a=44., b=3)], c=c)
ComponentVector{Float64}(a = 5.0, b = ComponentVector{Float64, SubArray{Float64, 1, Vector{Float64}, Tuple{UnitRange{Int64}}, true}, Tuple{Axis{(a = 1, b = 2)}}}[(a = 20.0, b = 0.0), (a = 33.0, b = 0.0), (a = 44.0, b = 3.0)], c = (a = 2.0, b = [1.0, 2.0]))

julia> rand(10,10)*x
10-element Vector{Float64}:
 53.280246927016925
 57.22007605159049
 49.206319389153876
 63.135902809337765
 50.88459580133911
 73.58796576720756
 46.62055260143546
 44.463608176249025
 68.52373868783468
 46.42311997871232
```

src/hamiltonian.jl
ErikQQY: Yeah, in the diagonal case, since the mass matrix is only represented using a Vector, so the original way is better

test/hamiltonian.jl
```diff
@@ -68,3 +69,32 @@ end
...
+
+            M⁻¹ = ones(T, 2 * D) + abs.(randn(T, 2 * D))
+            h = Hamiltonian(DiagEuclideanMetric(M⁻¹), ℓπ, ∂ℓπ∂θ)
+            @test -AdvancedHMC.neg_energy(h, r_init, θ_init) ≈
+                r_init' * diagm(0 => M⁻¹) * r_init / 2
+            @test AdvancedHMC.∂H∂r(h, r_init) == M⁻¹ .* r_init
+            @test typeof(AdvancedHMC.∂H∂r(h, r_init)) == typeof(r_init)
+
+            m = randn(T, 2 * D, 2 * D)
+            M⁻¹ = m' * m
```
devmotion, line 93: Isn't this the (only?) problem - if you're working with ComponentArrays, the Hamiltonian should also be constructed with a ComponentArray as matrix (with the correct axes), to exactly avoid this matrix multiplication issue? IMO users are just lucky that the other examples work (due to special broadcasting behaviour) but in principle also the DiagEuclideanMetric should be constructed with a ComponentArray (since equivalently - and probably without performance regression - one could apply it by multiplication with a `Diagonal` matrix).

So I actually think it seems to be a user issue and the only problem is that the Hamiltonian is not defined correctly.

devmotion: > thinking of mass matrix being a ComponentMatrix is bit of weird

I'd disagree, I'd say that's how you have to think about your mass matrix if your samples etc are ComponentArrays. For instance, in the example here there's an `a` and a `b` component of theta and r. So also the mass matrix (be it dense, diagonal or uniform) needs to consist of `a` and `b` components.

So I'd argue, instead of "fixing" the dense case an error should be thrown for the diagonal (and unit?) case as well. For instance, maybe it would be possible to check upfront that the axes of the mass matrices match the axis of sample, momentum etc and throw a descriptive error otherwise that makes it easier for users to realize that their Hamiltonian is incorrect.

ErikQQY: Fair point, I will add these axes checks and warnings when constructing Hamiltonian with ComponnetArrays.jl

ErikQQY: Okay, AdvancedHMC now has axes checks and warnings when dealing with ComponentArrays

src/hamiltonian.jl
```diff
@@ -61,7 +61,7 @@ struct PhasePoint{T<:AbstractVecOrMat{<:AbstractFloat},V<:DualValue}
     ℓπ::V # Cached neg potential energy for the current θ.
     ℓκ::V # Cached neg kinect energy for the current r.
     function PhasePoint(θ::T, r::T, ℓπ::V, ℓκ::V) where {T,V}
-        @argcheck length(θ) == length(r) == length(ℓπ.gradient) == length(ℓπ.gradient)
+        @argcheck length(θ) == length(r) == length(ℓπ.gradient) == length(ℓκ.gradient)
```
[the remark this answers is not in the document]
ErikQQY: Fixed and tested in #426

src/adaptation/massmatrix.jl
```diff
@@ -171,13 +172,13 @@ function get_estimation(nc::NaiveCov)
 end
 
 # Ref: https://github.com/stan-dev/math/blob/develop/stan/math/prim/mat/fun/welford_covar_estimator.hpp
-mutable struct WelfordCov{F<:AbstractFloat} <: DenseMatrixEstimator{F}
+mutable struct WelfordCov{F<:AbstractFloat,C} <: DenseMatrixEstimator{F}
```
yebai, line 175: For clarity, perhaps 

```suggestion
mutable struct WelfordCov{F<:AbstractFloat,C<:Matrix{F}} <: DenseMatrixEstimator{F}
```

ext/AdvancedHMCComponentArraysExt.jl
```diff
@@ -0,0 +1,42 @@
...
+function AdvancedHMC.∂H∂r(
+    h::Hamiltonian{<:UnitEuclideanMetric,<:GaussianKinetic}, r::ComponentVecOrMat
+)
+    return copy(r)
+end
+function AdvancedHMC.∂H∂r(
+    h::Hamiltonian{<:DiagEuclideanMetric,<:GaussianKinetic}, r::ComponentVecOrMat
+)
+    (; M⁻¹) = h.metric
+    @assert getaxes(M⁻¹) == getaxes(r) "Axes of mass matrix and momentum must match"
```
devmotion, line 23: Two comments:
- Assertions may be removed by the compiler, IMO it's bad style to use them for actual error checks (I only use them eg to reassert myself - or others - that an `else` branch satisfies certain properties where I know that the assertion should be guaranteed to always be satisfied).
- This requirement seems to be required more generally (and probably be checked internally in LinearAlgebra), so it could just be moved to the generic definition for AbstractVector?

ext/AdvancedHMCComponentArraysExt.jl
```diff
@@ -0,0 +1,42 @@
...
+    return h.metric.M⁻¹ .* r
+end
+function AdvancedHMC.∂H∂r(
+    h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::ComponentVector
+)
+    (; M⁻¹) = h.metric
+    @assert (first(getaxes(M⁻¹)) == first(getaxes(r))) &&
+        (last(getaxes(M⁻¹)) == first(getaxes(r))) "Axes of mass matrix and momentum must match"
```
devmotion, line 31: Same comments as above.

ext/AdvancedHMCComponentArraysExt.jl
```diff
@@ -0,0 +1,42 @@
...
+    return h.metric.M⁻¹ * r
+end
+function AdvancedHMC.∂H∂r(
+    h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::ComponentMatrix
+)
+    (; M⁻¹) = h.metric
+    @assert getaxes(M⁻¹) == getaxes(r) "Axes of mass matrix and momentum must match"
```
devmotion, line 38: And here.

ext/AdvancedHMCComponentArraysExt.jl
ErikQQY: > so it could just be moved to the generic definition for AbstractVector?

What do you mean by moving to the generic definition for AbstractVector? We only need to check whether the axes are matching for ComponentArrays.

devmotion: The matching axes are a property that's generally required for most generic matrix multiplications in Julia anyway. It's not really special about ComponentArrays IMO. It's just that it surfaces here since in all other cases that are tested the axes are matching.

ErikQQY: Actually, it is user's responsibility to input axes-matching ComponentArrays, here for example, axes-matching mass-matrix and momentum, to avoid `Matrix{Float64}*ComponentVector{Float64}=Vector{Float64}` issues. The extension here is to help users check if the axes match instead of allowing the above behaviour.

devmotion: Sure, but why not help users more generally, regardless of whether they use ComponentArrays, regular arrays, or other custom array types?

ErikQQY: Oh, I see, I will add the axes checks for general arrays as well

ext/AdvancedHMCComponentArraysExt.jl
```diff
@@ -0,0 +1,39 @@
...
+using AdvancedHMC:
+    AdvancedHMC,
+    Hamiltonian,
+    UnitEuclideanMetric,
+    DiagEuclideanMetric,
+    DenseEuclideanMetric,
+    GaussianKinetic
+using ComponentArrays: ComponentVecOrMat, ComponentVector, ComponentMatrix, getaxes
+
+function AdvancedHMC.∂H∂r(
```
yebai, line 14: We don't need to overload these functions `∂H∂r`. It seems all we need is to define a utility function `AHMC.__axes`, then overload `AHMC.__axes(x::ComponentVecOrMat) = getaxes(x)` in this `AHMCComponentArrays` extension.

src/hamiltonian.jl
```diff
@@ -43,9 +43,15 @@ end
 
 ∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) = copy(r)
 function ∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat)
+    (; M⁻¹) = h.metric
+    (first(axes(M⁻¹)) !== first(axes(r))) &&
```
yebai, line 47: See above. 

```suggestion
    (first(__axes(M⁻¹)) !== first(axes(r))) &&
```

src/hamiltonian.jl
```diff
@@ -43,9 +43,15 @@ end
```
yebai, line 43: See above. 

```suggestion
__axes(x::AbstractVecOrMat) = axes(x)
```

src/hamiltonian.jl
```diff
@@ -43,9 +43,15 @@ end
 
 ∂H∂r(h::Hamiltonian{<:UnitEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat) = copy(r)
 function ∂H∂r(h::Hamiltonian{<:DiagEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat)
+    (; M⁻¹) = h.metric
+    (first(__axes(M⁻¹)) !== first(__axes(r))) &&
+        throw(ArgumentError("Axes of mass matrix and momentum must match"))
     return h.metric.M⁻¹ .* r
```
yebai, line 49:
```suggestion
    return M⁻¹ .* r
```

src/hamiltonian.jl
```diff
@@ -43,9 +43,15 @@ end
...
 end
 function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat)
+    (; M⁻¹) = h.metric
+    (last(__axes(M⁻¹)) !== first(__axes(r))) &&
+        throw(ArgumentError("Axes of mass matrix and momentum must match"))
     return h.metric.M⁻¹ * r
```
yebai, line 55:
```suggestion
    return M⁻¹ * r
```

src/utilities.jl
```diff
@@ -22,6 +22,13 @@ function _randn(
     return out
 end
 
+"""
+    __axes(r::AbstractVecOrMat)
+
+Return the axes of input `r` where `r` can be generic arrays or custom arrays.
```
yebai, line 28:
```suggestion
Return the axes of input `r`, where `r` can be `AbstractArrays` or  `ComponentArrays`. 
```

yebai (approved):
Excellent work, @ErikQQY! 

I only checked the main code (`/src/`) but didn't look at the tests carefully.

yebai:
@penelopeysm, can you also review this please?

src/hamiltonian.jl
```diff
@@ -43,10 +43,16 @@ end
...
+    (first(__axes(M⁻¹)) !== first(__axes(r))) &&
+        throw(ArgumentError("Axes of mass matrix and momentum must match"))
+    return M⁻¹ .* r
 end
 function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat)
-    return h.metric.M⁻¹ * r
+    (; M⁻¹) = h.metric
+    (last(__axes(M⁻¹)) !== first(__axes(r))) &&
+        throw(ArgumentError("Axes of mass matrix and momentum must match"))
+    return M⁻¹ * r
```
penelopeysm, line 55: Fairly minor comment, when getting errors like these it's often really useful to know what the problematic axis dimensions were, so if the dimensions could be printed that would be great. Also I think it'd be great to be really precise - you could say 'first dimension of ...' or 'last dimension of ...' must match rather than just 'axes'.

penelopeysm (changes requested):
Very happy apart from the single comment!

penelopeysm (approved):

yebai:
Excellent work @ErikQQY!

PR #409, opened 2025-03-17
Remove Requires dependency

devmotion:
It's not needed anymore since Julia >= 1.10 supports extensions.


yebai:
Thanks @devmotion!

PR #416, opened 2025-04-01
Integrate DifferentiationInterface

ErikQQY:
Continue #405


Project.toml
```diff
@@ -3,8 +3,10 @@ uuid = "0bf59076-c3b1-5ca4-86bd-e02cd72cde3d"
 version = "0.7.0"
```
penelopeysm, line 3: 1. It looks to me that the changes here (and in the previous ADTypes PR) are breaking, so this should be incremented to 0.8

2. Could you also update the calls to `Hamiltonian(..., ..., ForwardDiff)` in the README?

src/AdvancedHMC.jl
```diff
@@ -160,22 +167,21 @@ end
...
             ),
         )
     end
-    ℓ = LogDensityProblemsAD.ADgradient(
-        kind isa Val ? kind : Val(Symbol(kind)), ℓπ; kwargs...
-    )
-    return Hamiltonian(metric, ℓ)
+    _logdensity = Base.Fix1(LogDensityProblems.logdensity, ℓπ)
+    _logdensity_and_gradient = x -> DI.value_and_gradient(_logdensity, kind, x)
+    return Hamiltonian(metric, _logdensity, _logdensity_and_gradient)
```
penelopeysm, line 184: More generally, it seems to me that some of this code is repeated from above.

I think it makes sense to do the capability checks etc. inside this method, and have the `kind`-less method (on line 139) just forward to this, i.e. change it to:

```julia
Hamiltonian(metric::AbstractMetric, ℓ; kwargs...) = Hamiltonian(metric, ℓ, AutoForwardDiff(); kwargs...)
```

I also generally think the `kind` parameter can be renamed to `adtype`, it's more descriptive and since it's a positional argument it won't cause undue breakage.

penelopeysm (changes requested):
Hello @ErikQQY & thanks for the PR! Some comments below.

src/AdvancedHMC.jl
ErikQQY: Yeah indeed, we can just reuse the explicit AD branch to avoid repetition.

Project.toml
devmotion: AFAICT #405 is non-breaking (it just adds a new currently not existing constructor), so it could be released in a non-breaking 0.7.x release.

penelopeysm: Oh, yes, that's true

src/AdvancedHMC.jl
```diff
@@ -134,48 +138,32 @@ function Hamiltonian(metric::AbstractMetric, ℓ::LogDensityModel; kwargs...)
...
         )
     end
-    ℓ = LogDensityProblemsAD.ADgradient(
-        kind isa Val ? kind : Val(Symbol(kind)), ℓπ; kwargs...
-    )
-    return Hamiltonian(metric, ℓ)
+    _logdensity = Base.Fix1(LogDensityProblems.logdensity, ℓπ)
+    _logdensity_and_gradient = if cap === LogDensityProblems.LogDensityOrder{0}()
+        # In this case ℓπ does not support evaluation of the gradient of the log density function
+        x -> DI.value_and_gradient(_logdensity, adtype, x)
```
devmotion, line 162: It's still unclear to me why it should be advantageous to avoid LogDensityProblemsAD... But whatever - if you'd like to remove it, you should copy all the things that `LogDensityProblemsAD.ADgradient` does automatically to ensure good compatibility with all ADTypes. For instance, Enzyme does not like `Base.FixXX`. So this should in principle be
```julia
        x -> DI.value_and_gradient(LogDensityProblems.logdensity, adtype, DI.Constant(ℓπ), x)
```
But then DI doesn't allow active after non-active arguments so you have to use something like
```suggested
        x -> DI.value_and_gradient(switched_logdensity, adtype, x, DI.Constant(ℓπ))
```
where
```julia
switched_logdensity(x, f::F) = LogDensityProblems.logdensity(f, x)
```
LogDensityProblemsAD does this already for you... And it even allows you to specify a typical `x` which would vastly improve performance for some backends - which this approach doesn't. So I'm still not convinced 😄

yebai: @gdalle, is there any plan to incorporate these optimisations in `LogDensityProblemsAD` into `DI`?

gdalle: > It's still unclear to me why it should be advantageous to avoid LogDensityProblemsAD... 

The goal is probably to rely on all the AD bindings being in a single place, aka DI, for mutualized maintenance, testing and updates. LogDensityProblemsAD is a set of AD bindings that was specific to the probabilistic programming lanscape, while DI aims to support the whole Julia ecosystem. As a result, when someone needs a feature, everyone gets it.

> And it even allows you to specify a typical x which would vastly improve performance for some backends - which this approach doesn't. So I'm still not convinced 

Specifying a typical `x` to speed things up is the whole point of DI's preparation mechanism (which LogDensityProblemsAD also relies on), so it is indeed supported. Whether downstream packages are able to provide such an `x` is a different issue, but not an API limitation.

> @gdalle, is there any plan to incorporate these optimisations in LogDensityProblemsAD into DI?

No, changing the DI API would be enormously breaking now that around 850 packages depend on it, directly or indirectly. For the foreseeable future it will keep taking the active argument first, and the context arguments next.

I'm not claiming this is the best design choice for each individual application. And indeed it seems you need to add something like these lines in your specific case:

```julia
struct SwitchArgs{F}
    f::F
end

(s::SwitchArgs)(x, y) = s.f(y, x)
```

It's up to users to decide whether this (minor?) inconvenience is worth the switch to DI.

yebai: Thanks, @gdalle. The `SwitchArgs {F}` trick looks like a reasonable workaround.

gdalle: Although for Enzyme specifically David's solution is probably better, since it avoids creating a functor. Either way it's a one-liner

devmotion: > Specifying a typical x to speed things up is the whole point of DI's preparation mechanism (which LogDensityProblemsAD also relies on), so it is indeed supported. Whether downstream packages are able to provide such an x is a different issue, but not an API limitation.

I don't know why this turned into a discussion of DI, my comment was not about the DI API at all (which I know very well). The only point I elaborated on was that I'm surprised about the goal of removing LogDensityProblemsAD and setting up LogDensityProblems with DI and ADTypes manually given that the current approach in this PR seems inferior and improving it would mean copying code from LogDensityProblemsAD.

penelopeysm: Even prior to this PR, `ADgradient` isn't instantiated with a typical input anyway, so there's not much loss there (And in practice when AdvancedHMC is used with Turing models, the Turing model will store the gradient prep so AdvancedHMC won't need to recalculate it.)

The point about using a closure vs DI.Constant is something that we looked at in DynamicPPL as well (see the resulting code in https://github.com/TuringLang/DynamicPPL.jl/blob/eed80e5901cc641e286bd923a628e0bf488fc676/src/logdensityfunction.jl#L252-L280). Basically, I ran some benchmarks to determine whether it's better to use a closure vs a constant and for backends the Constant wins out, except for ReverseDiff where the closure wins. You could adopt a similar approach here if you wanted.

In doing so, basically what you'd be doing is taking on responsibility for these performance optimisations, and in return you'd cut out the library in the middle (LDPAD). Whether that's better or not is up to you to decide (in the case of DynamicPPL/Turing, there were code simplifications that made it clearly a good choice; for AdvancedHMC it seems a bit less obvious to me how much there is to be gained).

gdalle: FYI the reason why a closure is faster with ReverseDiff is because it cannot store constants in the tape that get modified at a later date. Now if we really wanted to avoid closures everywhere we could define a new DI context called `ConstantConstant`, which does not vary from preparation to execution. That should be distinguished from the currently existing `(Variable)Constant`, which can be modified.

devmotion: > Even prior to this PR, `ADgradient` isn't instantiated with a typical input anyway, so there's not much loss there

That's not true, this PR is actually a regression: Prior to this change here you could do e.g.
```julia
hamiltonian = Hamiltonian(metric, ℓπ, AutoForwardDiff(); x = zeros(5))
```
to construct a `Hamiltonian` whose gradient calculations are optimized for inputs of the type and length of `zeros(5)`.

devmotion: > The point about using a closure vs DI.Constant is something that we looked at in DynamicPPL as well (see the resulting code in https://github.com/TuringLang/DynamicPPL.jl/blob/eed80e5901cc641e286bd923a628e0bf488fc676/src/logdensityfunction.jl#L252-L280). Basically, I ran some benchmarks to determine whether it's better to use a closure vs a constant and for a few backends the Constant wins out, except for ReverseDiff where the closure wins. You could adopt a similar approach here if you wanted.

Actually, if you'd continue using LogDensityProblemsAD, you'd get these AD-specific optimizations automatically: For AD types that are supported via DI it will use `DI.Constant` (and prepare the gradient cache if possible), and for others that are natively supported such as e.g. `AutoReverseDiff` it will use an optimized approach such as a `Base.Fix1` closure: https://github.com/tpapp/LogDensityProblemsAD.jl/blob/ee6f9fef6ee550cb35a3bd3abdbd860e342e7b60/ext/LogDensityProblemsADReverseDiffExt.jl#L70 https://github.com/tpapp/LogDensityProblemsAD.jl/blob/ee6f9fef6ee550cb35a3bd3abdbd860e342e7b60/ext/LogDensityProblemsADReverseDiffExt.jl#L54

gdalle: Since these kinds of tweaks are beneficial to several users, maybe it makes sense to integrate them to DI, i.e. with the `ConstantConstant` / `VariableConstant` outlined above (the names are voluntarily tongue-in-cheek)?

yebai: Isn't `Constant` and `ConstantConstant` better?

gdalle: No, if we actually introduce this distinction (which I'm not yet 100% sure is a good idea) I'll have to pick different names, like `FixedParameter` and `VariableParameter` (and set `Constant = VariableParameter` until the next breaking release)

yebai:
I suggest that we close this PR for now and revisit it later: 

- `LogDensityProblemsAD` already uses `DI` under the hood, whereas `DI` is better. 
- `LogDensityProblemsAD` contains further optimisations that are currently missing from `DI`.

Issue #330, opened 2023-07-25
Moving ReadMe to Docs


JaimeRZP:
The GitHub ReadMe file is getting a little bit to big. 
Eventually we will want to migrate most of the information there to the docs.

PR #414, opened 2025-03-31
Better documentation

ErikQQY:
Fix: #330 
Fix: #413

PR #417, opened 2025-04-01
Better documentation

ErikQQY:
Fix: #330
Fix: #413
Fix: #402
Replace #414 

Preview can be checked in https://turinglang.org/AdvancedHMC.jl/previews/PR417


yebai:
Thanks @ErikQQY. I suggest that we 

- move `Interface` `news` and `API changes` to doc too
- keep a worked example in the README file for illustration purposes and point the audience to relevant docs for more advanced use

docs/src/references.md
```diff
@@ -0,0 +1,4 @@
+# References
```
mhauru, line 1: I'm unable to leave a comment in the right location, so I'll just dump it here: In index.md, would be good to spell out HMC as Hamiltonian Monte Carlo the first time it's mentioned.

mhauru, line 1: Likewise in index.md, the sentence "An illustrative example of AdvancedHMC's usage is given below." is not up-to-date, there's no example below.

docs/src/get_started.md
```diff
@@ -0,0 +1,202 @@
+# Sampling from a multivariate Gaussian using NUTS
+
+In this get-start section, we demonstrate a minimal example of sampling from a multivariate Gaussian (10-dimensional) using the no U-turn sampler (NUTS). Below we describe the major components of the Hamiltonian system which are essential to sample using this approach:
```
mhauru, line 3:
```suggestion
In this section, we demonstrate a minimal example of sampling from a multivariate Gaussian (10-dimensional) using the No U-Turn Sampler (NUTS). Below we describe the major components of the Hamiltonian system which are essential to sample using this approach:
```
If you do want to keep the word, it should probably be "get-started", but I think it's unnecessary.

docs/src/get_started.md
```diff
@@ -0,0 +1,202 @@
...
+
+  - **Leapfrog integration**: Leapfrog integration is a second-order numerical method for integrating differential equations (In this case they are equations of motion for the relative position of one particle with respect to the other). The order of this integration signifies its rate of convergence. Any algorithm with a finite time step size will have numerical errors, and the order is related to this error. For a second-order algorithm, this error scales as the second power of the time step, hence, the name second-order. High-order integrators are usually complex to code and have a limited region of convergence; hence they do not allow arbitrarily large time steps. A second-order integrator is suitable for our purpose. Hence we opt for the leapfrog integrator. It is called `leapfrog` due to the ways this algorithm is written, where the positions and velocities of particles "leap over" each other.[^2]
```
mhauru, line 7:
```suggestion
  - **Leapfrog integration**: Leapfrog integration is a second-order numerical method for integrating differential equations (In this case they are equations of motion for the relative position of one particle with respect to the other). The order of this integration signifies its rate of convergence. Any algorithm with a finite time step size will have numerical errors, and the order is related to this error. For a second-order algorithm, this error scales as the second power of the time step, hence the name. High-order integrators are usually complex to code and have a limited region of convergence; thus they do not allow arbitrarily large time steps. A second-order integrator is suitable for our purpose. Hence we opt for the leapfrog integrator. It is called `leapfrog` due to the ways this algorithm is written, where the positions and velocities of particles "leap over" each other.[^2]
```
Replaced one "hence" with "thus" just because this line had "hence" quite a lot of times.

docs/src/autodiff.md
```diff
@@ -0,0 +1,5 @@
+# Gradient in AdvancedHMC.jl
+
+AdvancedHMC.jl supports automatic differentiation using [`LogDensityProblemsAD`](https://github.com/tpapp/LogDensityProblemsAD.jl) and user-specified gradients. While the default AD backend for AdvancedHMC.jl is ForwardDiff.jl, we can seamlessly change to other backend like Zygote.jl using various syntax like `Hamiltonian(metric, ℓπ, Zygote)`, `Hamiltonian(metric, ℓπ, Val(:Zygote))` or via ADTypes.jl `Hamiltonian(metric, ℓπ, AutoZygote())`.
```
mhauru, line 3: I wonder if we should use Mooncake as an example rather than Zygote, because Zygote is dying and we are trying to promote Mooncake as the replacement.

mhauru, line 3: Might we also want to be opinionated in favouring the ADTypes approach, as the new standard way of specifying AD backends? I'm not suggesting leaving the other two out completely, but just putting ADTypes at the default solution and the other two as alternatives.

docs/src/autodiff.md
```diff
@@ -0,0 +1,5 @@
...
+
+In order to use user-specified gradients, please replace ForwardDiff.jl with `ℓπ_grad` in the `Hamiltonian` constructor, where the gradient function `ℓπ_grad` should return a tuple containing both the log-posterior and its gradient.
```
mhauru, line 5: I found this sentence a bit confusing, it took me a while to realise that this refers to passing the `ForwardDiff` module as an argument to the constructor. Maybe an example could help?

docs/src/api.md
```diff
@@ -1,26 +1,80 @@
...
+  - The adaptor is set by `adaptor`, for which the default is no adaptation.
+    
+      + It will perform `n_adapts` steps of adaptation, for which the default is `1_000` or 10% of `n_samples`, whichever is lower.
+  - `drop_warmup` specifies whether to drop samples.
+  - `verbose` controls the verbosity.
+  - `progress` controls whether to show the progress meter or not.
+
+Note that the function signature of the `sample` function exported by `AdvancedHMC.jl` differs from the [`sample`](https://turinglang.org/dev/docs/using-turing/guide#modelling-syntax-explained) function used by `Turing.jl`. We refer to the documentation of `Turing.jl` for more details on the latter.
+
+Note that the function signature of the `sample` function exported by `AdvancedHMC.jl` differs from the [`sample`](https://turinglang.org/dev/docs/using-turing/guide#modelling-syntax-explained) function used by `Turing.jl`. We refer to the documentation of `Turing.jl` for more details on the latter.
```
mhauru, line 76:
```suggestion
```
Repeat of the above line.

docs/src/api.md
```diff
@@ -1,26 +1,80 @@
-# AdvancedHMC.jl
+# Detailed API for AdvancedHMC.jl
 
-Documentation for AdvancedHMC.jl
+An important design goal of AdvancedHMC.jl is modularity; we would like to support algorithmic research on HMC.
+This modularity means that different HMC variants can be easily constructed by composing various components, such as preconditioning metric (i.e., mass matrix), leapfrog integrators, trajectories (static or dynamic), adaption schemes, etc. In this documentation, we will explain the detailed usage of different modules in AdancedHMC.jl to provide a comprehensive udnerstanding of how AdvancedHMC.jl can achieve both modularity and efficiency.
 
-```@contents
-```
+### [Hamiltonian mass matrix (`metric`)](@id hamiltonian_mm)
```
mhauru, line 6: Could we add a sentence before this section saying that we first list some of the exported types/features by categories, but that full docs can be found below? It took me a while to realise that the full ```@docs``` entries of all these are below. Alternatively, we could put in ```@docs``` entries here for all the ones we want to highlight category-by-category, and leave the rest for the ```@autodocs``` block. That way there would also be less to maintain (e.g. if the constructor arguments for one of these change).

docs/src/api.md
```diff
@@ -1,26 +1,80 @@
...
+      + It will perform `n_adapts` steps of adaptation, for which the default is `1_000` or 10% of `n_samples`, whichever is lower.
+  - `drop_warmup` specifies whether to drop samples.
+  - `verbose` controls the verbosity.
+  - `progress` controls whether to show the progress meter or not.
+
+Note that the function signature of the `sample` function exported by `AdvancedHMC.jl` differs from the [`sample`](https://turinglang.org/dev/docs/using-turing/guide#modelling-syntax-explained) function used by `Turing.jl`. We refer to the documentation of `Turing.jl` for more details on the latter.
+
+Note that the function signature of the `sample` function exported by `AdvancedHMC.jl` differs from the [`sample`](https://turinglang.org/dev/docs/using-turing/guide#modelling-syntax-explained) function used by `Turing.jl`. We refer to the documentation of `Turing.jl` for more details on the latter.
+
 ## More types
```
mhauru, line 78: If all docstrings are in this section then the name should probably be changed. See the above comment though for a proposal to move some of the docstrings above, in which case the name would remain appropriate.

docs/src/refs.bib
```diff
@@ -0,0 +1,48 @@
+@article{neal2011mcmc,
+  title={Mcmc using hamiltonian dynamics (handbook of markov chain monte carlo) ed s brooks et al},
```
mhauru, line 2:
```suggestion
  title={MCMC using Hamiltonian dynamics (Handbook of Markov Chain Monte Carlo) ed s Brooks et al},
```
Also the `ed s Brooks et al` should probably be formatted differently somehow, not sure what the best way is.

mhauru (changes requested):
Looks good. I know this is mostly moving stuff around, but I took the opportunity to read through the docs as well, so I had a few small requests.

docs/src/autodiff.md
ErikQQY: Yeah, actually, I think the ADTypes approach is better since that would allow users to have more control in the AD backend. For example, we can specify `config` in the Mooncake backend, but the other two ways can not do such things.

sunxd3 (commented):
Some nitpicking.

Also, the reference list at https://turinglang.org/AdvancedHMC.jl/previews/PR417/#References and https://turinglang.org/AdvancedHMC.jl/previews/PR417/references/ are not identical. Not sure if we want to unify them.

sunxd3:
extra line at https://github.com/TuringLang/AdvancedHMC.jl/blob/e23c7a06af9a0188e71327413b237a954245e3e0/docs/src/index.md?plain=1#L58, otherwise good from me

yebai:
Thanks @ErikQQY!

docs/src/autodiff.md
```diff
@@ -1,5 +1,5 @@
 # Gradient in AdvancedHMC.jl
 
-AdvancedHMC.jl supports automatic differentiation using [`LogDensityProblemsAD`](https://github.com/tpapp/LogDensityProblemsAD.jl) and user-specified gradients. While the default AD backend for AdvancedHMC.jl is ForwardDiff.jl, we can seamlessly change to other backend like Zygote.jl using various syntax like `Hamiltonian(metric, ℓπ, Zygote)`, `Hamiltonian(metric, ℓπ, Val(:Zygote))` or via ADTypes.jl `Hamiltonian(metric, ℓπ, AutoZygote())`.
+AdvancedHMC.jl supports automatic differentiation using [`LogDensityProblemsAD`](https://github.com/tpapp/LogDensityProblemsAD.jl) across various AD backends and allows user-specified gradients. While the default AD backend for AdvancedHMC.jl is ForwardDiff.jl, we can seamlessly change to other backend like Mooncake.jl using various syntax like `Hamiltonian(metric, ℓπ, AutoZygote())`. Different AD backend can also be pluged in using `Hamiltonian(metric, ℓπ, Zygote)`, `Hamiltonian(metric, ℓπ, Val(:Zygote))` but we recommend using ADTypes since that would allow you to have more freedom for specifying the AD backend.
```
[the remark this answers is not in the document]
yebai: I probably message this too soon. Please open another follow-up PR.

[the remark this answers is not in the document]
ErikQQY: Follow-up in #423

PR #426, opened 2025-04-13
Fix gradient length mismatch in PhasePoint

ErikQQY:
Part of #344


ErikQQY:
Hmmm, seems there might be some code relying on this issue, need further investigation

yebai:
Good work, @ErikQQY!

PR #427, opened 2025-04-13
Reduce allocation in mass matrix adaptor

ErikQQY:
Instead of allocating using `zeros` and `ones` when resizing the adaptor, we can directly use in-place operations like `resize!` and `fill!` to reduce allocations.

### DIFF
--- a/src/adaptation/massmatrix.jl
+++ b/src/adaptation/massmatrix.jl
@@ -13,7 +13,7 @@ function adapt!(
     α::AbstractScalarOrVec{<:AbstractFloat},
     is_update::Bool=true,
 )
-    resize!(adaptor, θ)
+    resize_with!(adaptor, θ)
     push!(adaptor, θ)
     is_update && update!(adaptor)
     return nothing
@@ -29,7 +29,7 @@ UnitMassMatrix() = UnitMassMatrix{Float64}()
 
 Base.string(::UnitMassMatrix) = "I"
 
-Base.resize!(pc::UnitMassMatrix, θ::AbstractVecOrMat) = nothing
+resize_with!(pc::UnitMassMatrix, θ::AbstractVecOrMat) = nothing
 
 reset!(::UnitMassMatrix) = nothing
 
@@ -102,7 +102,7 @@ function WelfordVar(sz::Union{Tuple{Int},Tuple{Int,Int}}; kwargs...)
     return WelfordVar{Float64}(sz; kwargs...)
 end
 
-function Base.resize!(wv::WelfordVar, θ::AbstractMatrix{T}) where {T<:AbstractFloat}
+function resize_with!(wv::WelfordVar, θ::AbstractMatrix{T}) where {T<:AbstractFloat}
     size_θ = size(θ)
     if size_θ != size(wv.var)
         @assert wv.n == 0 "Cannot resize a var estimator when it contains samples."
@@ -113,33 +113,15 @@ function Base.resize!(wv::WelfordVar, θ::AbstractMatrix{T}) where {T<:AbstractF
     end
 end
 
-function Base.resize!(wv::WelfordVar, θ::AbstractVector{T}) where {T<:AbstractFloat}
-    n = length(θ)
-    if n != size(wv.var, 1)
+function resize_with!(wv::WelfordVar, θ::AbstractVector{T}) where {T<:AbstractFloat}
+    length_θ = length(θ)
+    if length_θ != size(wv.var, 1)
         @assert wv.n == 0 "Cannot resize a var estimator when it contains samples."
-        __resize_fill!(wv.μ, n, T(0))
-        __resize_fill!(wv.M, n, T(0))
-        __resize_fill!(wv.δ, n, T(0))
-        __resize_fill!(wv.var, n, T(1))
+        fill!(resize!(wv.μ, length_θ), T(0))
+        fill!(resize!(wv.M, length_θ), T(0))
+        fill!(resize!(wv.δ, length_θ), T(0))
+        fill!(resize!(wv.var, length_θ), T(1))
     end
-end
-
-"""
-    __resize_fill!(x::AbstractVector, new_length::Int, v::T)
-
-Resize `x` according to `new_length`:
- - If `new_length` is bigger than the current length of `x`, expand `x` to `new_length` and fill `x` with `v`.
- - If `new_length` is smaller than the current length of `x`, resize `x` to `new_length` and fill `x` with `v`.
-"""
-function __resize_fill!(x::AbstractVector, new_length::Int, v::T) where {T<:Number}
-    length_diff = new_length - length(x)
-    if length_diff > 0
-        append!(x, Vector{T}(undef, length_diff))
-    else
-        resize!(x, new_length)
-    end
-    fill!(x, v)
-    return nothing
 end
 
 function reset!(wv::WelfordVar{T}) where {T<:AbstractFloat}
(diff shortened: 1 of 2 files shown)


src/adaptation/massmatrix.jl
```diff
@@ -112,10 +112,32 @@ function Base.resize!(wv::WelfordVar, θ::AbstractVecOrMat{T}) where {T<:Abstrac
...
+        @assert wv.n == 0 "Cannot resize a var estimator when it contains samples."
+        n = length(θ)
+        __fill!(wv.μ, n, T(0))
+        __fill!(wv.M, n, T(0))
+        __fill!(wv.δ, n, T(0))
+        __fill!(wv.var, n, T(1))
+    end
+end
+
+function __fill!(x::AbstractVector, n, V)
```
yebai, line 126: For clarity: 

```suggestion
function __resize_fill!(x::AbstractVector, n, V)
```

Also, please add a docstring to this function.

src/adaptation/massmatrix.jl
```diff
@@ -102,20 +102,50 @@ function WelfordVar(sz::Union{Tuple{Int},Tuple{Int,Int}}; kwargs...)
     return WelfordVar{Float64}(sz; kwargs...)
 end
 
-function Base.resize!(wv::WelfordVar, θ::AbstractVecOrMat{T}) where {T<:AbstractFloat}
-    if size(θ) != size(wv.var)
+function Base.resize!(wv::WelfordVar, θ::AbstractMatrix{T}) where {T<:AbstractFloat}
```
devmotion, line 105: IMO this function must be renamed, it's a misuse of `Base.resize!` as the second argument of `resize!` is supposed to be an integer. It should just be an internal function.

src/adaptation/massmatrix.jl
```diff
@@ -102,20 +102,50 @@ function WelfordVar(sz::Union{Tuple{Int},Tuple{Int,Int}}; kwargs...)
...
+        wv.δ = zeros(T, size_θ)
+        wv.var = ones(T, size_θ)
     end
 end
 
+function Base.resize!(wv::WelfordVar, θ::AbstractVector{T}) where {T<:AbstractFloat}
+    n = length(θ)
+    if n != size(wv.var, 1)
+        @assert wv.n == 0 "Cannot resize a var estimator when it contains samples."
+        __resize_fill!(wv.μ, n, T(0))
```
devmotion, line 120: I don't really see the need for a custom function, it seems this is just
```suggestion
        fill!(resize!(wv.μ, n), T(0))
```

src/adaptation/massmatrix.jl
```diff
@@ -102,20 +102,50 @@ function WelfordVar(sz::Union{Tuple{Int},Tuple{Int,Int}}; kwargs...)
...
+ - If `new_length` is bigger than the current length of `x`, expand `x` to `new_length` and fill `x` with `v`.
+ - If `new_length` is smaller than the current length of `x`, resize `x` to `new_length` and fill `x` with `v`.
+"""
+function __resize_fill!(x::AbstractVector, new_length::Int, v::T) where {T<:Number}
+    length_diff = new_length - length(x)
+    if length_diff > 0
+        append!(x, Vector{T}(undef, length_diff))
+    else
+        resize!(x, new_length)
+    end
```
devmotion, line 140: Both branches seem to be just
```suggestion
    resize!(x, new_length)
```
? As mentioned above, in that case `fill!(resize!(..), ..)` would be simpler than a separate function IMO.

src/adaptation/massmatrix.jl
ErikQQY: Yeah, `resize!` here is not appropriate for the context

src/adaptation/massmatrix.jl
ErikQQY: Good idea! Just changed to simpler resize and fill operations

src/adaptation/massmatrix.jl
```diff
@@ -102,20 +102,32 @@ function WelfordVar(sz::Union{Tuple{Int},Tuple{Int,Int}}; kwargs...)
     return WelfordVar{Float64}(sz; kwargs...)
 end
 
-function Base.resize!(wv::WelfordVar, θ::AbstractVecOrMat{T}) where {T<:AbstractFloat}
-    if size(θ) != size(wv.var)
+function resize_with!(wv::WelfordVar, θ::AbstractMatrix{T}) where {T<:AbstractFloat}
```
devmotion, line 105: `T` should be determined by `WelfordVar`, shouldn't it?
```suggestion
function resize_with!(wv::WelfordVar{T}, θ::AbstractMatrix{<:Real}) where {T<:AbstractFloat}
```
? Or possibly only matching inputs should be allowed?

src/adaptation/massmatrix.jl
```diff
@@ -102,20 +102,32 @@ function WelfordVar(sz::Union{Tuple{Int},Tuple{Int,Int}}; kwargs...)
...
-        wv.δ = zeros(T, size(θ))
-        wv.var = ones(T, size(θ))
+        wv.μ = zeros(T, size_θ)
+        wv.M = zeros(T, size_θ)
+        wv.δ = zeros(T, size_θ)
+        wv.var = ones(T, size_θ)
+    end
+end
+
+function resize_with!(wv::WelfordVar, θ::AbstractVector{T}) where {T<:AbstractFloat}
```
devmotion, line 116:
```suggestion
function resize_with!(wv::WelfordVar{T}, θ::AbstractVector{<:Real}) where {T<:AbstractFloat}
```

src/adaptation/massmatrix.jl
```diff
@@ -191,20 +203,21 @@ end
 
 WelfordCov(sz::Tuple{Int}; kwargs...) = WelfordCov{Float64}(sz; kwargs...)
 
-function Base.resize!(wc::WelfordCov, θ::AbstractVector{T}) where {T<:AbstractFloat}
-    if length(θ) != size(wc.cov, 1)
+function resize_with!(wc::WelfordCov, θ::AbstractVector{T}) where {T<:AbstractFloat}
```
devmotion, line 206:
```suggestion
function resize_with!(wc::WelfordCov{T}, θ::AbstractVector{<:Real}) where {T<:AbstractFloat}
```

src/adaptation/massmatrix.jl
yebai: @ErikQQY If one passes `length(θ)` (i.e. `resize!(wv::WelfordVar, length(θ)`), then we could keep the function name `Base.resize!`. I slightly prefer `Base.resize!` since it is more readable than `resize_with`, whose meaning is unclear from the function name alone.

ErikQQY: Commit https://github.com/TuringLang/AdvancedHMC.jl/pull/427/commits/b4e60f872eec1a09950ec42888d02297baec1004 extends `Base.resize!` to allow passing `resize!(wv::WelfordVar, size(θ))` so that we can have a unifying resizing and also reduce the allocations. Let me know if this way introduces any misinterpretation.

src/adaptation/massmatrix.jl
ErikQQY: Now `T` is determined by `WelfordVar`

src/adaptation/massmatrix.jl
```diff
@@ -29,7 +29,7 @@ UnitMassMatrix() = UnitMassMatrix{Float64}()
 
 Base.string(::UnitMassMatrix) = "I"
 
-Base.resize!(pc::UnitMassMatrix, θ::AbstractVecOrMat) = nothing
+Base.resize!(pc::UnitMassMatrix, size_θ::Tuple) = nothing
```
devmotion, line 32: This still should not be `Base.resize!` - to confirm with its API and expected behaviour the second argument must be an integer.

ErikQQY: The main issue here is the naming of the resizing of mass matrix adaptor, extending `Base.resize!` and allowing its second argument to be a non-integer is not a good idea, and according to @yebai, we should have a consistent naming, so to keep it simple and concise, I think maybe we can change to `resize_adaptor!(pc::WelfordVar, size::Tuple)`?

yebai: Sure, happy with `resize_adaptor!`.

yebai (approved):

yebai:
Thanks @ErikQQY and @devmotion!

Issue #429, opened 2025-04-23
GPU tests via buildkite


yebai:
This requires 

- Remove `ReTest` in tests #327 
- Re-enable CUDA [test](https://github.com/TuringLang/AdvancedHMC.jl/blob/567b9964c5b43ac81875db45d76414232392e932/test/cuda.jl#L26)

Buildkite setup [example](https://github.com/TuringLang/NormalizingFlows.jl/blob/main/.buildkite/pipeline.yml)

PR #430, opened 2025-04-24
Revive GPU tests

ErikQQY:
Fix: #429 

Instead of completely removing the current ReTest.jl test suite, this PR moves the GPU tests to another test environment so that we can only use Test.


ErikQQY:
It occurs to me that while we no longer throw warning messages for numerical errors, we can safely use ReTest.jl for GPU tests as well?

https://github.com/TuringLang/AdvancedHMC.jl/blob/7067a90051680c800c16a4e52ff17ca7d828db8e/test/hamiltonian.jl#L34-L36

yebai:
@sunxd3, can you review this PR and help set up buildkite for this repo?

sunxd3:
The changes looks sensible. 

I messaged Tim who manages JuliaGPU on slack and I'll update when he returns.

In the meantime, I don't know AHMC code in close detail. I remember `ReTest`'s purpose is to allow writing tests in source files. Are we using this pattern in AHMC? If not, are there other reasons we would still need `ReTest`?

ErikQQY:
ReTest.jl is introduced in https://github.com/TuringLang/AdvancedHMC.jl/pull/287 with the intention of allowing a subset of testing and providing a better summary of test results, see https://github.com/TuringLang/AdvancedHMC.jl/blob/main/test/README.md

sunxd3:
Got it, thanks.

I wonder if we can use [`Pkg.test`](https://pkgdocs.julialang.org/v1/api/#Pkg.test) directly now, it also takes a `test_args` argument. (I am not sure if this was not true back then, or ReTest simply has a better interface.)

ErikQQY:
> I wonder if we can use [Pkg.test](https://pkgdocs.julialang.org/v1/api/#Pkg.test) directly now, it also takes a test_args argument.

I may not quite understand here, we can directly call `Pkg.test()` to test the package. And while the CUDA tests are no longer affected by ReTest.jl, I don't think it really matters if we are using Test or ReTest.

sunxd3:
You are right -- this current PR is not discussing the use of ReTest 👍

yebai:
I'm merging this now since we need the Buildkite script to be on `main` to be activated

sunxd3:
The GPU pipeline should be set up

PR #431, opened 2025-04-26
Add quality tests

ErikQQY:
Helps remove potential bugs and type instabilities


src/integrator.jl
```diff
@@ -252,5 +252,6 @@ function step(
             break
         end
     end
-    return res
+    FullTraj && return res
+    return first(res)
```
penelopeysm, line 256:
```suggestion
    return if FullTraj
        res
    else
        first(res)
    end
```

src/AdvancedHMC.jl
```diff
@@ -172,9 +172,8 @@ function Hamiltonian(metric::AbstractMetric, ℓπ, kind::Union{Symbol,Val,Modul
             ),
         )
     end
-    ℓ = LogDensityProblemsAD.ADgradient(
-        kind isa Val ? kind : Val(Symbol(kind)), ℓπ; kwargs...
-    )
+    _kind = ifelse((kind isa Val) || (kind isa Symbol), kind, Symbol(kind))
```
penelopeysm, line 175:
```suggestion
    _kind = if kind isa Val || kind isa Symbol
        kind
    else
        Symbol(kind))
    end
```

penelopeysm (changes requested):
Generally, I prefer the explicit `if... else... end` expressions

src/integrator.jl
penelopeysm: The reason why I prefer this (even though it's longer) is because it's a more functional style of programming. The effect of `FullTraj && return res` isn't localised - in that if you move the line up a couple of lines it can drastically change the effect of the programme.

src/AdvancedHMC.jl
penelopeysm: This one is more stylistic, I think I just like having things spelled out rather than having one-liners

src/integrator.jl
ErikQQY: Yeah, just to the `if...else` style

src/AdvancedHMC.jl
ErikQQY: Indeed, we should prioritize readability.

penelopeysm (approved):
Thank you for adding these :)

Is the failure on 1.12 expected?

ErikQQY:
Yeah, the failures on 1.12 are expected

yebai:
Thanks @ErikQQY and @penelopeysm. Please feel free to merge once you have two independent approvals!

yebai (approved):

src/integrator.jl
```diff
@@ -218,7 +218,7 @@ function step(
     res = if FullTraj
         Vector{P}(undef, n_steps)
     else
-        z
+        Vector{P}(undef, 1)
```
devmotion, line 221: Why was this changed? This creates unnecessary allocations in every `step`. There's no type inference issue here since `FullTraj` is a type parameter of a `Val`, so the compiler compiles and optimizes separate methods for `FullTraj = true` and `FullTraj = false` where these branches are eliminated.

PR #432, opened 2025-04-27
Add Riemannian manifold HMC

ErikQQY:
Add the Riemannian manifold HMC from "Riemann manifold Langevin and Hamiltonian Monte  Carlo methods" which is implemented in the research directory but has not been merged into the package.

Though this PR is relatively complete, there are still some things TODO:

- Register the [MCMCLogDensityProblems.jl](https://github.com/chalk-lab/MCMCLogDensityProblems.jl) package since some of its functionalities are used in RMHMC.
- Discuss the appropriate style for constructing a `DenseRiemannianMetric`, e.g., how should the metric tensor `G` be provided, etc.
- Still need to add some docstrings and related documentation.


yebai:
> Register the [MCMCLogDensityProblems.jl](https://github.com/chalk-lab/MCMCLogDensityProblems.jl) package since some of its functionalities are used in RMHMC.

Let's avoid a hard dependency on MCMCLogDensityProblems from AHMC.

ErikQQY:
Now MCMCLogDensityProblems is only used for tests, but without the generative hessian utilities from MCMCLogDensityProblems, users might have to build a Riemannian manifold metric with `G` and `∂G∂θ` computed from themselves.

yebai:
@ErikQQY, it seems the new files in this PR are copied from the `research` folder. 

Can you try to move or rename these files instead of copying them, so that we can preserve the Git history of these codes?

yebai:
@ErikQQY, it might be easier to open a new PR from scratch where 

- First, move the code, 
- Then add back your changes

This makes it easier for Git to recognise and track the changes.

ErikQQY:
Close in favor of #437

PR #433, opened 2025-04-28
Fix CUDA tests

ErikQQY:
With the GPU pipeline up in https://github.com/TuringLang/AdvancedHMC.jl/pull/430, we should be able to test GPU compatibility of AdvancedHMC.jl


yebai (approved):
Thanks @ErikQQY!

Issue #248, opened 2021-02-21
Unify argument orders


xukai92:
Currently the order of internal functions on the hamiltonian, phasepoint, kernel/trajectory are not all consistent.
We should make all of them having a order of `(rng, hamiltonian, kernel, z)`.

Also see https://github.com/TuringLang/AdvancedHMC.jl/pull/247/files#r579401575

PR #435, opened 2025-04-30
Unify argument order in phasepoint and transition

ErikQQY:
Fix: #248 

Now we should only use `transition(rng, h, τ, z)` or `transition(rng, h, κ, z)`.

### DIFF
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -167,6 +167,18 @@ end
 
 energy(args...) = -neg_energy(args...)
 
+####
+#### Momentum refreshment
+####
+
+function phasepoint(
+    rng::Union{AbstractRNG,AbstractVector{<:AbstractRNG}},
+    θ::AbstractVecOrMat{T},
+    h::Hamiltonian,
+) where {T<:Real}
+    return phasepoint(h, θ, rand_momentum(rng, h.metric, h.kinetic, θ))
+end
+
 abstract type AbstractMomentumRefreshment end
 
 "Completly resample new momentum."
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -167,18 +167,6 @@ end
 
 energy(args...) = -neg_energy(args...)
 
-####
-#### Momentum refreshment
-####
-
-function phasepoint(
-    rng::Union{AbstractRNG,AbstractVector{<:AbstractRNG}},
-    θ::AbstractVecOrMat{T},
-    h::Hamiltonian,
-) where {T<:Real}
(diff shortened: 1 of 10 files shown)


src/sampler.jl
```diff
@@ -40,8 +40,9 @@ function sample_init(
 )
     # Ensure h.metric has the same dim as θ.
     h = resize(h, θ)
+    refresh_r = rand_momentum(rng, h.metric, h.kinetic, θ) # Momentum refreshment
```
yebai, line 43:
```suggestion
    refreshed_r = rand_momentum(rng, h.metric, h.kinetic, θ) # Momentum refreshment
```

src/sampler.jl
```diff
@@ -40,8 +40,9 @@ function sample_init(
...
     # Initial transition
-    t = Transition(phasepoint(rng, θ, h), NamedTuple())
+    t = Transition(phasepoint(h, θ, refresh_r), NamedTuple())
```
yebai, line 45:
```suggestion
    t = Transition(phasepoint(h, θ, refreshed_r), NamedTuple())
```

and similarly, all other places.

yebai (changes requested):

yebai:
It is more prudent to mark this as a breaking release, given that some other packages likely depend on `AHMC.transition`, even though it is not exported.

src/hamiltonian.jl
```diff
@@ -167,18 +167,6 @@ end
 
 energy(args...) = -neg_energy(args...)
 
-####
-#### Momentum refreshment
-####
-
-function phasepoint(
```
yebai, line 174: On second thought, let's revert the change to the `phasepoint` function, since the current function is meant to support vectorised HMC. In comparison, the proposed changes in this PR are a minor regression.

ErikQQY: Oh yeah, that would lose the vectorized HMC support, just reverted the change.

ErikQQY:
> It is more prudent to mark this as a breaking release, given that some other packages likely depend on AHMC.transition, even though it is not exported.

Yeah, I bumped the version to 0.8.0, so we can release a breaking change. As far as I know, Turing.jl is using `AHMC.transition`, I can fix its usage after the release.

mhauru (approved):
Looks all good to me. Thanks for this, I think I've tripped over that argument ordering once or twice myself.

yebai (dismissed):

yebai:
Excellent work - @ErikQQY, can you add an item to HISTORY.md before merging?

yebai (approved):

PR #437, opened 2025-05-04
Add Riemannian manifold HMC

ErikQQY:
Add the Riemannian manifold HMC from "Riemann manifold Langevin and Hamiltonian Monte Carlo methods" which is implemented in the research directory but has not been merged into the package. Supersede #432 

Though this PR is relatively complete, there are still some things TODO:

- Discuss the appropriate style for constructing a DenseRiemannianMetric, e.g., how should the metric tensor G be provided, etc.

- Still need to add some docstrings and related documentation.

PR #439, opened 2025-05-05
Add Riemannian manifold HMC

ErikQQY:
Add the Riemannian manifold HMC from "Riemann manifold Langevin and Hamiltonian Monte Carlo methods" which is implemented in the research directory but has not been merged into the package. Supersede #432 #437


src/riemannian/metric.jl
```diff
@@ -0,0 +1,63 @@
...
+end
+
+function rand_momentum(
+    rng::Union{AbstractRNG,AbstractVector{<:AbstractRNG}},
+    metric::DenseRiemannianMetric{T},
+    kinetic,
+    θ::AbstractVecOrMat,
+) where {T}
+    r = _randn(rng, T, size(metric)...)
+    G⁻¹ = inv(metric.map(metric.G(θ)))
```
THargreaves, line 59: It's very inefficient and numerically unstable to compute a direct inverse. Worse, you're not taking advantage of the fact that G is PSD. Instead, you should compute the Cholesky decomposition first.

Let G = U'U where U are the upper triangles from the Cholesky decomposition.

You're computing the Cholesky of G⁻¹ which is (U⁻¹)(U⁻¹)'.

You then do a ldiv to compute r = (U'⁻¹)⁻¹ * r, but that's just U' * r.

Instead, you can just do `lmul!(chol.L, r)` where `chol` is the Cholesky of G directly.

src/riemannian/hamiltonian.jl
```diff
@@ -1,257 +1,16 @@
...
-    r::T;
-    ℓπ=∂H∂θ(h, θ),
-    ℓκ=DualValue(neg_energy(h, r, θ), ∂H∂r(h, θ, r)),
-) where {T<:AbstractVecOrMat}
-    return PhasePoint(θ, r, ℓπ, ℓκ)
-end
-
-# Negative kinetic energy
-#! Eq (13) of Girolami & Calderhead (2011)
-function neg_energy(
```
THargreaves, line 241: Is this correct? Eq (13) of Girolami & Calderhead (2011) also contains the log-likelihood term, -L(θ). Why is this not included here?

Even if this is correct, we should clarify the naming conventions as it's quite hard to follow.

src/riemannian/hamiltonian.jl
```diff
@@ -1,257 +1,16 @@
...
-Base.size(e::DenseRiemannianMetric, dim::Int) = e.size[dim]
-Base.show(io::IO, dem::DenseRiemannianMetric) = print(io, "DenseRiemannianMetric(...)")
-
-function rand_momentum(
-    rng::Union{AbstractRNG,AbstractVector{<:AbstractRNG}},
-    metric::DenseRiemannianMetric{T},
-    kinetic,
+#! Eq (14) of Girolami & Calderhead (2011)
+function ∂H∂r(
+    h::Hamiltonian{<:DenseRiemannianMetric,<:GaussianKinetic},
```
THargreaves, line 3: `AbstractRiemannianMetric`?

This logic is not unique to the dense metric.

nsiccha:
I've had a quick look - this PR clearly still needs some work. The sampling of the random momentum seems to be numerically unstable and inefficient, see (this PR) https://github.com/TuringLang/AdvancedHMC.jl/blob/c9e6b0af7dfc273f6310a403869698c2510d2078/src/riemannian/metric.jl#L52-L63 and maybe also compare with ([that](https://github.com/TuringLang/AdvancedHMC.jl/pull/481) PR) https://github.com/TuringLang/AdvancedHMC.jl/blob/9cac513e683412f63a5148ee7de968271cef28ac/src/riemannian/metric.jl#L63-L73.

For the Softabs map, which internally does an eigendecomposition (see https://github.com/TuringLang/AdvancedHMC.jl/blob/c9e6b0af7dfc273f6310a403869698c2510d2078/src/riemannian/metric.jl#L13-L19), not even the cholesky decomposition will be required if we hold onto the eigendecomposition.

THargreaves:
> not even the cholesky decomposition will be required if we hold onto the eigendecomposition.

This is a good point. Though for code modularity it would be nice if we didn't have to redefine H and its derivatives specifically for the eigendecomposition case.

I have two thoughts on this:

1. We could write H and its derivatives in a generic way using solves — the only issue is that `E = eigen(A)` isn't treated as a proper factorisation in Julia so things like `E \ x` don't work.
2. You can compute the Cholesky of G from the decomposition by taking the QR decomposition of the eigenvector matrix. I did a back of the envelope FLOP calculation and this is ends up being slower if you only using it to perform vector solves and is a bit faster (but both O(D^3)) when performing matrix solves.

nsiccha:
> We could write H and its derivatives in a generic way using solves — the only issue is that E = eigen(A) isn't treated as a proper factorisation in Julia so things like E \ x don't work.

Yeah, we'd either have to find a wrapper for this somewhere in the wild (I'm sure it exists somewhere) or write our own.

> You can compute the Cholesky of G from the decomposition by taking the QR decomposition of the eigenvector matrix.

What would you compute the cholesky factor for? You can just compute the matrix's square root using the eigendecompoisition, which you already have, which should be all that you need.

THargreaves:
> What would you compute the cholesky factor for? You can just compute the matrix's square root using the eigendecompoisition, which you already have, which should be all that you need.

Hypothetically, if you could convert eigen to Cholesky for free, it would be more efficient to compute H and its derivatives using the low triangular L than the dense V from the eigendecomposition. But FLOPS-wise the overhead of doing the conversion only ends up being worth it if you are doing O(D) vector multiplies/solves.

This doesn't apply to us in the dHdp. It sort of does for the trace term in dHdθ though.

nsiccha:
> This doesn't apply to us in the dHdp. It sort of does for the trace term in dHdθ though.

The one in here?

https://github.com/TuringLang/AdvancedHMC.jl/blob/c9e6b0af7dfc273f6310a403869698c2510d2078/src/riemannian/hamiltonian.jl#L12-L34

(Edit for in place display of code)

nsiccha:
The trace of the product of two (square) matrices can be computed in O(n^2) without computing the O(n^3) matrix product, see e.g. https://en.wikipedia.org/wiki/Trace_(linear_algebra)#Trace_of_a_product.

Furthermore, do we think that Julia is clever enough to reorder this expression: `r' * invG * ∂G∂θᵢ * invG * r`? It's in any case unncessary to compute it as it's implemented there. Do the `tmp = invG * r` once, then compute the `tmp' * ∂G∂θᵢ * tmp` in one go - I'm sure there's a function for that.

THargreaves (changes requested):

THargreaves:
I made this review weeks ago but not sure it was actually showing up for others until I marked it as "complete".

nsiccha:
> I made this review weeks ago but not sure it was actually showing up for others until I marked it as "complete".

No haha, I hadn't seen this before :sweat_smile:

nsiccha:
I've got another, maybe silly, conceptual question.

If we're using the softabs map, then the Riemannian metric (right?) G is not the Fisher Information Metric, and it's also not the Hessian (which after all is not guaranteed to always be SPD and thus no metric), but `G(theta) = softabs(Hessian(theta))`.

Shouldn't the `G` field in the `DenseRiemannianMetric` type then already include the map? And the `∂G∂θ` should then compute the derivatives wrt to `softabs(Hessian(theta))`?

Edit: maybe all that modulo some inverses? I'm really not familiar with this algorithm.

THargreaves:
> Shouldn't the G field in the DenseRiemannianMetric type then already include the map? And the ∂G∂θ should then compute the derivatives wrt to softabs(Hessian(theta))?

That would make sense to me. All computations in the RHMC integrator are wrt G = softabs(H). H is never used directly.

This is the point I was alluding to when talking about separating the logic for the DenseRiemannanMetric in particular (softabs/eigen stuff) from the AbstractRiemannianMetric stuff.

zuhengxu:
Sorry guys, I'm finally getting started to review this! After a quick skim (on this particular PR), my first impression was:
- this is for dense riemanian metric 
- and the subsequent PRs are also for dense metric (I haven't examined them thoroughly so I could be wrong)
Is Diagonal approximation not in the roadmap so far?

nsiccha:
> Sorry guys, I'm finally getting started to review this! After a quick skim (on this particular PR), my first impression was:
> 
> * this is for dense riemanian metric
> * and the subsequent PRs are also for dense metric (I haven't examined them thoroughly so I could be wrong)

AFAIK, that's correct!

src/riemannian/hamiltonian.jl
nsiccha: I agree - created an issue here: https://github.com/TuringLang/AdvancedHMC.jl/issues/483

src/riemannian/metric.jl
nsiccha: Haven't checked the math, but it's probably correct. 

(Just adding this comment because I've accidentally started a review)

nsiccha:
This can be closed in favour of #485, right?

Issue #219, opened 2020-10-12
Vectorised sampling documentation


treigerm:
The link in the README pointing towards more information about the vectorized implementation about static HMC points to a webpage which errors with 404:
```
It also supports vectorized sampling for static HMC and has been discussed in more detail in the documentation [here](https://turing.ml/dev/docs/library/advancedhmc/parallel_sampling).
```

I was looking for the following information: I think I remember reading somewhere that the implementation for static HMC is using coupled random numbers. However, I cannot find any information about that anymore so I'm not sure whether I am 
just misremembering.

yebai:
Let's add an example utilising vectorised HMC to the docs

PR #440, opened 2025-05-07
Add vectorized HMC docs

ErikQQY:
Fix: #219


mhauru:
I think this would benefit from a high-level explanation of what vectorised sampling is, and how it differs from parallel sampling as explained elsewhere in the docs. Also things like when/why is it better, or not possible to do, etc.

Issue #277, opened 2021-09-11
Add low-rank metric adaptation


sethaxen:
The following paper presented a novel approach for adapting the metric:

Ben Bales, Arya Pourzanjani, Aki Vehtari, Linda Petzold, Selecting the Metric in Hamiltonian Monte Carlo, 2019. https://arxiv.org/abs/1905.11916

The idea is to adapt a low-rank approximation to the covariance matrix using a diagonal estimate and a low-rank approximation of the Hessian matrix. The main advantage of the estimated covariance being low-rank is that it can be stably estimated with many fewer draws than the dense covariance matrix. The paper also included a selection criterion for determining which rank to use.

They demonstrated in several benchmarks that the method often (but not always) outperformed diagonal and dense metric adaptation in terms of ESS/s.

sethaxen:
See also #70

sethaxen:
> As noted by @nsiccha on Slack, the reference code for the paper is available at [stan-dev/stan@develop...bbbales2:stan:experimental/warmup-wishart](https://github.com/stan-dev/stan/compare/develop...bbbales2:stan:experimental/warmup-wishart).

Specifically, these files seem to contain all code for the metric adaptation, while the rest is mostly glue code so it can be used with NUTS:
- https://github.com/bbbales2/stan/blob/25c276adf74956f57fbcb2e5fe526ce20d2cb1a5/src/stan/mcmc/var_experimental_adaptation.hpp
- https://github.com/bbbales2/stan/blob/25c276adf74956f57fbcb2e5fe526ce20d2cb1a5/src/stan/mcmc/covar_experimental_adaptation.hpp

nsiccha:
Yeah, after having a really short look, I also arrived at those two files. Haven't had the opportunity to port i anything yet though.

yebai:
Fixed by https://github.com/TuringLang/AdvancedHMC.jl/pull/511; thanks to @devmotion and @sethaxen.

PR #443, opened 2025-05-09
Add RankUpdateEuclideanMetric

### LINKED ISSUE #411: Making rand_momentum part of the API
As instructed by the v0.7.0 release notes, I'm opening this issue to propose making `rand_momentum` part of the API.

Pathfinder.jl provides an approximation of a target distribution as a multivariate normal whose covariance matrix has a particular structure that can be exploited for efficient sampling. Pathfinder implements `RankUpdateEuclideanMetric` [here](https://github.com/mlcolab/Pathfinder.jl/blob/44355ea64364404d0183fc7395849dafee195823/src/integration/advancedhmc.jl), which allows this covariance matrix to be used as an (inverse) metric in AdvancedHMC (for usage examples, see https://mlcolab.github.io/Pathfinder.jl/v0.9/examples/initializing-hmc/#AdvancedHMC.jl)

The recent replacement of `rand` methods with internal `rand_momentum` methods require Pathfinder to implement this internal method to support integration with AdvancedHMC v0.7.0. It would be useful if `rand_momentum` was added to the public API to maker overloading it safer.

Alternatives:
- The proposal in #34 to implement a single `Metric` object wrapping an `AbstractPDMat` would also work, since Pathfinder's covariance matrix type is also an `AbstractPDMat`.
- `RankUpdateEuclideanMetric` could live here in an extension instead of Pathfinder. However, since Pathfinder is more specific than this package, I think that makes less sense.
- One could blend the two alternatives above by adding a `EuclideanMetric` struct that wraps an `AbstractPDMat` and allows any PDMat to be wrapped as a metric that implements the necessary interface. It would be good if this could support wrapping either the inverse metric or the metric itself. For the former, doing this efficiently and generically requires some package (maybe PDMats) implements something like a `sqrt_fact` function that returns a factor `R` such that `R' * R` is the PDMat and `R` implement `ldiv!`. This generalizes the Cholesky factorization for structured matrices for which a different square root factorization may be more efficient to compute and use than a Cholesky factorization.

ErikQQY:
Fix: #411
Part of #277

This PR ports `RankUpdateEuclideanMetric` from Pathfinder.jl, but don't depend on PDMats.jl, see why not use PDMats.jl in #34.

With this metric in, we can take a step forward low-rank metric adaption, and Pathfinder.jl can directly use this metric from AHMC without unsafe overloading.

### DIFF
--- a/src/hamiltonian.jl
+++ b/src/hamiltonian.jl
@@ -61,6 +61,18 @@ function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::A
     return M⁻¹ * r
 end
 
+function ∂H∂r(
+    h::Hamiltonian{<:RankUpdateEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat
+)
+    (; M⁻¹) = h.metric
+    axes_M⁻¹ = __axes(M⁻¹)
+    axes_r = __axes(r)
+    (first(axes_M⁻¹) !== first(axes_r)) && throw(
+        ArgumentError("AxesMismatch: M⁻¹ has axes $(axes_M⁻¹) but r has axes $(axes_r)")
+    )
+    return M⁻¹ * r
+end
+
 # TODO (kai) make the order of θ and r consistent with neg_energy
 # TODO (kai) add stricter types to block hamiltonian.jl#L37 from working on unknown metric/kinetic
 # The gradient of a position-dependent Hamiltonian system depends on both θ and r. 
@@ -163,6 +175,13 @@ function neg_energy(
 ) where {T<:AbstractVecOrMat}
     mul!(h.metric._temp, h.metric.M⁻¹, r)
     return -dot(r, h.metric._temp) / 2
+end
+
+function neg_energy(
+    h::Hamiltonian{<:RankUpdateEuclideanMetric,<:GaussianKinetic}, r::T, θ::T
+) where {T<:AbstractVecOrMat}
+    M⁻¹ = h.metric.M⁻¹
+    return -r' * M⁻¹ * r / 2
 end
 
 energy(args...) = -neg_energy(args...)
--- a/src/metric.jl
+++ b/src/metric.jl
@@ -99,15 +99,22 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
 end
 
 """
-    RankUpdateEuclideanMetric{T,M} <: AbstractMetric
+    RankUpdateEuclideanMetric{T,AM,AB,AD,F} <: AbstractMetric
 
 A Gaussian Euclidean metric whose inverse is constructed by rank-updates.
+
+# Fields
+
+$(TYPEDFIELDS)
 
 # Constructors
 
     RankUpdateEuclideanMetric(n::Int)
+    RankUpdateEuclideanMetric(M⁻¹, B, D)
 
-Construct a Gaussian Euclidean metric of size `(n, n)` with inverse of `M⁻¹`.
+ - Construct a Gaussian Euclidean metric of size `(n, n)` with `M⁻¹` being diagonal matrix.
+ - Construct a Gaussian Euclidean metric of `M⁻¹`, where `M⁻¹` should be a full rank positive definite matrix,
+    and `B` `D` must be chose so that the Woodbury matrix `W = M⁻¹ + B D B^\\mathrm{T}` is positive definite.
 
 # Example
 
@@ -115,9 +122,13 @@ Construct a Gaussian Euclidean metric of size `(n, n)` with inverse of `M⁻¹`.
 julia> RankUpdateEuclideanMetric(3)
 RankUpdateEuclideanMetric(diag=[1.0, 1.0, 1.0])
 ```
+
+# References
+
+ - Ben Bales, Arya Pourzanjani, Aki Vehtari, Linda Petzold, Selecting the Metric in Hamiltonian Monte Carlo, 2019
 """
 struct RankUpdateEuclideanMetric{T,AM<:AbstractVecOrMat{T},AB,AD,F} <: AbstractMetric
-    # Diagnal of the inverse of the mass matrix
+    "Diagnal of the inverse of the mass matrix"
     M⁻¹::AM
     B::AB
     D::AD
@@ -146,18 +157,23 @@ function RankUpdateEuclideanMetric(::Type{T}, n::Int) where {T}
     factorization = woodbury_factorize(M⁻¹, B, D)
     return RankUpdateEuclideanMetric(M⁻¹, B, D, factorization)
 end
+
+function RankUpdateEuclideanMetric(M⁻¹, B, D)
+    factorization = woodbury_factorize(M⁻¹, B, D)
+    return RankUpdateEuclideanMetric(M⁻¹, B, D, factorization)
+end
+
 function RankUpdateEuclideanMetric(::Type{T}, sz::Tuple{Int}) where {T}
     return RankUpdateEuclideanMetric(T, first(sz))
 end
 RankUpdateEuclideanMetric(sz::Tuple{Int}) = RankUpdateEuclideanMetric(Float64, sz)
 
-AdvancedHMC.renew(::RankUpdateEuclideanMetric, M⁻¹) = RankUpdateEuclideanMetric(M⁻¹)
+renew(::RankUpdateEuclideanMetric, (M⁻¹, B, D)) = RankUpdateEuclideanMetric(M⁻¹, B, D)
 
 Base.size(metric::RankUpdateEuclideanMetric, dim...) = size(metric.M⁻¹.diag, dim...)
 
-function Base.show(io::IO, metric::RankUpdateEuclideanMetric)
-    print(io, "RankUpdateEuclideanMetric(diag=$(diag(metric.M⁻¹)))")
-    return nothing
+function Base.show(io::IO, ::MIME"text/plain", metric::RankUpdateEuclideanMetric)
+    return print(io, "RankUpdateEuclideanMetric(diag=$(diag(metric.M⁻¹)))")
 end
 
 # `rand` functions for `metric` types.
--- a/src/metric.jl
+++ b/src/metric.jl
@@ -153,7 +153,7 @@ RankUpdateEuclideanMetric(sz::Tuple{Int}) = RankUpdateEuclideanMetric(Float64, s
 
 AdvancedHMC.renew(::RankUpdateEuclideanMetric, M⁻¹) = RankUpdateEuclideanMetric(M⁻¹)
 
-Base.size(metric::RankUpdateEuclideanMetric, dim...) = size(metric.M⁻¹, dim...)
+Base.size(metric::RankUpdateEuclideanMetric, dim...) = size(metric.M⁻¹.diag, dim...)
 
 function Base.show(io::IO, metric::RankUpdateEuclideanMetric)
     print(io, "RankUpdateEuclideanMetric(diag=$(diag(metric.M⁻¹)))")
--- a/src/metric.jl
+++ b/src/metric.jl
@@ -98,6 +98,68 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
     return print(io, "DenseEuclideanMetric(diag=$(_string_M⁻¹(dem.M⁻¹)))")
 end
 
+"""
+    RankUpdateEuclideanMetric{T,M} <: AbstractMetric
+
+A Gaussian Euclidean metric whose inverse is constructed by rank-updates.
+
+# Constructors
+
+    RankUpdateEuclideanMetric(n::Int)
+
+Construct a Gaussian Euclidean metric of size `(n, n)` with inverse of `M⁻¹`.
+
+# Example
+
+```julia
+julia> RankUpdateEuclideanMetric(3)
+RankUpdateEuclideanMetric(diag=[1.0, 1.0, 1.0])
+```
+"""
+struct RankUpdateEuclideanMetric{T,AM<:AbstractVecOrMat{T},AB,AD,F} <: AbstractMetric
+    # Diagnal of the inverse of the mass matrix
+    M⁻¹::AM
+    B::AB
+    D::AD
+    factorization::F
+end
+
+function woodbury_factorize(A, B, D)
+    cholA = cholesky(A isa Diagonal ? A : Symmetric(A))
+    U = cholA.U
+    Q, R = qr(U' \ B)
+    V = cholesky(Symmetric(muladd(R, D * R', I))).U
+    return (U=U, Q=Q, V=V)
+end
+
+function RankUpdateEuclideanMetric(n::Int)
+    M⁻¹ = Diagonal(ones(n))
+    B = zeros(n, 0)
+    D = zeros(0, 0)
+    factorization = woodbury_factorize(M⁻¹, B, D)
+    return RankUpdateEuclideanMetric(M⁻¹, B, D, factorization)
+end
+function RankUpdateEuclideanMetric(::Type{T}, n::Int) where {T}
+    M⁻¹ = Diagonal(ones(T, n))
+    B = Matrix{T}(undef, n, 0)
+    D = Matrix{T}(undef, 0, 0)
+    factorization = woodbury_factorize(M⁻¹, B, D)
+    return RankUpdateEuclideanMetric(M⁻¹, B, D, factorization)
+end
+function RankUpdateEuclideanMetric(::Type{T}, sz::Tuple{Int}) where {T}
+    return RankUpdateEuclideanMetric(T, first(sz))
+end
+RankUpdateEuclideanMetric(sz::Tuple{Int}) = RankUpdateEuclideanMetric(Float64, sz)
+
+AdvancedHMC.renew(::RankUpdateEuclideanMetric, M⁻¹) = RankUpdateEuclideanMetric(M⁻¹)
+
+Base.size(metric::RankUpdateEuclideanMetric, dim...) = size(metric.M⁻¹, dim...)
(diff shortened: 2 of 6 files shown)


src/metric.jl
```diff
@@ -98,6 +98,68 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
     return print(io, "DenseEuclideanMetric(diag=$(_string_M⁻¹(dem.M⁻¹)))")
 end
 
+"""
+    RankUpdateEuclideanMetric{T,M} <: AbstractMetric
```
mhauru, line 102: The type parameters seem to be out of date.

src/metric.jl
```diff
@@ -98,6 +98,68 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+
+A Gaussian Euclidean metric whose inverse is constructed by rank-updates.
+
```
mhauru, line 105: Could we add an `$(TYPEDFIELDS)` and add docstrings explaining each field?

src/metric.jl
```diff
@@ -98,6 +98,68 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+# Constructors
+
+    RankUpdateEuclideanMetric(n::Int)
+
+Construct a Gaussian Euclidean metric of size `(n, n)` with inverse of `M⁻¹`.
```
mhauru, line 110: This seems out of date. There seem to be several other constructors as well. Also, `M⁻¹` is not explained (see above for `$(TYPEDFIELDS)`).

src/metric.jl
```diff
@@ -98,6 +98,68 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+function RankUpdateEuclideanMetric(::Type{T}, n::Int) where {T}
+    M⁻¹ = Diagonal(ones(T, n))
+    B = Matrix{T}(undef, n, 0)
+    D = Matrix{T}(undef, 0, 0)
+    factorization = woodbury_factorize(M⁻¹, B, D)
+    return RankUpdateEuclideanMetric(M⁻¹, B, D, factorization)
+end
+function RankUpdateEuclideanMetric(::Type{T}, sz::Tuple{Int}) where {T}
+    return RankUpdateEuclideanMetric(T, first(sz))
+end
```
mhauru, line 151: What's the use case for this constructor? Do we need the ability for the user to pass singleton tuples, or could we only have the `(n::Int)` version?

src/metric.jl
```diff
@@ -98,6 +98,68 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+RankUpdateEuclideanMetric(sz::Tuple{Int}) = RankUpdateEuclideanMetric(Float64, sz)
+
+AdvancedHMC.renew(::RankUpdateEuclideanMetric, M⁻¹) = RankUpdateEuclideanMetric(M⁻¹)
+
+Base.size(metric::RankUpdateEuclideanMetric, dim...) = size(metric.M⁻¹.diag, dim...)
+
+function Base.show(io::IO, metric::RankUpdateEuclideanMetric)
+    print(io, "RankUpdateEuclideanMetric(diag=$(diag(metric.M⁻¹)))")
+    return nothing
+end
```
mhauru, line 161: This doesn't strictly follow the convention that the output of `show(io, x)` methods should include the full details of the object, preferably as parseable Julia code (see [here](https://docs.julialang.org/en/v1/base/io-network/#Base.show-Tuple%7BIO,%20Any%7D)). I'm not too picky about this, it's a convention we violate in a lot of places in TuringLang, but we could consider having a more extensive print out here, and then another method for `how(io::IO, ::MIME"text/plain", x)` that produces more human-readable output.

test/metric.jl
```diff
@@ -25,8 +26,12 @@ using ReTest, Random, AdvancedHMC
         rng = MersenneTwister(1)
         θ = randn(rng, D)
         ℓπ(θ) = 1
-        for metric in
-            [UnitEuclideanMetric(1), DiagEuclideanMetric(1), DenseEuclideanMetric(1)]
+        for metric in [
+            UnitEuclideanMetric(1),
+            DiagEuclideanMetric(1),
+            DenseEuclideanMetric(1),
+            RankUpdateEuclideanMetric(1),
```
mhauru, line 33: This seems to test only the constructor and resizing. Could we have a more extensive test that actually uses `RankUpdateEuclideanMetric` in a sampler and hits all the key methods?

mhauru (changes requested):
I had some comments about docstrings and tests and such, but would be good to also get a review from someone who understands the algorithm that is being implemented here.

sethaxen:
On a semi-related note, this rank-update covariance is the same as the marginal covariance of latent factor models of the following structure:
```math
\begin{aligned}
z &\sim \mathcal{N}(0, D)\\
y &\sim \mathcal{N}(0, A)\\
x &= y + B z  \implies x \sim \mathcal{N}(0, A + B D B^\top)
\end{aligned}
```

There's a rich literature on algorithms for estimating $(A,B,D)$ from draws that we could draw on to implement efficient mass matrix adaptation. I'm working on experiments to see how well this works and how it compares to Pathfinder and to the methods in https://arxiv.org/abs/1905.11916 and whether some combination of these methods outperforms either in isolation.

ErikQQY:
> I'm happy to review, but it would first be nice to get an answer from the maintainers to this question, which I don't think has been answered yet, and which I also raised in https://github.com/TuringLang/AdvancedHMC.jl/issues/411:

IIUC, the main reason why the metrics in AHMC don't use PDMats.jl is that the introduction of PDMats.jl would introduce unnecessary overhead, for example, need some wrappers to store unnecessary matrices etc. But as for the `RankUpdateEuclideanMetric`, I think PDMats.jl is suitable here, especially with the `invwhite` and `invunwhitten` functionality in https://github.com/JuliaStats/PDMats.jl/pull/221.

src/metric.jl
ErikQQY: Agreed, AFAIK, most of the SciML printing interfaces are also using this method, I think we can just switch to a more friendly printing interface, Base.show(io::IO, ::MIME"text/plain", x)` for all the necessary stuff in AHMC here, I will make these changes.

src/metric.jl
ErikQQY: Actually, after second thoughts, we should allow this constructor to enable the resizing of Hamiltonian, see for example,

https://github.com/TuringLang/AdvancedHMC.jl/blob/316e3d9305f93e6374c46304714286a600e5bef6/src/sampler.jl#L24-L31

so we need to keep this.

src/metric.jl
```diff
@@ -99,25 +99,36 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
 RankUpdateEuclideanMetric(diag=[1.0, 1.0, 1.0])
 ```
+
+# References
+
+ - Ben Bales, Arya Pourzanjani, Aki Vehtari, Linda Petzold, Selecting the Metric in Hamiltonian Monte Carlo, 2019
 """
 struct RankUpdateEuclideanMetric{T,AM<:AbstractVecOrMat{T},AB,AD,F} <: AbstractMetric
-    # Diagnal of the inverse of the mass matrix
+    "Diagnal of the inverse of the mass matrix"
```
mhauru, line 131:
```suggestion
    "Diagonal of the inverse of the mass matrix"
```

src/metric.jl
```diff
@@ -99,25 +99,36 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
     M⁻¹::AM
     B::AB
     D::AD
```
mhauru, line 134:
```suggestion
    D::AD
    "Woodbury factorisation of M⁻¹ + B D transpose(B)"
```

src/metric.jl
```diff
@@ -99,25 +99,36 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
 end
 
 """
-    RankUpdateEuclideanMetric{T,M} <: AbstractMetric
+    RankUpdateEuclideanMetric{T,AM,AB,AD,F} <: AbstractMetric
```
mhauru, line 102:
```suggestion
    RankUpdateEuclideanMetric{T,AM<:AbstractVecOrMat{T},AB,AD,F} <: AbstractMetric
```

src/metric.jl
mhauru, line 134: Is there some nice explanation of what the purpose of `B` and `D` is, why we want this Woodbury factorisation, or how this is important for RankUpdateEuclideanMetric? If not, can just refer to somewhere else for an explanation, but would be good to say something about them.

src/metric.jl
```diff
@@ -99,25 +99,36 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
 ```julia
 julia> RankUpdateEuclideanMetric(3)
 RankUpdateEuclideanMetric(diag=[1.0, 1.0, 1.0])
 ```
+
+# References
+
+ - Ben Bales, Arya Pourzanjani, Aki Vehtari, Linda Petzold, Selecting the Metric in Hamiltonian Monte Carlo, 2019
 """
 struct RankUpdateEuclideanMetric{T,AM<:AbstractVecOrMat{T},AB,AD,F} <: AbstractMetric
```
mhauru, line 130: More a question than a request: Is there a reason why `AM` can be a Vector? Also, is it intentional that `AB` and `AD` don't have to have the same element type?

mhauru (changes requested):
Thanks @ErikQQY! I have a few more small points, I'm admittedly being picky about the docs.

src/metric.jl
```diff
@@ -98,6 +98,84 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+$(TYPEDFIELDS)
+
+# Constructors
+
+    RankUpdateEuclideanMetric(n::Int)
+    RankUpdateEuclideanMetric(M⁻¹, B, D)
+
+ - Construct a Gaussian Euclidean metric of size `(n, n)` with `M⁻¹` being diagonal matrix.
+ - Construct a Gaussian Euclidean metric of `M⁻¹`, where `M⁻¹` should be a full rank positive definite matrix,
+    and `B` `D` must be chose so that the Woodbury matrix `W = M⁻¹ + B D B^\\mathrm{T}` is positive definite.
```
sethaxen, line 117: For the other metrics `M⁻¹` is defined as the inverse of the metric. Wouldn't it be more consistent to here define `M⁻¹ = A + B D B'` as the inverse metric and require A be diagonal and positive-definite? (also, "full-rank" for diagonal PD-mat is redundant).

sethaxen: For Pathfinder to safely convert its `WoodburyPDMat` to this type, we'll need the contents/form of the `factorization` field to be documented.

src/metric.jl
```diff
@@ -98,6 +98,84 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+struct RankUpdateEuclideanMetric{T,AM<:AbstractVecOrMat{T},AB,AD,F} <: AbstractMetric
+    "Diagnal of the inverse of the mass matrix"
+    M⁻¹::AM
+    B::AB
+    D::AD
+    factorization::F
+end
+
+function woodbury_factorize(A, B, D)
+    cholA = cholesky(A isa Diagonal ? A : Symmetric(A))
```
sethaxen, line 139: Pathfinder's implementation allows for arbitrary PD `A` but for the use-cases in Pathfinder and Bales's paper, diagonal `A` is sufficient. Since you've already documented `A` as diagonal, perhaps you can drop this check.

src/metric.jl
```diff
@@ -98,6 +98,84 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+    U = cholA.U
+    Q, R = qr(U' \ B)
+    V = cholesky(Symmetric(muladd(R, D * R', I))).U
+    return (U=U, Q=Q, V=V)
```
sethaxen, line 143:
```suggestion
    return (; U, Q, V)
```

src/metric.jl
```diff
@@ -98,6 +98,84 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+end
+
+function RankUpdateEuclideanMetric(n::Int)
+    M⁻¹ = Diagonal(ones(n))
+    B = zeros(n, 0)
+    D = zeros(0, 0)
+    factorization = woodbury_factorize(M⁻¹, B, D)
+    return RankUpdateEuclideanMetric(M⁻¹, B, D, factorization)
+end
```
sethaxen, line 152: Note: this is probably fine for now, but there are multiple ways to form the identity matrix here, and later it might be better to initialize a different way (e.g. for tuning a covariance matrix for factor analysis, `B=0` and `D=0` prohibits convergence)

sethaxen: I know it breaks with the API of other metric constructors, but I think you want to be able to initialize the rank of the update as well. Otherwise you have no way to fix the rank for future tuning algorithms.

src/metric.jl
```diff
@@ -98,6 +98,84 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+    return RankUpdateEuclideanMetric(T, first(sz))
+end
+RankUpdateEuclideanMetric(sz::Tuple{Int}) = RankUpdateEuclideanMetric(Float64, sz)
+
+renew(::RankUpdateEuclideanMetric, (M⁻¹, B, D)) = RankUpdateEuclideanMetric(M⁻¹, B, D)
+
+Base.size(metric::RankUpdateEuclideanMetric, dim...) = size(metric.M⁻¹.diag, dim...)
+
+function Base.show(io::IO, ::MIME"text/plain", metric::RankUpdateEuclideanMetric)
+    return print(io, "RankUpdateEuclideanMetric(diag=$(diag(metric.M⁻¹)))")
```
sethaxen, line 176: Here's another example where calling the full-rank part of the inverse-metric `M⁻¹` leads to a different diagonal being shown than the diagonal of the full inverse-metric. The diagonal entries needed can be efficiently computed, see https://github.com/mlcolab/Pathfinder.jl/blob/f4ca90dc3d91f077f479d13904a2b6bf99e8ee25/src/woodbury.jl#L326-L329

src/metric.jl
```diff
@@ -98,6 +98,84 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
...
+# Example
+
+```julia
+julia> RankUpdateEuclideanMetric(3)
+RankUpdateEuclideanMetric(diag=[1.0, 1.0, 1.0])
+```
+
+# References
+
+ - Ben Bales, Arya Pourzanjani, Aki Vehtari, Linda Petzold, Selecting the Metric in Hamiltonian Monte Carlo, 2019
```
sethaxen, line 128: The factorized form used here is due to the Pathfinder paper, so I recommend also citing http://jmlr.org/papers/v23/21-0889.html

src/metric.jl
sethaxen, line 117:
```suggestion
    and `B` `D` must be chosen so that the Woodbury matrix `W = M⁻¹ + B D B^\\mathrm{T}` is positive definite.
```

src/metric.jl
```diff
@@ -98,6 +98,84 @@ function Base.show(io::IO, dem::DenseEuclideanMetric)
     return print(io, "DenseEuclideanMetric(diag=$(_string_M⁻¹(dem.M⁻¹)))")
 end
 
+"""
+    RankUpdateEuclideanMetric{T,AM,AB,AD,F} <: AbstractMetric
+
+A Gaussian Euclidean metric whose inverse is constructed by rank-updates.
```
sethaxen, line 104:
```suggestion
A Gaussian Euclidean metric whose inverse is constructed by low-rank updates to a diagonal matrix.
```

src/hamiltonian.jl
```diff
@@ -165,6 +177,13 @@ function neg_energy(
     return -dot(r, h.metric._temp) / 2
 end
 
+function neg_energy(
+    h::Hamiltonian{<:RankUpdateEuclideanMetric,<:GaussianKinetic}, r::T, θ::T
+) where {T<:AbstractVecOrMat}
+    M⁻¹ = h.metric.M⁻¹
+    return -r' * M⁻¹ * r / 2
```
sethaxen, line 184: Note that because you've defined `M` as just the diagonal part, this equation is incorrect. The full one is https://github.com/mlcolab/Pathfinder.jl/blob/f4ca90dc3d91f077f479d13904a2b6bf99e8ee25/src/integration/advancedhmc.jl#L82, which uses https://github.com/mlcolab/Pathfinder.jl/blob/f4ca90dc3d91f077f479d13904a2b6bf99e8ee25/src/woodbury.jl#L384-L388

sethaxen (changes requested):

src/metric.jl
sethaxen: I haven't seen one place that explains it all. `D` is not *needed*, since it can be absorbed into `B` with any square-root factorization; it's just sometimes more convenient to have it (e.g. Pathfinder directly constructs it). On the other hand, when fitting a factor model, `D` is usually constrained to be `I`.

One interpretation comes from the factor model written in https://github.com/TuringLang/AdvancedHMC.jl/pull/443#issuecomment-2872256440. It's useful when most of the structure of the covariance lies on a low-dimensional subspace. One such example is when the a small number of parameters are highly correlated with each other and not with any other parameters, and all remaining parameters are uncorrelated. Typically efficiency of HMC would be hindered by the high correlations, and one would need a dense covariance matrix, but then you end up with quadratic cost for each leapfrog stop. But if you can roughly upper bound the number of correlated parameters as << the number of sampled parameters, you can use a rank-update metric and get linear cost.

It's a mouthful, so I'm not certain the explanation belongs here. I think an explanation could wait until adaptation methods (e.g. Bales') are included here.

yebai:
Fixed by https://github.com/TuringLang/AdvancedHMC.jl/pull/511; thanks to @devmotion and @sethaxen.

PR #445, opened 2025-05-11
Add relativistic kinetic

ErikQQY:
Add the relativistic kinetic from "Relativistic Monte Carlo" which is implemented in the research directory but has not been merged into the package.

There are still some TODOs:
- The usage of AdaptiveRejectionSampling.jl might have to be a hard dep, not sure how we can make it a weak dep here since we need them when we are making a random momentum for the phasepoint.
- Need some documentation and docstrings.
- Oh, it occurs to me that maybe we also need to first move these files from the research directory to the package repo so that the git history can be preserved?

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -4,6 +4,7 @@ version = "0.8.0"
 
 [deps]
 AbstractMCMC = "80f14c24-f653-4e6a-9b94-39d6b0f70001"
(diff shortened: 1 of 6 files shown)


nsiccha:
To me it looks like the rejection sampling is not needed, @yebai.

Issue #271, opened 2021-07-15
Adaptation tests fail if unlucky with random seed


torfjelde:
https://github.com/TuringLang/AdvancedHMC.jl/blob/564af93598a671c89722ef4c0795f196418c56d5/test/adaptation.jl#L148

fails sometimes, e.g. https://github.com/TuringLang/AdvancedHMC.jl/runs/3073138476.

To determine the fix/cause needs further investigation.

PR #446, opened 2025-05-14
Better printing

ErikQQY:
Use `Base.show(io::IO, ::MIME"text/plain", ::Type)` for more fine-grained printing.

### DIFF
--- a/src/adaptation/stepsize.jl
+++ b/src/adaptation/stepsize.jl
@@ -84,8 +84,8 @@ getϵ(fss::FixedStepSize) = fss.ϵ
 struct ManualSSAdaptor{T<:AbstractScalarOrVec{<:AbstractFloat}} <: StepSizeAdaptor
     state::MSSState{T}
 end
-function Base.show(io::IO, mime::MIME"text/plain", a::ManualSSAdaptor)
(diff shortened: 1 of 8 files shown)


ErikQQY:
Seems CI hit #271 again, need to investigate

src/adaptation/stepsize.jl
```diff
@@ -1,10 +1,22 @@
 ### Mutable states
+"""
+    Dual Averaging state
```
mhauru, line 3:
```suggestion
    DAState{T<:AbstractScalarOrVec{<:AbstractFloat}}
```
Thanks for spotting that this and a few other types needed docstrings. I know that the description "Dual Averaging state" is more informative than the name "DAState", but the Julia docstring [docs](https://docs.julialang.org/en/v1/manual/documentation/#Writing-Documentation) do request that the first line would always be the actual signature of the function/struct.

ErikQQY: Agreed, there are still tons of docstrings that need such formation, but let's focus on the printing stuff here and leave these docstrings for another thorough update.

yebai:
Thanks @ErikQQY and @mhauru!

PR #454, opened 2025-06-08
Fix allocations introduced in `step`

devmotion:
Fixes the unnecessary allocation introduced in https://github.com/TuringLang/AdvancedHMC.jl/pull/431 (see https://github.com/TuringLang/AdvancedHMC.jl/pull/431/files#r2128763944). I could have just reverted to the state prior to #431 but I felt that it would be cleaner to just not use `res` at all in the case `fullTraj = false`.

### DIFF
--- a/src/integrator.jl
+++ b/src/integrator.jl
@@ -218,10 +218,8 @@ function step(
     ϵ = fwd ? step_size(lf) : -step_size(lf)
     ϵ = ϵ'
 
-    res = if FullTraj
-        Vector{P}(undef, n_steps)
-    else
-        Vector{P}(undef, 1)
+    if FullTraj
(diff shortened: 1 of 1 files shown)


src/integrator.jl
```diff
@@ -244,20 +242,18 @@ function step(
...
         if FullTraj
             res[i] = z
-        else
-            res[1] = z
         end
         if !isfinite(z)
             # Remove undef
             if FullTraj
-                res = res[isassigned.(Ref(res), 1:n_steps)]
+                resize!(res, i)
```
devmotion, line 249: It seemed a bit useless to compute the not-`undef`ed entries if we already know that exactly entries 1 to `i` are assigned some value.

ErikQQY (approved):

PR #456, opened 2025-06-08
No string interpolation in `show` methods

devmotion:
String interpolation in `show` methods (many instances recently introduced in https://github.com/TuringLang/AdvancedHMC.jl/pull/446) is unnecessary. The Julia docs even explicitly advise against it: https://docs.julialang.org/en/v1/manual/performance-tips/#Avoid-string-interpolation-for-I/O


ErikQQY (approved):
LGTM, string interpolations is not good since they introduce overheads.

PR #471, opened 2025-10-29
Implement `AbstractMCMC.getstats`

penelopeysm:


penelopeysm:
I ran the tests locally and it's only JET that's failing, so I'll open a separate issue for that.

PR #481, opened 2025-12-02
Consolidate RHMC Changes

J-Price-3:
Changes include:
- Type stability (updated core RHMC functions as well as utility functions in research folder)
- Small optimisations
- Addition of Implicit Midpoint Integrator

We should consolidate all the RHMC changes so that there is one PR to the main branch.


src/riemannian/hamiltonian.jl
THargreaves, line 211: `map` isn't defined for an arbitrary Riemannian metric. You'll need to introduce a layer of abstraction.

A abstract metric could have a `get_G` method which internally calls `map` if it is a Dense metric

nsiccha:
This can be closed in favour of #485, right?

PR #496, opened 2026-05-05
Replace OrdinaryDiffEq weakdep with OrdinaryDiffEqSymplecticRK

ChrisRackauckas-Claude:
> [!NOTE]
> This PR should be ignored until reviewed by @ChrisRackauckas.

## Summary

Replaces the `OrdinaryDiffEq` weakdep with the smaller `OrdinaryDiffEqSymplecticRK` sub-package. The `DiffEqIntegrator` extension only ever used the symplectic solvers (`VerletLeapfrog` and friends) plus the standard `init` / `step!` / `DynamicalODEProblem` interface — none of which requires the full meta-package.

## Why

- **Lighter dep graph.** Users no longer pull in BDF, Rosenbrock, Tsit5, Verner, Default, OrdinaryDiffEqCore, NonlinearSolve, LinearSolve, etc. just to use a leapfrog inside HMC.
- **Honest deps.** The extension actually only uses symplectic RK methods.
- **Sidesteps the `OrdinaryDiffEq` v6 ↔ v7 compat ratchet entirely.** `OrdinaryDiffEqSymplecticRK` v1 (built against `OrdinaryDiffEq` v6) and v2 (built against v7) both expose the same surface, so compat is just `"1, 2"`.
- **Aligns with the v7 ecosystem story.** In `OrdinaryDiffEq` v7, `VerletLeapfrog` no longer re-exports out of `OrdinaryDiffEq` itself (it lives in the symplectic sublib). Depending on the sublib directly is the forward-compatible shape.
- **Unblocks downstream `Turing` v0.45 + SciML stack composition** (e.g. `SciMLExpectations.jl/docs`), where `Turing 0.45 → AdvancedHMC ≥ 0.8.3` and `OrdinaryDiffEq 7` cannot currently coexist due to the `OrdinaryDiffEq = "6"` weakdep compat.

## Changes

- `Project.toml`: weakdep `OrdinaryDiffEq` → `OrdinaryDiffEqSymplecticRK`; extension entry, compat (`"1, 2"`), and `[extras]` updated.
- `ext/AdvancedHMCOrdinaryDiffEqExt.jl` → `ext/AdvancedHMCOrdinaryDiffEqSymplecticRKExt.jl` (rename + module name + import).
- `test/runtests.jl`, `test/integrator.jl`: switch `using OrdinaryDiffEq` to `using OrdinaryDiffEqSymplecticRK`.
- `src/AdvancedHMC.jl`: `register_error_hint` message updated ("Did you forget to load `OrdinaryDiffEqSymplecticRK`?").
- `src/integrator.jl`: inline comment updated.

## Verification

Tested locally on Julia 1.10:

- `AHMC_TEST_GROUP=AdvancedHMC` test suite: **5052 tests, exit 0, no errors**.
- `Integrator` testset: **35/35 pass**.
- Numerical equivalence check (independent script): `DiffEqIntegrator(VerletLeapfrog())` produces bit-identical trajectories to native `Leapfrog` over 1000 steps; Hamiltonian drift ~1.7e-5 over the same trajectory.

## Relation to existing CompatHelper PRs

CompatHelper opened #494 and #495 with the simpler compat-only bump on the `OrdinaryDiffEq` weakdep (`"6"` → `"6, 7"`). That approach works but keeps the whole meta-package as a dep, and breaks the existing `test/integrator.jl` test which references `OrdinaryDiffEq.VerletLeapfrog` (in v7 that name moved to `OrdinaryDiffEqSymplecticRK`). This PR avoids both problems by depending on the sublib directly. Closing #494 and #495 in favor of this would be reasonable.

## Test plan

- [x] Extension precompiles against both `OrdinaryDiffEqSymplecticRK` v1 and v2
- [x] Full `AdvancedHMC` test group passes locally on Julia 1.10
- [x] `DiffEqIntegrator(VerletLeapfrog())` matches native `Leapfrog` numerically
- [ ] CI: GitHub Actions (Julia 1, min, pre × ubuntu/macOS/windows)
- [ ] CI: buildkite CUDA tests

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ChrisRackauckas:
This version I think is better, just moving to the smaller dep that you actually use. I think this is good to go.

ChrisRackauckas:
No review here? No merge?

yebai:
@ChrisRackauckas, would it be possible to move this extension to `SciML`?  

The PR looks sensible to me, other than the question of where this should live.

ChrisRackauckas:
That would violate all conventions. I'm not sure why it would be unclear it would live here. It only uses a small surface of SciML, and all of that is public API. Meanwhile, almost all of the extension is using AdvancedHMC non-public API. Unless all of that is going to get documented and tested to unsure it never changes, this is only maintainable as part of the package repo where the non-public API is used. That's a standard.

ChrisRackauckas:
That said, I'd be happy to help maintain the code anytime, so ping me if it ever needs help updating or if anything is up with the tests.

Project.toml
```diff
@@ -41,7 +41,11 @@ LinearAlgebra = "<0.1, 1"
 LogDensityProblems = "2"
 LogDensityProblemsAD = "1"
 MCMCChains = "5, 6, 7"
+<<<<<<< bump-ordinarydiffeq-v7-compat
```
bvdmitri:
```suggestion
```

Project.toml
```diff
@@ -41,7 +41,11 @@ LinearAlgebra = "<0.1, 1"
...
+OrdinaryDiffEqSymplecticRK = "1, 2"
+=======
 OrdinaryDiffEq = "6, 7"
+>>>>>>> main
```
bvdmitri:
```suggestion
```

Project.toml
```diff
@@ -41,7 +41,11 @@ LinearAlgebra = "<0.1, 1"
 LogDensityProblems = "2"
 LogDensityProblemsAD = "1"
 MCMCChains = "5, 6, 7"
+<<<<<<< bump-ordinarydiffeq-v7-compat
+OrdinaryDiffEqSymplecticRK = "1, 2"
+=======
```
bvdmitri:
```suggestion
```

bvdmitri (commented):
should be removed?

sunxd3:
@ChrisRackauckas thanks!

yebai:
@ChrisRackauckas we're happy to keep this pkg extension here for now as a workaround. TuringLang does not use this feature, so this is exclusively ScIML functionality. Julia’s package extension mechanism is useful in some cases, but it also opens a broader set of concerns around using package internals to extend functionality beyond a package’s intended scope, as in the case of AdvancedHMC.

sunxd3:
https://github.com/JuliaRegistries/General/pull/155178 for tracking release. Chris, I don't expect this to break things, but worth monitoring for the potential extension loading behavior change.

PR #500, opened 2026-05-28
Replace StatsFuns with LogExpFunctions + IrrationalConstants

devmotion:
I just released StatsFuns@2 (mainly a precaution because it breaks some existing Distributions releases which used some internals), but when I started to update AdvancedHMC I noticed that it shouldn't actually depend on the full-blown StatsFuns but instead use the two packages that were factored out from StatsFuns years ago.


yebai (approved):
Looks sensible, thanks @devmotion!

PR #511, opened 2026-06-25
Add `RankUpdateEuclideanMetric`

devmotion:
Resurrects #443 (by @ErikQQY), which was reviewed by @mhauru and @sethaxen but then stalled. Based on the original branch (commits preserved) with `main` merged in and the review addressed. Part of #277 and #411.

As in #443, this does not depend on PDMats.jl (cf. #34) and computes the Woodbury factorization directly.

Changes relative to #443:
- `∂H∂r` and `neg_energy` now use the full Woodbury inverse mass matrix `M⁻¹ = A + B D Bᵀ`. Previously only the diagonal `A` was used, inconsistently with `rand_momentum`, so the kinetic energy and its gradient were wrong whenever `B, D ≠ 0` (pointed out by @sethaxen). The diagonal field is renamed `M⁻¹` -> `A`, so `M⁻¹` refers to the full inverse metric as for the other metrics.
- The factorization is stored in a `WoodburyFactorization{T,TU<:Diagonal{T},TQ<:AbstractQ{T},TV<:UpperTriangular{T}}` instead of a `NamedTuple`, and the metric fields are typed (diagonal `A`, matching element types).
- Docstrings (`$(TYPEDFIELDS)`, reference to Zhang et al. (2022), whose Pathfinder factorization this uses) and `show` follow the `MIME"text/plain"` convention on `main`; `show` prints the diagonal of the full `M⁻¹`, like `DenseEuclideanMetric`.
- `rand_momentum` is restricted to a single chain (its last argument is `::AbstractVector`): the metric has no multi-chain `(n, n_chains)` form, so `size(metric)` is always a 1-tuple and the momentum is always a vector.
- Mass matrix adaptation is not implemented for this metric. There is no `MassMatrixAdaptor(::RankUpdateEuclideanMetric)` (and no `getM⁻¹` returning the `(A, B, D)` factors `renew` would need), so a `renew` method would be unreachable; it is omitted rather than left as an untested stub, and the metric is skipped for `MassMatrixAdaptor` in the sampler tests. Adding low-rank online covariance estimation (the Pathfinder/nutpie direction) is left as future work.
- Tests cross-check energy, gradient, and momentum covariance against a `DenseEuclideanMetric` with the same `M⁻¹`, check type stability with `@inferred`, cover the constructors, and check both `show` methods.

Still open from #443: whether to use PDMats.jl instead (#34) was never decided.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/metric.jl
```diff
@@ -134,6 +134,155 @@ function Base.show(io::IO, ::MIME"text/plain", dem::DenseEuclideanMetric{T}) whe
...
+definite diagonal `A`, as returned by [`woodbury_factorize`](@ref).
+
+The factors `U`, `Q`, and `V` are defined by the field descriptions below; together they
+allow sampling from `N(0, W⁻¹)` without forming `W`.
+
+# Fields
+
+$(TYPEDFIELDS)
+"""
+struct WoodburyFactorization{T,TU<:Diagonal{T},TQ<:AbstractQ{T},TV<:UpperTriangular{T}}
```
sethaxen, line 155: While typically a user won't need access to this, it would be useful if this was part of the API, so that Pathfinder could construct this from its own factorization and the metric doesn't need to be re-factorized when crossing the boundary from Pathfinder to AdvancedHMC:

see https://github.com/mlcolab/Pathfinder.jl/pull/395/changes#diff-fe44e2cab6f2e18ea6910928a0b507f70f795f0a25b518e77512ec62978aaf55R21

devmotion: Would making it `public` be sufficient? Or, more generally, would some other design be better for Pathfinder (and possibly other packages interacting with AdvancedHMC)?

My main motivation here is to improve the integration with Pathfinder, so my goal is to find a design that works well with it.

sethaxen: > Would making it `public` be sufficient? Or, more generally, would some other design be better for Pathfinder (and possibly other packages interacting with AdvancedHMC)?

I don't know about other packages, but for Pathfinder, yes, this would be enough.

src/metric.jl
```diff
@@ -134,6 +134,155 @@ function Base.show(io::IO, ::MIME"text/plain", dem::DenseEuclideanMetric{T}) whe
...
+end
+
+function RankUpdateEuclideanMetric(::Type{T}, sz::Tuple{Int}) where {T}
+    return RankUpdateEuclideanMetric(T, first(sz))
+end
+RankUpdateEuclideanMetric(sz::Tuple{Int}) = RankUpdateEuclideanMetric(Float64, sz)
+
+function renew(::RankUpdateEuclideanMetric, (A, B, D))
+    return RankUpdateEuclideanMetric(A, B, D)
+end
```
devmotion: I noticed that this function is currently unreachable since no `MassMatrixAdaptor` exists for `RankUpdateEuclideanMetric`.

devmotion: Removed in https://github.com/TuringLang/AdvancedHMC.jl/pull/511/commits/79285de3989e7e1464c91ed68f845cdcac155ff4

src/hamiltonian.jl
```diff
@@ -67,6 +67,19 @@ function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::A
     return M⁻¹ * r
 end
 
+function ∂H∂r(
+    h::Hamiltonian{<:RankUpdateEuclideanMetric,<:GaussianKinetic}, r::AbstractVecOrMat
+)
+    (; A, B, D) = h.metric
+    axes_A = __axes(A)
+    axes_r = __axes(r)
+    (last(axes_A) !== first(axes_r)) && throw(
```
sethaxen: Is the equivalence too strict? Would prevent axes with the same indices but different subtypes of `AbstractUnitRange` from being used together.
```suggestion
    (last(axes_A) != first(axes_r)) && throw(
```

src/hamiltonian.jl
```diff
@@ -67,6 +67,19 @@ function ∂H∂r(h::Hamiltonian{<:DenseEuclideanMetric,<:GaussianKinetic}, r::A
...
+        ArgumentError(lazy"AxesMismatch: A has axes $(axes_A) but r has axes $(axes_r)")
+    )
+    # M⁻¹ * r = A * r + B * (D * (Bᵀ * r))
```
sethaxen: This comment doesn't seem to add anything relative to the following expression.
```suggestion
```

src/hamiltonian.jl
```diff
@@ -171,6 +184,15 @@ function neg_energy(
     return -dot(r, h.metric._temp) / 2
 end
 
+function neg_energy(
+    h::Hamiltonian{<:RankUpdateEuclideanMetric,<:GaussianKinetic}, r::T, θ::T
+) where {T<:AbstractVector}
+    (; A, B, D) = h.metric
+    Btr = B' * r
+    # rᵀ M⁻¹ r = rᵀ A r + (Bᵀ r)ᵀ D (Bᵀ r)
```
sethaxen: Same, this is obvious from the following expression:
```suggestion
```

src/metric.jl
```diff
@@ -134,6 +134,151 @@ function Base.show(io::IO, ::MIME"text/plain", dem::DenseEuclideanMetric{T}) whe
...
+
+A factorization of a positive definite Woodbury matrix `W = A + B*D*Bᵀ`, with positive
+definite diagonal `A`, as returned by [`woodbury_factorize`](@ref).
+
+The factors `U`, `Q`, and `V` are defined by the field descriptions below; together they
+allow sampling from `N(0, W⁻¹)` without forming `W`.
+
+# Fields
+
+$(TYPEDFIELDS)
```
sethaxen, line 148: Since the factorization is the part of this that is due to the Pathfinder paper, I would include that reference in this docstring.

src/metric.jl
```diff
@@ -134,6 +134,151 @@ function Base.show(io::IO, ::MIME"text/plain", dem::DenseEuclideanMetric{T}) whe
...
+"""
+    woodbury_factorize(A::Diagonal, B::AbstractMatrix, D::AbstractMatrix)
+
+Return a [`WoodburyFactorization`](@ref) of the positive definite Woodbury matrix
+`W = A + B*D*Bᵀ`, where `A` is a positive definite diagonal matrix.
+"""
+function woodbury_factorize(
+    A::Diagonal{T}, B::AbstractMatrix{T}, D::AbstractMatrix{T}
+) where {T}
+    U = cholesky(A).U
```
sethaxen: Wouldn't this be simpler?
```suggestion
    U = sqrt(A)
```

src/metric.jl
```diff
@@ -134,6 +134,151 @@ function Base.show(io::IO, ::MIME"text/plain", dem::DenseEuclideanMetric{T}) whe
...
+    Q, R = qr(U' \ B)
```
sethaxen: It's diagonal, so no need to adjoint:
```suggestion
    Q, R = qr(U \ B)
```

src/metric.jl
```diff
@@ -134,6 +134,151 @@ function Base.show(io::IO, ::MIME"text/plain", dem::DenseEuclideanMetric{T}) whe
...
+    V = cholesky(Symmetric(muladd(R, D * R', I))).U
+    return WoodburyFactorization(U, Q, V)
+end
+
+"""
+    RankUpdateEuclideanMetric([T::Type=Float64,] n::Int)
+    RankUpdateEuclideanMetric([T::Type=Float64,] sz::Tuple{Int})
+    RankUpdateEuclideanMetric(A::Diagonal, B::AbstractMatrix, D::AbstractMatrix)
+
+A Gaussian Euclidean metric in `n` dimensions whose inverse mass matrix `M⁻¹ = A + B*D*Bᵀ`
```
sethaxen, line 184: Somewhere here it should specify that `D` must be symmetric.

src/metric.jl
```diff
@@ -134,6 +134,151 @@ function Base.show(io::IO, ::MIME"text/plain", dem::DenseEuclideanMetric{T}) whe
...
+
+Base.eltype(::RankUpdateEuclideanMetric{T}) where {T} = T
+Base.size(metric::RankUpdateEuclideanMetric, dim...) = size(metric.A.diag, dim...)
+
+# Work around JuliaLang/LinearAlgebra.jl#1485: 3-arg `dot` errored on empty `x`/`y` before
+# Julia 1.12.3, which the rank-0 metric triggers.
+@static if VERSION < v"1.12.3"
+    function _dot(x::AbstractVector, A::AbstractMatrix, y::AbstractVector)
+        any(isempty, (x, y)) &&
+            return zero(dot(zero(eltype(x)), zero(eltype(A)), zero(eltype(y))))
```
sethaxen, line 255: Wouldn't it make more sense to move the version check in here?

src/metric.jl
```diff
@@ -134,6 +134,151 @@ function Base.show(io::IO, ::MIME"text/plain", dem::DenseEuclideanMetric{T}) whe
...
+# Diagonal of the full inverse mass matrix `M⁻¹ = A + B*D*Bᵀ`, i.e. `diag(A) + [bᵢᵀ*D*bᵢ]`.
+function _diag_inv_metric(metric::RankUpdateEuclideanMetric)
+    return broadcast(
+        (aii, b) -> aii + _dot(b, metric.D, b), metric.A.diag, eachrow(metric.B)
+    )
+end
+
+function Base.show(io::IO, metric::RankUpdateEuclideanMetric)
+    return print(
+        io, "RankUpdateEuclideanMetric(", _string_M⁻¹(_diag_inv_metric(metric)), ")"
```
sethaxen, line 271: When logging, this might be called after every trajectory, even if the metric remains fixed, right? If so, it seems wasteful, and I wonder if it would be better to compute the diagonal upon construction and then store it.

sethaxen (changes requested):
Looks good overall, just some minor suggestions. I've also verified in mlcolab/Pathfinder.jl#395 that it works as intended with Pathfinder.jl.

Would it be useful to document somewhere that `RankUpdateEuclidenMetric` can be instantiated using Pathfinder?

src/metric.jl
devmotion: Yeah, we could do that. Having the check on the outside would allow us to just define `const _dot = dot` which might be even cleaner for more recent Julia versions.

src/metric.jl
devmotion: I'm generally not convinced of the display conventions in AdvancedHMC, without the truncation it would become unreadable but with the truncation it seems a bit questionable what value the displayed numbers actually provide. I'd also argue it's generally questionable whether these diagonals should be printed in the short single-line form of `show`. And additionally, it seems somewhat surprising that even for non-diagonal inverse metrics only the diagonal is displayed.

sethaxen: Yeah, I guess since changes to `show` methods are non-breaking, it would make more sense to not store the diagonal if the `show` behavior could change in the future to not show the diagonal.

sethaxen (approved):
LGTM!

